PR #3, opened 2025-05-21
Test other DBMSs with Ducklake

pdet:
This script is semi-ghetto, but it allows us to rewrite our existing ducklake tests to run with either sqlite, postgres or mysql.

To call it you just need to specify where the dbms extension is and what dbms you are calling:
e.g.,
`python3 scripts/change_db.py /Users/holanda/Documents/Projects/ducklake/duckdb/build/release/extension/postgres_scanner/postgres_scanner.duckdb_extension postgres
`

It then alters all tests to use that database, and run each test one by one, so it can also cleanup the database between tests.

It assumes a release build of ducklake.

After running tests for one of the systems you must restore the changed files, e.g.,
`git restore test ...`


Mytherin:
Thanks!

PR #8, opened 2025-05-23
Fail cleanly when GetEntry fails, plus autoloading of parquet

carlopi:
Most clients do ship `parquet` already, but still worth sorting out autoloading also in the insert path.


Mytherin:
Thanks!

PR #10, opened 2025-05-23
Add description, unsure on text

carlopi:
Due to current duckdb limitations, this shows up only after LOAD This is minor since it would be eventually handled duckdb side, but this allows to have it in some capacity already for 1.3.0


Mytherin:
Thanks!

PR #14, opened 2025-05-26
Fix autoloading required extension for data_path

carlopi:
```
ATTACH 'ducklake:autoload_problem.ducklake' (DATA_PATH 's3://some-bucket/');
```
will, without this PR, do not autoload httpfs, and eventually use the folder s3://some-bucket using the default folder when `httpfs` is missing OR remote bucket when `httpfs` is present.

This PR makes so that in either case an initialized DuckLake has to be usable, possible paths are:
* extension already present (or no extension needed), nop
* required extension is autoinstalled and autoloaded
* user is informed of which extension is needed, and ducklake ATTACH fails

As a consequence of this, secrets are also loaded, but there is no actual test the path is actually reachable (say networking or mismatched secrets). For that an insert or scan is actually needed (keeping same behaviour as now, so initialize perform no network request to `data_path`).

Test shows what are the possibilities, to be able to run the test (but for the last line), the following can be used
```
GEN=ninja make
LOCAL_EXTENSION_REPO='build/release/repository' ./build/release/test/unittest --test-dir . test/sql/autoloading/autoload_data_path.test
```

To manage also the last like, `httpfs` extension is needed to be built (but not linked) together with ducklake.

### DIFF
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -90,16 +90,17 @@ void DuckLakeInitializer::InitializeDataPath() {
 	if (data_path.empty()) {
 		return;
 	}
-	if (!StringUtil::EndsWith(data_path, "/") && !StringUtil::EndsWith(data_path, "\\")) {
-		// data path does not end in a path separator - add it
-		auto &fs = FileSystem::GetFileSystem(context);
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #15, opened 2025-05-26
Add example using Minio for S3 storage and local DuckDB file

carlopi:
Unsure if here or in the main docs. I think possibly both are fine.


Mytherin:
Thanks! Ideally this is moved to the main repo later

PR #16, opened 2025-05-26
Publish extension in core-nightly bucket

carlopi:
After this, in released 1.3.0, it should be possible to do:
```
FORCE INSTALL ducklake FROM core_nightly;
```


Mytherin:
Thanks!

PR #17, opened 2025-05-26
Make db name unique in test

pdet:
This test fails when running all tests on a windows machine since the db name is used in a previous test.


Mytherin:
Thanks!

PR #18, opened 2025-05-26
[feature] Add a data_path tag, so that it can be retrived SQL side

carlopi:
Example:
```
FROM duckdb_databases() WHERE tags['data_path'] IS NOT NULL;
```

There are no tests (yet)


Mytherin:
Sure - can you also add a test?

PR #22, opened 2025-05-26
Fix `GetCopyFunction` for motherduck

NiclasHaderer:
For some reason the current way does not work for MotherDuck. This makes some minor adjustments to the copy function retrieval which makes it MotherDuck compatible.

PS: I would normally not open this PR without understanding why it is not working within MotherDuck, but I'm unfortunately kind of running out of time. I hope this is OK with you.


Mytherin:
Thanks! LGTM

PR #23, opened 2025-05-27
Add initialization for bool field in DuckLakeColumnInfo

carlopi:
This should be trivial, remove some run-time assertion when run with relassert.


Mytherin:
Thanks

PR #24, opened 2025-05-27
Bump to most recent of extension-ci-tools

carlopi:
This makes so that latest unittester improvements (in extension-ci-tools) will now be available by default when cloning the repository (and it's two submodules).


Mytherin:
Thanks

PR #122, opened 2025-06-02
Fix version handling for uncommitted changes

philippmd:
Similarly to DuckTransaction / DuckTransactionManager, ducklake should also have a notion of versioning uncommitted schema changes.

With this PR, we track uncommitted changes in the DucklakeTransactionManager, giving them unique ascending IDs.

The catalog version of a Ducklake Transaction if it has uncommitted changes is this unique ID, otherwise it's the catalog version.
We assume that catalog versions will always stay below 2^62

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -18,7 +18,7 @@ namespace duckdb {
 DuckLakeTransaction::DuckLakeTransaction(DuckLakeCatalog &ducklake_catalog, TransactionManager &manager,
                                          ClientContext &context)
     : Transaction(manager, context), ducklake_catalog(ducklake_catalog), db(*context.db),
-      local_catalog_id(DuckLakeConstants::TRANSACTION_LOCAL_ID_START) {
+      local_catalog_id(DuckLakeConstants::TRANSACTION_LOCAL_ID_START), catalog_version(0) {
 	metadata_manager = make_uniq<DuckLakeMetadataManager>(*this);
 }
 
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -189,7 +189,7 @@ private:
 	//! Snapshot cache for the AT (...) conditions that are referenced in the transaction
 	value_map_t<DuckLakeSnapshot> snapshot_cache;
 
-	atomic<idx_t> catalog_version = 0;
+	atomic<idx_t> catalog_version;
 };
 
 } // namespace duckdb


src/include/storage/ducklake_transaction_manager.hpp
```diff
@@ -24,10 +25,19 @@ class DuckLakeTransactionManager : public TransactionManager {
...
+	//! If there are no uncommitted changes, this is the schema version of the snapshot.
+	//! Otherwise, an id that is incremented whenever the schema changes (not stored between restarts)
+	idx_t GetCatalogVersion(Transaction &transaction);
+
 private:
 	DuckLakeCatalog &ducklake_catalog;
 	mutex transaction_lock;
 	reference_map_t<Transaction, shared_ptr<DuckLakeTransaction>> transactions;
+
+	atomic<idx_t> last_uncommitted_catalog_version = {TRANSACTION_ID_START};
```
Mytherin: Can we track this in the `DuckLakeCatalog` instead of in the transaction manager? That seems to be where this is used/needed.

src/include/storage/ducklake_transaction_manager.hpp
```diff
@@ -24,10 +25,19 @@ class DuckLakeTransactionManager : public TransactionManager {
...
+
+	friend DuckLakeTransaction;
```
Mytherin: Can we avoid using friend classes and expose a method instead?

src/storage/ducklake_catalog.cpp
```diff
@@ -1,21 +1,23 @@
...
-#include "duckdb/parser/constraints/not_null_constraint.hpp"
-#include "duckdb/common/types/uuid.hpp"
+#include "duckdb/storage/database_size.hpp"
+#include "storage/ducklake_initializer.hpp"
+#include "storage/ducklake_schema_entry.hpp"
+#include "storage/ducklake_table_entry.hpp"
+#include "storage/ducklake_transaction.hpp"
+#include "storage/ducklake_view_entry.hpp"
+
+#include <duckdb/transaction/duck_transaction_manager.hpp>
```
Mytherin: nit: this is the wrong include style, we use angle brackets only for system headers

philippmd: Thanks, fixed it

src/include/storage/ducklake_transaction_manager.hpp
philippmd: Done.

src/include/storage/ducklake_transaction_manager.hpp
philippmd: Moved it - Originally I tried to build a similar structure like `DuckTransactionManager` already has.

With the move we need to ensure that CatalogEntry.Parent is always set and a DuckLake catalog. Looking through the code I think that is the case.

src/storage/ducklake_transaction.cpp
```diff
@@ -1598,6 +1600,7 @@ DuckLakeTransaction &DuckLakeTransaction::Get(ClientContext &context, Catalog &c
 }
 
 void DuckLakeTransaction::CreateEntry(unique_ptr<CatalogEntry> entry) {
+	catalog_version = entry->ParentCatalog().Cast<DuckLakeCatalog>().GetNewUncommittedCatalogVersion();
```
Mytherin: `DuckLakeTransaction` has a member `DuckLakeCatalog &ducklake_catalog` - no need to get it from the passed in entry

src/storage/ducklake_catalog.cpp
```diff
@@ -480,8 +480,11 @@ void DuckLakeCatalog::OnDetach(ClientContext &context) {
 }
 
 optional_idx DuckLakeCatalog::GetCatalogVersion(ClientContext &context) {
-	auto &transaction = DuckLakeTransaction::Get(context, *this);
-	return transaction.GetSnapshot().schema_version;
+	DuckLakeTransactionManager &transaction_manager =
+	    static_cast<DuckLakeTransactionManager &>(DuckLakeTransactionManager::Get(db));
+	auto transaction = GetCatalogTransaction(context);
```
Mytherin: Can we just use `DuckLakeTransaction::Get(context, *this);` here?

src/storage/ducklake_transaction_manager.cpp
```diff
@@ -34,4 +34,12 @@ void DuckLakeTransactionManager::RollbackTransaction(Transaction &transaction) {
 	transactions.erase(transaction);
 }
 
+idx_t DuckLakeTransactionManager::GetCatalogVersion(Transaction &transaction_p) {
+	auto &transaction = transaction_p.Cast<DuckLakeTransaction>();
+	if (transaction.catalog_version > 0) {
```
Mytherin: I think this function is better suited to live in the `DuckLakeTransaction` after these changes

Mytherin (commented):
Thanks! Some more comments then I think this is good to go.

src/storage/ducklake_transaction.cpp
philippmd: Thanks, that is much better indeed

src/storage/ducklake_catalog.cpp
philippmd: Done

src/storage/ducklake_transaction_manager.cpp
philippmd: Done

src/storage/ducklake_catalog.cpp
```diff
@@ -480,8 +480,9 @@ void DuckLakeCatalog::OnDetach(ClientContext &context) {
 }
 
 optional_idx DuckLakeCatalog::GetCatalogVersion(ClientContext &context) {
-	auto &transaction = DuckLakeTransaction::Get(context, *this);
-	return transaction.GetSnapshot().schema_version;
+	DuckLakeTransactionManager &transaction_manager =
```
Mytherin: last nit: this can go now

philippmd: Done ;)

src/include/storage/ducklake_catalog.hpp
```diff
@@ -125,6 +129,8 @@ class DuckLakeCatalog : public Catalog {
 	DuckLakeOptions options;
 	// The path separator
 	string separator = "/";
+	//! A unique tracker for catalog changes in uncommitted transactions.
+	atomic<idx_t> last_uncommitted_catalog_version = {TRANSACTION_ID_START};
```
Mytherin: This is causing CI to fail - can we move this initialization to the constructor?

philippmd: Interesting.. I moved it :)

PR #126, opened 2025-06-02
Add support for schema + table sub-paths, and make data paths relative to the schema + table paths by default

Mytherin:
Implements https://github.com/duckdb/ducklake/discussions/55

This PR reworks the way that paths are chosen for data files. Previously data files would be relative only to the global catalog data path. This PR adds two extra layers:

* `ducklake_schema` now has the `path` and `path_is_relative` fields, if `path_is_relative` is set to true, this path is relative to the global data path (`DATA_PATH`), otherwise it is an absolute path
* `ducklake_table` now has the `path` and `path_is_relative` fields, if `path_is_relative` is set to true, this path is relative to the schema path
* Relative paths in `ducklake_data_file` and `ducklake_delete_file` are now relative to the path in their corresponding table, instead of relative to the global data path

By default the schema path is set to `{schema_name}/` for alphanumeric names, and `{schema_uuid}/` for names containing special characters. Similarly, the table path is set to `{table_name}/` for alphanumeric names, and `{table_uuid}/` for table names containing special characters.

For example:

```sql
ATTACH 'ducklake:path.db' AS ducklake;
-- default schema
CREATE TABLE ducklake.my_table(i INTEGER);
INSERT INTO ducklake.my_table VALUES (42);

-- new schema
CREATE SCHEMA ducklake.my_schema;
CREATE TABLE ducklake.my_schema.other_table(i INTEGER);
INSERT INTO ducklake.my_schema.other_table VALUES (84);

-- special characters in schema/table name
CREATE SCHEMA ducklake."schema/with/special/characters";
CREATE TABLE ducklake."schema/with/special/characters"."/table/"(i INTEGER);
INSERT INTO ducklake."schema/with/special/characters"."/table/" VALUES (84);
```

This will create the following files:

```
main/
	my_table/
		ducklake-{uuid}.parquet

my_schema/
	other_table/
		ducklake-{uuid}.parquet

{schema_uuid}/
	{table_uuid}/
		ducklake-{uuid}.parquet
```

The path for a schema/table is set once and then kept, even if the table is renamed. 

### DuckLake v0.2

This PR also bumps DuckLake to v0.2 as it introduces a change to the DuckLake tables. When attaching to a DuckLake v0.1 database - the DuckLake is automatically upgraded to DuckLake v0.2 through the following `ALTER` statements:

```sql
ALTER TABLE ducklake_schema ADD COLUMN path VARCHAR DEFAULT '';
ALTER TABLE ducklake_schema ADD COLUMN path_is_relative BOOLEAN DEFAULT TRUE;
ALTER TABLE ducklake_table ADD COLUMN path VARCHAR DEFAULT '';
ALTER TABLE ducklake_table ADD COLUMN path_is_relative BOOLEAN DEFAULT TRUE;
```


jordandakota:
Would using these paths instead of the old global remove some of the trustless security where you've encrypted the files and have the metadata server in a different trust zone? The schema and table names would now be exposed? Still impossible to use the data, but still exposed.

Mytherin:
Yes, you are correct that using these paths would leak schema and table names in the directory structure. Note that using these paths is not required - the old behavior can still be obtained by setting the table and schema paths to the empty string with `path_is_relative = true`. I will add some more documentation around paths soon.

src/storage/ducklake_schema_entry.cpp
```diff
@@ -48,29 +50,39 @@ bool DuckLakeSchemaEntry::HandleCreateConflict(CatalogTransaction transaction, C
...
 	return result;
 }
 
+optional_ptr<CatalogEntry> DuckLakeSchemaEntry::CreateTable(CatalogTransaction transaction,
+                                                            BoundCreateTableInfo &info) {
+	auto &duck_transaction = transaction.transaction->Cast<DuckLakeTransaction>();
+	auto &duck_catalog = catalog.Cast<DuckLakeCatalog>();
+	auto &base_info = info.Base();
+	auto table_uuid = duck_transaction.GenerateUUID();
+	auto table_data_path = DataPath() + duck_catalog.GeneratePathFromName(table_uuid, base_info.table);
```
Tishj, line 82: Isn't there an issue with renames of the table lurking here?
What happens when we rename a table and then add a new table with the same name

I guess those now occupy the same folder?
If we want to enable table-level authorization, we should make sure two tables won't use the same folder right?
So this should instead always use the table_uuid / schema_uuid I think?

PR #140, opened 2025-06-04
Add options function

lonless9:
Add ducklake_options() table function for retrieving configuration options.

`option_name VARCHAR, description VARCHAR, value VARCHAR, scope VARCHAR, scope_entry VARCHAR`


src/functions/ducklake_get_option.cpp
```diff
@@ -0,0 +1,284 @@
...
+static constexpr const char *OPTION_TYPE_USER = "user";
+static constexpr const char *OPTION_TYPE_SYSTEM = "system";
+static constexpr const char *DEFAULT_OPTION_DESCRIPTION = "Custom configuration option";
+
+static constexpr const char *COLUMN_NAME = "name";
+static constexpr const char *COLUMN_VALUE = "value";
+static constexpr const char *COLUMN_TYPE = "type";
+static constexpr const char *COLUMN_DESCRIPTION = "description";
+
+static const std::unordered_map<string, OptionMetadata> DUCKLAKE_OPTIONS = {
```
Mytherin: In general we avoid static C++ objects because of destructor order - can we turn this into a C-style array instead (see e.g. `src/common/ducklake_types.cpp`)?

src/functions/ducklake_get_option.cpp
```diff
@@ -0,0 +1,284 @@
...
+struct OptionMetadata {
+	string description;
+	string type;
+};
+
+static constexpr const char *OPTION_TYPE_USER = "user";
+static constexpr const char *OPTION_TYPE_SYSTEM = "system";
+static constexpr const char *DEFAULT_OPTION_DESCRIPTION = "Custom configuration option";
+
+static constexpr const char *COLUMN_NAME = "name";
```
Mytherin: This being static isn't done elsewhere in the code - perhaps we can just inline it for consistency?

Mytherin (commented):
Thanks for the PR!

Instead of having a function `get_option()`, perhaps we can add a single function `options()` that returns all available options. When retrieving options - having the filtering done at the function level is not particularly useful in my opinion. If we want to obtain a specific option - we can always just run `SELECT * FROM ducklake.options() WHERE name='...'`.

My suggestion here would be to have the function return the following schema:

```
option_name VARCHAR, description VARCHAR, value VARCHAR, scope VARCHAR, scope_entry VARCHAR
```

`scope` is set to either `GLOBAL`, `SCHEMA` or `TABLE`. For `SCHEMA` and `TABLE`, `scope_entry` is set to the name of the schema or table. 

I've left some minor code comments below

lonless9:
Thanks! I've made the adjustment.

test/sql/settings/per_table_settings.test
```diff
@@ -106,3 +106,24 @@ statement ok
...
+SELECT option_name, value, scope FROM ducklake.options() WHERE option_name='parquet_compression' ORDER BY scope, value;
+----
+parquet_compression	uncompressed	GLOBAL
+parquet_compression	lz4	SCHEMA
+parquet_compression	gzip	TABLE
+parquet_compression	zstd	TABLE
+
+# error handling - non-existent table
+statement error
+SELECT * FROM ducklake.options(table_name => 'nonexistent_table')
```
Mytherin: This option is not supported so we shouldn't need these tests

test/sql/settings/per_table_settings.test
```diff
@@ -106,3 +106,24 @@ statement ok
 ATTACH 'ducklake:__TEST_DIR__/ducklake_per_table_settings.db' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_per_table_settings_files');
 
 endloop
+
+# options retrieval - check that all options are present with correct values and scopes
+query III
+SELECT option_name, value, scope FROM ducklake.options() WHERE option_name='parquet_compression' ORDER BY scope, value;
```
Mytherin: Can we check `scope_entry` here as well?

src/functions/ducklake_options.cpp
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	// Schema options
+	for (auto &schema_setting : metadata.schema_settings) {
+		DuckLakeOptionInfo option_info;
+		option_info.option_name = schema_setting.tag.key;
+		option_info.value = schema_setting.tag.value;
+		option_info.description = GetOptionDescription(schema_setting.tag.key);
+		option_info.scope = "SCHEMA";
+		option_info.scope_entry = metadata_manager.GetPathForSchema(schema_setting.schema_id);
```
Mytherin: This returns the path - we should return the name instead. You can use `DuckLakeCatalog::GetEntryById` to get the catalog entry which has the name (passing in `transaction.GetSnapshot()`).

Mytherin (commented):
Thanks! Looks good - some more comments then this is good to go

Mytherin:
Thanks!

PR #175, opened 2025-06-08
Add support for adding external files to DuckLake using the `ducklake_add_data_files` method

Mytherin:
This PR adds support for adding Parquet files to tables in DuckLake directly through the `ducklake_add_data_files` method.

For example:

```sql
ATTACH 'ducklake:my_ducklake.db' AS my_ducklake;
CREATE TABLE my_ducklake.people(id BIGINT, name VARCHAR);
COPY (SELECT 42 AS id, 'Mark' AS name) TO 'people.parquet';

CALL ducklake_add_data_files('my_ducklake', 'people', 'people.parquet');

FROM my_ducklake.people;

┌───────┬─────────┐
│  id   │  name   │
│ int64 │ varchar │
├───────┼─────────┤
│  42   │ Mark    │
└───────┴─────────┘
```

The function only reads the metadata from the file in order to register the files in the DuckLake catalog directly. No data is moved or copied, and the Parquet file is left as-is in whichever location it is stored.

Note that ownership of the Parquet file is transferred to DuckLake. As such, compaction operations (such as those triggered through `merge_adjacent_files` or `expire_snapshots` followed by `cleanup_old_files`) can cause the file to be deleted by DuckLake.

#### Field ID Mapping

DuckLake normally uses **field ids** to perform column mapping. Every column has a unique field id. This field id is stored in the [`ducklake_column` table as the `column_id` column](https://ducklake.select/docs/stable/specification/tables/ducklake_column). In addition, every Parquet file that is written by DuckLake has the field ids written for every column in the `field id` field in the Parquet schema. When scanning, DuckLake maps the field ids as defined in the metadata to the field ids defined in the Parquet file.

#### Name Mapping

When adding files that are written outside of DuckLake, these field ids are not present. As such, we need to perform the column mapping in a different manner. This PR introduces support for this by adding the concept of *column mappings* that can be defined per-file. Column mappings define how the *source* columns in a file map to field ids in a table.

This works by adding two new tables, and a new column `mapping_id` to the `ducklake_data_file` table:

```sql
ALTER TABLE ducklake_data_file ADD COLUMN mapping_id BIGINT;
CREATE TABLE ducklake_column_mapping(mapping_id BIGINT, table_id BIGINT, type VARCHAR);
CREATE TABLE ducklake_name_mapping(mapping_id BIGINT, column_id BIGINT, source_name VARCHAR, target_field_id BIGINT, parent_column BIGINT);
```

For files that use the regular field id-based mapping, `mapping_id` is `NULL`. For files that are added through `ducklake_add_data_files`, the `mapping_id` refers to a mapping as defined in the `ducklake_column_mapping` table. The `ducklake_name_mapping` then has, for every source column name in the file, a `target_field_id`. When performing the mapping, we can thus read the column for `target_field_id` by reading the column of name `source_column`. This allows us to map the columns to the corresponding field ids without the field ids actually being present in the file. Here is an example of these tables for the example given above:

```sql
FROM ducklake_column_mapping;
┌────────────┬──────────┬─────────────┐
│ mapping_id │ table_id │    type     │
│   int64    │  int64   │   varchar   │
├────────────┼──────────┼─────────────┤
│     0      │    2     │ map_by_name │
└────────────┴──────────┴─────────────┘
FROM ducklake_name_mapping;
┌────────────┬───────────┬─────────────┬─────────────────┬───────────────┐
│ mapping_id │ column_id │ source_name │ target_field_id │ parent_column │
│   int64    │   int64   │   varchar   │      int64      │     int64     │
├────────────┼───────────┼─────────────┼─────────────────┼───────────────┤
│          0 │         0 │ id          │               1 │          NULL │
│          0 │         1 │ name        │               2 │          NULL │
└────────────┴───────────┴─────────────┴─────────────────┴───────────────┘
```

As the mappings are defined per-file and we map back to the standard field ids, there are no restrictions for files that are added to DuckLake tables in this manner:

* Mappings can be mixed. Files written by DuckLake will continue to use field id mappings, and files added through `ducklake_add_data_files` will use name mapping. Files can be added to any DuckLake table at any point with no restrictions. 
* Files added through `ducklake_add_data_files` support the same set of operations as files written natively through DuckLake - i.e. they support full schema evolution, time travel, data change feed, etc.

#### Missing Columns

When adding files to a table, all columns that are present in the table must be present in the Parquet file, otherwise an error is thrown. The `allow_missing` option can be used to add the file anyway - in which case any missing columns will be substituted with the `initial_default` value of the column.

```sql
COPY (SELECT 84 AS id) TO 'missing_name.parquet';

CALL ducklake_add_data_files('my_ducklake', 'people', 'missing_name.parquet');
-- Column "name" exists in table "people" but was not found in file "missing_name.parquet"
-- * Set allow_missing => true to allow missing fields and columns


CALL ducklake_add_data_files('my_ducklake', 'people', 'missing_name.parquet', allow_missing => true);

FROM my_ducklake.people;
┌───────┬─────────┐
│  id   │  name   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │ Mark    │
│    84 │ NULL    │
└───────┴─────────┘
```

#### Extra Columns

When adding files to a table, if there are any columns present that are not present in the table, an error is thrown by default. The `ignore_extra_columns` option can be used to add the file anyway - any extra columns will be ignored and unreachable.

```sql
COPY (SELECT 100 AS id, 'Hannes' AS name, 'Duck' AS pet) TO 'extra_column.parquet';

CALL ducklake_add_data_files('my_ducklake', 'people', 'extra_column.parquet');
-- Column "pet" exists in file "extra_column.parquet" but was not found in table "people"
-- * Set ignore_extra_columns => true to add the file anyway


CALL ducklake_add_data_files('my_ducklake', 'people', 'extra_column.parquet', ignore_extra_columns => true);

FROM my_ducklake.people;
┌───────┬─────────┐
│  id   │  name   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │ Mark    │
│    84 │ NULL    │
│   100 │ Hannes  │
└───────┴─────────┘
```

### Type Mapping

In general, types of columns in the source Parquet file must match the type as defined in the table, otherwise an error is thrown. Types in the Parquet file can be narrower than the type defined in the table. Below is a supported mapping type:

|  Table Type   |          Supported Parquet Types          |
|---------------|-------------------------------------------|
| bool          | bool                                      |
| int8          | int8                                      |
| int16         | int[8/16], uint8                          |
| int32         | int[8/16/32], uint[8/16]                  |
| int64         | int[8/16/32/64], uint[8/16/32]            |
| uint8         | uint8                                     |
| uint16        | uint[8/16]                                |
| uint32        | uint[8/16/32]                             |
| uint64        | uint[8/16/32/64]                          |
| float         | float                                     |
| double        | float/double                              |
| decimal(P, S) | decimal(P',S'), where P' <= P and S' <= S |
| blob          | blob                                      |
| varchar       | varchar                                   |
| date          | date                                      |
| time          | time                                      |
| timestamp     | timestamp, timestamp_ns                   |
| timestamp_ns  | timestamp, timestamp_ns                   |
| timestamptz   | timestamptz                               |

For nested types, the nested type and all of the child types must match.

Struct columns map like top-level columns: all fields must be present unless `allow_missing` is set to true, and no extra fields are permitted unless `ignore_extra_columns` is set to true.


khepin:
This doesn't look like it correctly populates the `ducklake_file_column_statistics` table and both min and max values are left null, which prevents evaluating the predicates at that level.

Mytherin:
They are correctly populated when using Parquet files written by DuckDB, per the example above:

```sql
ATTACH 'ducklake:my_ducklake.db' AS my_ducklake;
CREATE TABLE my_ducklake.people(id BIGINT, name VARCHAR);
COPY (SELECT 42 AS id, 'Mark' AS name) TO 'people.parquet';
CALL ducklake_add_data_files('my_ducklake', 'people', 'people.parquet');
FROM __ducklake_metadata_my_ducklake.ducklake_file_column_statistics;
┌──────────────┬──────────┬───────────┬───────────────────┬─────────────┬────────────┬───────────┬───────────┬──────────────┐
│ data_file_id │ table_id │ column_id │ column_size_bytes │ value_count │ null_count │ min_value │ max_value │ contains_nan │
│    int64     │  int64   │   int64   │       int64       │    int64    │   int64    │  varchar  │  varchar  │   boolean    │
├──────────────┼──────────┼───────────┼───────────────────┼─────────────┼────────────┼───────────┼───────────┼──────────────┤
│            1 │        1 │         1 │                47 │           1 │          0 │ 42        │ 42        │ NULL         │
│            1 │        1 │         2 │                51 │           1 │          0 │ Mark      │ Mark      │ NULL         │
└──────────────┴──────────┴───────────┴───────────────────┴─────────────┴────────────┴───────────┴───────────┴──────────────┘

```

The stats are read from the Parquet files directly. Perhaps the tool you are using to write the Parquet files is not writing stats?

khepin:
Oh I'm so sorry, I left this comment and went to investigate some more then forgot to come back here.
This was indeed an issue with the parquet exporter I am using. Found a solution there too.

Thanks!!

yan-hic:
Would be nice to have an option to create the table if not exists using the schemas of the underlying parquet files. Currently, this requires creating an empty table shell like so:
```
COPY (SELECT 42 AS id, 'Mark' AS name) TO 'people.parquet';

CREATE TABLE my_ducklake.people AS FROM 'people.parquet' LIMIT 0;
CALL ducklake_add_data_files('my_ducklake', 'people', 'people.parquet');

FROM my_ducklake.people;
```

MiguelElGallo:
> Oh I'm so sorry, I left this comment and went to investigate some more then forgot to come back here. This was indeed an issue with the parquet exporter I am using. Found a solution there too.
> 
> Thanks!!

Hi,
I have colleagues, asking why parquet files do perform badly sometimes. I decided to created a simple tool that shows metadata about a parquet file. 

`uvx --refresh iparq inspect dummy.Parquet`

Hope it helps... And sorry for spamming here :)

PR #177, opened 2025-06-09
make max_retry_count configurable

jayhan94:
make transaction's `max_retry_count` configurable


Mytherin:
Thanks for the PR!

`ducklake_set_option` is specifically designed for options that are persisted in the catalog. Since this option is not persisted in the catalog, perhaps we can make it a regular configuration option instead? See e.g. [the postgres extension](https://github.com/duckdb/duckdb-postgres/blob/main/src/postgres_extension.cpp#L162) for how this would work.

jayhan94:
> Thanks for the PR!
> 
> `ducklake_set_option` is specifically designed for options that are persisted in the catalog. Since this option is not persisted in the catalog, perhaps we can make it a regular configuration option instead? See e.g. [the postgres extension](https://github.com/duckdb/duckdb-postgres/blob/main/src/postgres_extension.cpp#L162) for how this would work.

@Mytherin Thank you for the help! I'll take a look.

src/ducklake_extension.cpp
```diff
@@ -17,6 +17,9 @@ static void LoadInternal(DatabaseInstance &instance) {
 	auto &config = DBConfig::GetConfig(instance);
 	config.storage_extensions["ducklake"] = make_uniq<DuckLakeStorageExtension>();
 
+	config.AddExtensionOption("ducklake_txn_max_retry_count", "The maximum amount of retry attempts for a ducklake transaction",
```
Mytherin: Can we rename this to `ducklake_max_retry_count`?

Mytherin (commented):
Thanks! Looks good - one minor comment

src/ducklake_extension.cpp
jayhan94: Done

jayhan94:
> Thanks! Looks good - one minor comment

@Mytherin Appreciate the quick reply - I've followed your suggestion.

Mytherin:
Thanks - can we add a test as well?

jayhan94:
I encountered an issue. I don't know how to implement transaction retries, and I also don't know how to obtain the number of transaction retry attempts to verify that it actually retried up to the max_retry_count.

Mytherin:
The concurrent test (`test/sql/concurrent/concurrent_insert_conflict.test`) should trigger retries. Perhaps you can increase the concurrency there to e.g. 5, and then set the max retry count higher to ensure it still passes?

src/storage/ducklake_transaction.cpp
```diff
@@ -1300,8 +1301,15 @@ void DuckLakeTransaction::FlushChanges() {
...
 		return;
 	}
-	// FIXME: this should probably be configurable
+
 	idx_t max_retry_count = 5;
+	Value max_retry_count_value;
+	auto context_ref = context.lock();
+	auto lookup_result = context_ref->TryGetCurrentSetting("ducklake_max_retry_count", max_retry_count_value);
+	if (lookup_result) {
+		max_retry_count = std::max((idx_t)1, max_retry_count_value.GetValue<idx_t>());
```
Mytherin: Maybe this should be `max_retry_count_value.GetValue<idx_t>() + 1` instead? Given our first try is a "retry" - I would say `max_retry_count` should be the actual number of retries we attempt. Currently `max_retry_count = 0` and `max_retry_count = 1` have the exact same behavior.

jayhan94: Would it be more accurate to use `for (idx_t i = 0; i < max_retry_count + 1; i++)`? This makes the distinction between retries and initial try more explicit.

Mytherin: Also fine by me

jayhan94:
@Mytherin PTAL

Mytherin:
Thanks!

PR #201, opened 2025-06-11
Add support for directly attaching DuckLake

Mytherin:
Building on top of https://github.com/duckdb/duckdb/pull/17868, this PR adds support for directly attaching DuckLake catalogs while opening DuckDB, i.e. this now works:

```bash
duckdb ducklake:my_file.db
```


djouallah:
@Mytherin how do you pass data_path here ?

Mytherin:
You can't pass a data path as a parameter directly in this way - but you only need to pass data_path when creating a new DuckLake. This is currently mostly useful for connecting to an existing DuckLake.

PR #202, opened 2025-06-11
Add support for the year/month/day/hour partitioning transforms

Mytherin:
This PR adds support for the `year / month / day / hour` partitioning transforms, allowing partitions to be defined on subsets of a timestamp using these partitioning transforms, e.g.:

```sql
CREATE TABLE partitioned_tbl(ts TIMESTAMP, values VARCHAR);
ALTER TABLE partitioned_tbl SET PARTITIONED BY (year(ts), month(ts));
```

### DIFF
--- a/test/sql/partitioning/partition_tpch.test_slow
+++ b/test/sql/partitioning/partition_tpch.test_slow
@@ -47,6 +47,18 @@ SELECT COUNT(*) FROM glob('__TEST_DIR__/ducklake_partition_tpch/main/part/**')
 statement ok
 USE ducklake
 
+# verify partitions are used
+query II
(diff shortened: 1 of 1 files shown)


yan-hic:
Thinking about that, this would (still) require the year/month predicate on SELECT i.e. 
```
FROM partitioned_tbl WHERE ts >= '2025-06-01' AND year >= 2025 -- more complex filter on month

full scan:
FROM partitioned_tbl WHERE ts >= '2025-06-01'
```
Ideally the partitions keep a reference to the originating fieldname, and a dynamic file filter is built if filtered thereon i.e. the second SQL reads only the relevant partitions

Mytherin:
That should already work - the second SQL query would only read the relevant partitions because the `min/max` on the `ts` is automatically set to only the values that occur in the file, which in case of partitioning by year/month would be only the relevant year/month for the partition.

PR #226, opened 2025-06-18
[chore] Bump to 1.3.1 and reenable publishing

carlopi:
Changes:
* duckdb 1.3.1 both as submodule and as CI target
* publishing extension on 1.3.1
* skip Windows test currently failing (it's independent, and I can't make a call whether problem is test side or need fixing for real)

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -20,22 +20,22 @@ jobs:
       ci_tools_version: main
       extension_name: ducklake
 
-#  duckdb-stable-build:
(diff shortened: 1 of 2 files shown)


djouallah:
@carlopi it does crash on windows using python but works fine using the cli and the UI for that matter

Mytherin:
@djouallah can you expand on what you mean? Could you file an issue?

I think the test failure is unrelated - it's caused by the local file system being more strict on Windows.

djouallah:
@Mytherin sorry for the noise, I think I messed up my python environment, it is not related to ducklake

PR #247, opened 2025-06-24
Allow compaction of non-adjacent files using `ducklake_merge_adjacent_files`

Mytherin:
This makes the function a bit of a misnomer, but oh well :)

This PR allows `ducklake_merge_adjacent_files` to merge files that are non-adjacent, i.e. that have row-ids that are disjoint. 

Previously compaction operations would be limited to merging files with adjacent row-ids. This works well when we have standard sequential inserts, but can be limiting in particular with non-temporal partitioning where files might be spread across partitions in non-adjacent ways. For example, we might first insert into `partition 1`, then insert into `partition 2`, then insert into `partition 1` again. 

By writing the row ids to the file instead of relying on `row_id_start`, we can compact together files that are non-adjacent as well without loss of information: the row identifiers will be written to the file on compaction, and read from the file during reads. This PR also enables the `row_id_start` to be set to `NULL` in the metadata layer - signaling the row ids should be read from the file instead. 

Support for reading and writing row-ids to files already existed as part of the `UPDATE` infrastructure - so this PR is not a very big lift.

### DIFF
--- a/src/functions/ducklake_merge_adjacent_files.cpp
+++ b/src/functions/ducklake_merge_adjacent_files.cpp
@@ -153,6 +153,41 @@ DuckLakeCompactor::DuckLakeCompactor(ClientContext &context, DuckLakeCatalog &ca
     : context(context), catalog(catalog), transaction(transaction), binder(binder), table_id(table_id) {
 }
 
+struct DuckLakeCompactionCandidates {
+	vector<idx_t> candidate_files;
+};
+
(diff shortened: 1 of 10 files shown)


djouallah:
@Mytherin do you have any plan to compact deletion files too please, currently I am rewriting the table to do that which is not very efficient

Mytherin:
Yes, it is on the roadmap but that is more work

PR #248, opened 2025-06-24
Add `CREATE_IF_NOT_EXISTS` attach option (which defaults to true)

Mytherin:
Usage:

```sql
ATTACH 'ducklake:new_ducklake.db' (CREATE_IF_NOT_EXISTS false);
-- Existing DuckLake at metadata catalog "new_ducklake.db" does not exist - and creating a new DuckLake is explicitly disabled
```

### DIFF
--- a/src/include/common/ducklake_options.hpp
+++ b/src/include/common/ducklake_options.hpp
@@ -27,6 +27,7 @@ struct DuckLakeOptions {
 	string data_path;
 	AccessMode access_mode = AccessMode::AUTOMATIC;
 	DuckLakeEncryption encryption = DuckLakeEncryption::AUTOMATIC;
+	bool create_if_not_exists = true;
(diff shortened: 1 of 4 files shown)


carlopi:
I see this for ducklake right now solves a problem, should it also be considered for duckdb/duckb to support `ATTACH IF EXISTS 'ducklake:new_ducklake.db'` (that will not create a new one) ?

Mytherin:
`IF EXISTS` already exists and is about catalog objects - not about the underlying database, so that is not a good fit. It works with DuckLake and DuckDB files, but does not make sense with e.g. a Postgres connection.

ywelsch:
Would it make sense to have CREATE_IF_NOT_EXISTS == false when attaching a ducklake in read-only mode? Right now it creates the ducklake schema in the target database (even though it's supposed to be read-only access)

PR #252, opened 2025-06-26
Add support for adding hive partitioned files using `ducklake_add_data_files`

Mytherin:
Follow-up from https://github.com/duckdb/ducklake/pull/175

This PR adds support for adding hive partitioned files using `ducklake_add_data_files`. Hive partitioned files are "special" because they have columns that are not present in the files themselves but only in the file path (e.g. `date=1992-01-01/file.parquet`).

This PR solves that by adding a new field - `is_partition` - to the columns in `ducklake_name_mapping`. When this field is set to `true`, the column is not read from the file but instead read from the file name. 

Whether or not files are hive partitioned is automatically detected by `ducklake_add_data_files` by default. It can be disabled using `hive_partitioning => false`, e.g.:

```sql
-- explicitly disable hive partitioning
CALL ducklake_add_data_files('ducklake', 'test', 'files/**/*.parquet', hive_partitioning => false);
```

This PR moves the DuckLake version to `v0.3-dev1` and adds migration from `v0.2 -> v0.3` - as such we will not merge this (yet) and this will likely only land together with DuckDB v1.4.

### DIFF
--- a/src/common/ducklake_name_map.cpp
+++ b/src/common/ducklake_name_map.cpp
@@ -41,6 +41,9 @@ bool DuckLakeNameMapEntry::IsCompatibleWith(const DuckLakeNameMapEntry &other) c
 	if (target_field_id != other.target_field_id) {
 		return false;
 	}
+	if (hive_partition != other.hive_partition) {
+		return false;
+	}
 	return ListIsCompatible(child_entries, other.child_entries);
 }
 
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -7,6 +7,8 @@
 
 namespace duckdb {
 
+enum class HivePartitioningType { AUTOMATIC, YES, NO };
+
 struct DuckLakeAddDataFilesData : public TableFunctionData {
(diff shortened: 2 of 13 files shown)


yan-hic:
@Mytherin will that cover the following use case: https://github.com/duckdb/duckdb/discussions/14422#discussioncomment-13671311 ?

i.e. I have tried
```
call ducklake_add_data_files('my_ducklake','my_new_table', 'gs://my-bucket/my-folder/year=*/xtr=*/*.parquet')
```
but am getting `Table Function with name ducklake_add_data_files does not exist!`. If only available in 0.2, how can i force to load the latest version ? `LOAD ducklake` always shows 0.1 (as per ducklake_metadata).

Mytherin:
Are you using the latest version of DuckDB (v1.3.1)?

yan-hic:
my bad - was on 1.3.0. 
1.3.1 does indeed autoload/use ducklake 0.2

The "error" is now that the table must exist - sounds more a limitation. I have added https://github.com/duckdb/ducklake/pull/175#issuecomment-3040337282 or is it more appropriate to create a separate issue/request for that ? 
That would address altogether the external table discussion in my mind.

PR #287, opened 2025-07-14
Add `lz4_raw` parquet compression as one of the supported compression algorithms

gahtan-syarif:
lz4_raw is part of both the [parquet spec](https://parquet.apache.org/docs/file-format/data-pages/compression/) and also [available in duckdb](https://github.com/duckdb/duckdb/blob/4b9b91aa88dfaed7244b0965bb127642cfd07c18/extension/parquet/parquet_extension.cpp#L318)


Mytherin:
Thanks!

PR #289, opened 2025-07-15
Fix #284: correctly set table id of newly written files when resolving conflicts between separate transactions

### LINKED ISSUE #284: Creating 2 tables at the same time on 2 different threads will lead to errors in data from one table being loaded into the other table.
If the structures of the 2 tables are not the same, there will be errors in data not matching the column.

Error description:
Failed to commit: Failed to commit DuckLake transaction: Failed to cast value: Could not convert string " " to DECIMAL(12,0)

Environment:
- Metadata catalog: PostgreSQL
- OS: Windows

Mytherin:
Fixes #284 

While writing a new table to DuckLake, we pick a table id for that table during the commit stage, and then try to commit. If we need to retry, the table id of that table will change. In the retry code, there was an issue where we would update the table id of the table in the retry correctly - but would leave any data associated with that transaction-local table stuck on the first table id picked. If two transactions performed a `CREATE TABLE AS` that would conflict this could cause data files from one table to erroneously point to the other table.

This PR fixes this issue by, instead of modifying the table id of the table in-place with the new table-id, constructing a map of `transaction_local_id -> final_id` and then clearing that map upon retry.

### DIFF
--- a/src/include/storage/ducklake_catalog_set.hpp
+++ b/src/include/storage/ducklake_catalog_set.hpp
@@ -34,8 +34,6 @@ public:
 	optional_ptr<CatalogEntry> GetEntryById(SchemaIndex index);
 	optional_ptr<CatalogEntry> GetEntryById(TableIndex index);
 	void AddEntry(DuckLakeSchemaEntry &schema, TableIndex id, unique_ptr<CatalogEntry> entry);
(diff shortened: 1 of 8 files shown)


steven-luabase:
@Mytherin is this fix available in the core version of ducklake or do I need to install from core-nightly to pull this in?

Issue #69, opened 2025-05-28
DROP TABLE on partitioned table causes "Invalid Input Error" on all subsequent queries


csmithxc:
Setup:
- Postgres catalog (Supabase, Postgres 15.8.1)
- S3 data path
- DuckDB 1.3.0 via CLI

Reproduce:
```sql
CREATE TABLE IF NOT EXISTS test_update AS (SELECT 1 as id, 'hello' as greeting, 60 as amount) WITH NO DATA;
INSERT INTO test_update SELECT 1 as id, 'hello' as greeting, 60 as amount;
ALTER TABLE test_update SET PARTITIONED BY (greeting);

DROP TABLE test_update;

-- Can be literally any query
SHOW ALL TABLES;
```

Resulting error:
- `Invalid Input Error: Could not find matching table for partition entry`

Mytherin:
Thanks for the report! I've pushed a fix in https://github.com/duckdb/ducklake/pull/75

PR #317, opened 2025-07-24
Set `end_snapshot` on all metadata entries when dropping table

gahtan-syarif:
To prevent issues similar to https://github.com/duckdb/ducklake/issues/69 , especially when working with other engines. Also fixes issue where table renames causes partition info to be dropped.

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1156,22 +1156,16 @@ struct CompactionInformation {
 void DuckLakeTransaction::CommitChanges(DuckLakeCommitState &commit_state,
                                         TransactionChangeInformation &transaction_changes) {
 	auto &commit_snapshot = commit_state.commit_snapshot;
-
-	for (const auto &val : dropped_tables) {
-	    if (renamed_tables.find(val) == renamed_tables.end()) {
-	        real_dropped_tables.insert(val);
-	    }
-	}
 	
 	// drop entries
 	if (!dropped_tables.empty()) {
-		if (!renamed_tables.empty()){
-			metadata_manager->DropTables(commit_snapshot, dropped_tables, true);
-		} 
-		if (!real_dropped_tables.empty()) {
-			metadata_manager->DropTables(commit_snapshot, dropped_tables, false);
-		}
+		metadata_manager->DropTables(commit_snapshot, dropped_tables, true);
 	}
+
+	if (!renamed_tables.empty()){
+		metadata_manager->DropTables(commit_snapshot, renamed_tables, true);
+	} 
+
 	if (!dropped_views.empty()) {
 		metadata_manager->DropViews(commit_snapshot, dropped_views);
 	}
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks! Could you just merge with main so the tests can run again?

gahtan-syarif:
> Thanks! Could you just merge with main so the tests can run again?

Sure 👍

gahtan-syarif:
It failed on the rename table test,
Does ducklake table renames involve dropping the tables?

Mytherin:
A rename is a drop + insert in the `ducklake_table` table - maybe that is now also dropping other stuff incorrectly. This might already have been a problem with the partition info but that might have gone unnoticed.

gahtan-syarif:
Yeah, i guess this would mean renames currently erases partition info

gahtan-syarif:
@Mytherin i fixed the issue, PR is ready, but perhaps you have a better idea in mind?

src/storage/ducklake_transaction.cpp
```diff
@@ -1863,6 +1875,7 @@ void DuckLakeTransaction::AlterEntryInternal(DuckLakeTableEntry &table, unique_p
 			// table is not transaction local - add to drop list
 			auto table_id = table.GetTableId();
 			dropped_tables.insert(table_id);
+			renamed_tables.insert(table_id);
```
Mytherin, line 1871: Can we perhaps insert renamed tables only into the `renamed_tables` set, instead of in both the `dropped_tables` and `renamed_tables` sets?

src/storage/ducklake_transaction.cpp
```diff
@@ -1156,9 +1156,21 @@ struct CompactionInformation {
...
+	    if (renamed_tables.find(val) == renamed_tables.end()) {
+	        real_dropped_tables.insert(val);
+	    }
+	}
+	
 	// drop entries
 	if (!dropped_tables.empty()) {
-		metadata_manager->DropTables(commit_snapshot, dropped_tables);
+		if (!renamed_tables.empty()){
+			metadata_manager->DropTables(commit_snapshot, dropped_tables, true);
```
Mytherin: This seems like it won't work if we rename and drop a table in the same transaction - given we pass `dropped_tables` in both scenarios

Mytherin (commented):
Thanks for picking up the fix - I've left some comments. Could you also add a test that partition info is now preserved when renaming?

src/storage/ducklake_transaction.cpp
gahtan-syarif: ahh right, forgot to change the set, fixed 👍

src/storage/ducklake_transaction.cpp
gahtan-syarif: sure 👍 fixed

gahtan-syarif:
> Thanks for picking up the fix - I've left some comments. Could you also add a test that partition info is now preserved when renaming?

sure 👍 test added

Mytherin:
Thanks!

PR #319, opened 2025-07-25
Implement `migrate_if_required` Option

### LINKED ISSUE #304: Management Mode to Prevent Auto-Running Migrations
In a similar vein to the `READ_ONLY` and `CREATE_IF_NOT_EXISTS` Flags when Attaching DuckLake, it would be useful to have a `RUN_MIGRATIONS` Boolean Flag to prevent DuckLake from Auto-Migrating itself invisibly.

Perhaps instead it could just fail if it needs to migrate and RUN_MIGRATIONS is False.

ben-burwood:
Closes #304 

This throws on Attach if the Ducklake Extension requires Migrations and the `MIGRATE_IF_REQUIRED` Flag is False (default True).

This is critical for us as we have a Master Ducklake Metadata Catalog which we want to ensure is tightly managed and hence should not Invisibly Migrate. Particularly as we may have many different clients accessing this Catalog.

Not sold on the option name (perhaps `migrate_automatically` is clearer...) or implementation (particularly hardcoding so many version strings) so open to suggestions...

### DIFF
--- a/src/include/storage/ducklake_initializer.hpp
+++ b/src/include/storage/ducklake_initializer.hpp
@@ -24,7 +24,7 @@ public:
 
 private:
 	void InitializeNewDuckLake(DuckLakeTransaction &transaction, bool has_explicit_schema);
-	void LoadExistingDuckLake(DuckLakeTransaction &transaction);
+	void LoadExistingDuckLake(DuckLakeTransaction &transaction, bool migrate_if_required);
 	void InitializeDataPath();
 	string GetAttachOptions();
 	void CheckAndAutoloadedRequiredExtension(const string &pattern);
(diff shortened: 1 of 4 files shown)


src/include/storage/ducklake_initializer.hpp
```diff
@@ -24,7 +24,7 @@ class DuckLakeInitializer {
 
 private:
 	void InitializeNewDuckLake(DuckLakeTransaction &transaction, bool has_explicit_schema);
-	void LoadExistingDuckLake(DuckLakeTransaction &transaction);
+	void LoadExistingDuckLake(DuckLakeTransaction &transaction, bool migrate_if_required);
```
Mytherin: Can we add this to `DuckLakeOptions` instead?

Mytherin (commented):
Thanks! LGTM - one comment

Mytherin:
We can solve the hard-coding of version strings separately

src/include/storage/ducklake_initializer.hpp
ben-burwood: Duh! Not sure why I added the parameter

Mytherin:
Thanks!

PR #335, opened 2025-07-31
Add Author and Commit Message to Snapshot

pdet:
This PR adds support for including the author and commit message in a DuckLake snapshot.

It extends the `snapshot` table with two new columns: `author` and `commit_message`.

To include this information in a snapshot, the user must start a transaction and set the commit message before committing the transaction using `ducklake.set_commit_message({author}, {commit_message});`.
 For example:

```sql
CREATE TABLE ducklake.people(a integer, b varchar);

BEGIN;

INSERT INTO ducklake.people VALUES (1, 'pedro');

CALL ducklake.set_commit_message('Pedro', 'Inserting myself');

COMMIT;

SELECT snapshot_id, author, commit_message FROM ducklake.snapshots() where snapshot_id = 2
----
2	Pedro	Inserting myself
```

We also added an option that makes it mandatory for the user to provide this information. This can be enabled with `CALL ducklake.set_option('require_commit_message', True)`. The option is set to `False` by default. When set to `True`, the user must always call `set_commit_message`. Both `NULL` and `empty` strings are considered valid values for `author` and `commit_message`, so this option only enforces that the user explicitly sets them.

Additionally, this PR includes migration code from previous versions of DuckLake and fixes a small bug related to setting the same option multiple times.

### DIFF
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -337,9 +337,10 @@ struct DuckLakePath {
 
 struct DuckLakeSnapshotCommit {
 	//! Author of the commit
-	string author;
+	Value author;
 	//! The commit message for the snapshot
-	string commit_message;
+	Value commit_message;
+	//! If the user set the commit info for the snapshot
 	bool is_commit_info_set = false;
 };
 
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -340,6 +340,7 @@ struct DuckLakeSnapshotCommit {
 	string author;
 	//! The commit message for the snapshot
 	string commit_message;
+	bool is_commit_info_set = false;
 };
 
 struct DuckLakeConfigOption {
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -184,6 +184,8 @@ struct DuckLakeSnapshotInfo {
 	timestamp_tz_t time;
 	idx_t schema_version;
 	SnapshotChangeInfo change_info;
+	string author;
+	string commit_message;
 };
 
 struct DuckLakeViewInfo {
@@ -331,6 +333,13 @@ struct DuckLakeTableSizeInfo {
 struct DuckLakePath {
 	string path;
 	bool path_is_relative;
+};
+
+struct DuckLakeSnapshotCommit {
+	//! Author of the commit
+	string author;
+	//! The commit message for the snapshot
+	string commit_message;
 };
 
 struct DuckLakeConfigOption {
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -184,8 +184,8 @@ struct DuckLakeSnapshotInfo {
 	timestamp_tz_t time;
 	idx_t schema_version;
 	SnapshotChangeInfo change_info;
-	string author;
-	string commit_message;
+	Value author;
+	Value commit_message;
 };
 
 struct DuckLakeViewInfo {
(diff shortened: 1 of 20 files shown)


src/storage/ducklake_catalog.cpp
```diff
@@ -619,7 +619,7 @@ void DuckLakeCatalog::SetConfigOption(const DuckLakeConfigOption &option) {
 		return;
 	}
 	// scoped globally
-	options.config_options.emplace(key, value);
+	options.config_options[key] = value;
```
Mytherin, line 622: Can we also fix this for the other two cases?

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -184,6 +184,8 @@ struct DuckLakeSnapshotInfo {
 	timestamp_tz_t time;
 	idx_t schema_version;
 	SnapshotChangeInfo change_info;
+	string author;
```
Mytherin: Should this not be a `Value` so we can distinguish between `NULL` and the empty string?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2015,6 +2018,8 @@ ORDER BY snapshot_id
 		snapshot_info.time = GetTimestampTZFromRow(*context, row, 1);
 		snapshot_info.schema_version = row.GetValue<idx_t>(2);
 		snapshot_info.change_info.changes_made = row.IsNull(3) ? string() : row.GetValue<string>(3);
+		snapshot_info.author = row.IsNull(3) ? string() : row.GetValue<string>(4);
```
Mytherin: This is checking `NULL-ness` for the wrong column

src/storage/ducklake_transaction.cpp
```diff
@@ -1415,6 +1427,17 @@ unique_ptr<QueryResult> DuckLakeTransaction::Query(DuckLakeSnapshot snapshot, st
 	query = StringUtil::Replace(query, "{SCHEMA_VERSION}", to_string(snapshot.schema_version));
 	query = StringUtil::Replace(query, "{NEXT_CATALOG_ID}", to_string(snapshot.next_catalog_id));
 	query = StringUtil::Replace(query, "{NEXT_FILE_ID}", to_string(snapshot.next_file_id));
+	if (commit_info.author.IsNull()) {
+		query = StringUtil::Replace(query, "{AUTHOR}", "NULL");
+	} else {
+		query = StringUtil::Replace(query, "{AUTHOR}", commit_info.author.ToSQLString());
```
Mytherin: Doesn't `ToSQLString` map to `NULL` if it is NULL? Do we need this branching?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -50,7 +50,7 @@ void DuckLakeMetadataManager::InitializeDuckLake(bool has_explicit_schema, DuckL
 	string encryption_str = encryption == DuckLakeEncryption::ENCRYPTED ? "true" : "false";
 	initialize_query += StringUtil::Format(R"(
 CREATE TABLE {METADATA_CATALOG}.ducklake_metadata(key VARCHAR NOT NULL, value VARCHAR NOT NULL, scope VARCHAR, scope_id BIGINT);
-CREATE TABLE {METADATA_CATALOG}.ducklake_snapshot(snapshot_id BIGINT PRIMARY KEY, snapshot_time TIMESTAMPTZ, schema_version BIGINT, next_catalog_id BIGINT, next_file_id BIGINT);
+CREATE TABLE {METADATA_CATALOG}.ducklake_snapshot(snapshot_id BIGINT PRIMARY KEY, snapshot_time TIMESTAMPTZ, schema_version BIGINT, next_catalog_id BIGINT, next_file_id BIGINT, author VARCHAR, commit_message VARCHAR);
```
Mytherin: Can we add this to the `ducklake_snapshot_changes` table instead?

Mytherin (commented):
Thanks for the PR! Looks good - some comments

src/storage/ducklake_metadata_manager.cpp
pdet: Sure, should it still be part of the return of the  `ducklake_snapshots` functions?

Mytherin: Yes

Mytherin:
Thanks!

Issue #324, opened 2025-07-26
`ducklake_add_data_files()` fails with parquet files created by `INSERT INTO`


gacolitti:
`ducklake_add_data_files()` cannot add parquet files that were previously created by DuckLake's `INSERT INTO` operations.

### Steps to Reproduce

1. Create a DuckLake table and insert data:
 ```sql
 CREATE TABLE my_table AS SELECT * FROM source LIMIT 0;
 INSERT INTO my_table SELECT * FROM source;
```

2. Drop the table but keep the parquet files
3. Try to recreate the table and add the existing files:

```sql
CREATE TABLE my_table AS SELECT * FROM source LIMIT 0;
CALL ducklake_add_data_files('catalog', 'my_table', '/path/to/files/*.parquet');
```

Expected Behavior
The existing parquet files should be successfully added to the new table.

Actual Behavior
```
Error: Invalid Input Error: File has field ids defined - only mapping by name is supported currently
```

My guess is that when DuckLake creates parquet files via `INSERT INTO`, it automatically assigns field IDs for schema evolution. However, `ducklake_add_data_files()` currently only supports name-based mapping, not field ID mapping. But I'm not really sure what's going on. Would be nice to be able to recreate a table from existing parquet files.

ssidorenko:
What makes this issue particularly annoying is that if the metadata db gets corrupted, it becomes impossible to reconstruct it from the existing parquet files using `ducklake_add_data_files`.

guillesd:
@gacolitti this should be fixed in core nightly now!

gacolitti:
@guillesd Do you have an example I can run to see what should be working? The meta example I included still doesn't seem to be working after installing from core nightly.

guillesd:
@gacolitti What version of DuckDB are you on? Also if you could please provide a reproducer, something that I can run and is self-contained (preferably with the CLI).

gacolitti:
@guillesd I'm using v1.3.2.

This produces the error for me:

```sql
-- 1. Setup
FORCE INSTALL ducklake FROM core_nightly;
LOAD ducklake;
ATTACH 'ducklake:/tmp/ducklake_test/catalog' AS test_catalog (DATA_PATH '/tmp/ducklake_test');
USE test_catalog;

-- 2. Create and populate table
CREATE TABLE test_table AS SELECT random() as value LIMIT 0;
INSERT INTO test_table SELECT random() as value;

-- 3. Get file path (copy the result)
[...]

PR #338, opened 2025-08-01
[BUG] Remove Field ID check when adding parquet file

pdet:
Fix:  https://github.com/duckdb/ducklake/issues/324


Mytherin:
Thanks!

PR #340, opened 2025-08-04
Change retry options to Global and adjust error message to suggest an increase of the option.

pdet:


src/storage/ducklake_transaction.cpp
```diff
@@ -1361,7 +1361,13 @@ void DuckLakeTransaction::FlushChanges() {
 			if (!can_retry || !retry_on_error || finished_retrying) {
 				// we abort after the max retry count
 				CleanupFiles();
-				error.Throw("Failed to commit DuckLake transaction: ");
+				// Add additional information on number of retries and suggest to increase it
+				std::ostringstream error_message;
+				error_message << "Failed to commit DuckLake transaction." << '\n';
+				error_message << "* ducklake_max_retry_count is set to " << max_retry_count
```
Mytherin: Let's only add this if `finished_retrying` is triggered - otherwise this is just noise

Mytherin: I would suggest this more explicit error message in this scenario:

`Exceeded the maximum retry count of {max_retry_count} set by the `ducklake_max_retry_count` setting. Consider increasing the value with: "SET ...."`

Mytherin:
Thanks!

PR #343, opened 2025-08-05
Add extra_info parameter to commit messages.

pdet:
This PR allows users to insert extra information into commit messages. This extra information can be, for example, but is not limited to, a `JSON`.

Usage example:

```sql
CALL ducklake.set_commit_message('Pedro', 'Inserting Ollie', extra_info=>'{''barks'':7, ''cuteness'':10}');
```

Which will then result in:
```sql
SELECT snapshot_id, author, commit_message, commit_extra_info FROM ducklake.snapshots() where snapshot_id = 12
----
12	Pedro	Inserting Ollie	{'barks':7, 'cuteness':10}
```


Mytherin:
Thanks!

vcanaran:
Got it extra_info replaces params thank you. @pdet

PR #344, opened 2025-08-05
Automatically modify CREATE_IF_NOT_EXISTS to false if not manually set and attach is on read only mode

pdet:


Mytherin:
Thanks!

PR #345, opened 2025-08-05
added ducklake label to the internal forwarding bot

guillesd:
This PR just adds the label `ducklake` to the internal mirroring repository that we use for issue tracking


Mytherin:
Thanks!

PR #346, opened 2025-08-05
Add file footer to ducklake_add_data_files

pdet:


src/functions/ducklake_add_data_files.cpp
```diff
@@ -291,7 +292,7 @@ FROM read_blob(%s)
 	}
 	// use parquet_file_metadata to get the num rows
 	result = transaction.Query(StringUtil::Format(R"(
-SELECT file_name, num_rows
+SELECT file_name, num_rows, footer_size
 FROM parquet_file_metadata(%s)
```
Mytherin, line 280: `parquet_file_metadata` now also contains `file_size_bytes` - can we use that and get rid of the call to `read_blob` which is currently used to get the file size?

test/sql/add_files/add_file_footer_size.test
```diff
@@ -0,0 +1,25 @@
...
+
+# write a parquet file outside of DuckLake
+statement ok
+COPY (SELECT 200 col1, 'world' col2) TO '__TEST_DIR__/data.parquet';
+
+statement ok
+CALL ducklake_add_data_files('ducklake', 'test', '__TEST_DIR__/data.parquet')
+
+query I
+SELECT data_file_footer_size FROM ducklake_list_files('ducklake', 'test')
```
Mytherin: Let's just do `footer_size > 0`, otherwise this can randomly break (as already evidenced by the upstream change)

Mytherin (commented):
LGTM - can you just merge with main and fix the test issue?

Mytherin:
Thanks!

PR #347, opened 2025-08-05
Add use_hive_partitioning option

pdet:


guillesd:
I guess this will be an option exposed to the user. If that's the case let's add a `Needs documentation` label!

pdet:
> I guess this will be an option exposed to the user. If that's the case let's add a `Needs documentation` label!

You are absolutely right, bad practice from my side (:

src/functions/ducklake_options.cpp
```diff
@@ -22,6 +22,7 @@ static constexpr const ducklake_option_array DUCKLAKE_OPTIONS = {
      {"parquet_compression_level", "Compression level for Parquet files"},
      {"parquet_row_group_size", "Number of rows per row group in Parquet files"},
      {"parquet_row_group_size_bytes", "Number of bytes per row group in Parquet files"},
+     {"use_hive_partitioning", "If data should be partitioned in the hive format"},
```
Mytherin: In DuckDB this is called `hive_file_pattern` - maybe it makes sense to keep that convention?

Mytherin: Also maybe "`if partitioned data should be written in a hive-like folder structure`" is maybe more clear. The partitioning strategy used is unchanged after all  - the files will be partitioned in the same manner regardless of this setting.

src/storage/ducklake_insert.cpp
```diff
@@ -621,7 +621,12 @@ PhysicalOperator &DuckLakeInsert::PlanCopyForInsert(ClientContext &context, Phys
 	physical_copy.names = std::move(copy_options.names);
 	physical_copy.expected_types = std::move(copy_options.expected_types);
 	physical_copy.parallel = true;
-	physical_copy.hive_file_pattern = !is_encrypted ? true : false;
+	if (is_encrypted) {
+		// if encrypted we never use hive partitioning
+		physical_copy.hive_file_pattern = false;
```
Mytherin: I think for encrypted databases we just want to default this setting to `false` on creation - there might be cases where people want to write encrypted databases in a hive file pattern. The setting should still be respected.

Mytherin (commented):
Thanks! LGTM - some comments

Mytherin:
Thanks!

djouallah:
does not works in 1.4.0.dev2303 ?

Issue #334, opened 2025-07-30
Add schema support to `ducklake_add_data_files`


utay:
Currently the `ducklake_add_data_files` function takes a catalog and a table, but we can't add a data file to a table of a specific schema. Looking at the code, `schema_name` is left undefined:
https://github.com/duckdb/ducklake/blob/b242186a8c9a13d1f756fe29260f4c0f7ef988b6/src/functions/ducklake_add_data_files.cpp#L24-L34

It would be great to be able to specify the schema like:
```
CALL ducklake_add_data_files('my_ducklake', '<schema>', 'people', 'people.parquet');
```
or
```
CALL ducklake_add_data_files('my_ducklake', '<schema>.people', 'people.parquet');
```

PR #348, opened 2025-08-06
Allow schema to be defined in ducklake_add_data_files

pdet:
Fix: https://github.com/duckdb/ducklake/issues/334


src/functions/ducklake_add_data_files.cpp
```diff
@@ -30,6 +30,12 @@ static unique_ptr<FunctionData> DuckLakeAddDataFilesBind(ClientContext &context,
 		throw InvalidInputException("Table name cannot be NULL");
 	}
 	auto table_name = StringValue::Get(input.inputs[1]);
+	auto table_info = StringUtil::Split(table_name, '.');
```
Mytherin: Let's not do this because this means we can no longer support tables with a dot in them correctly. Elsewhere we use an explicit `schema` parameter.

pdet: What should the schema option be then? a named parameter so we can have a default value? Or the user always needs to specify it?

Mytherin: Yes, a named parameter with a default value just like in other places where this is used

Mytherin:
Thanks!

PR #350, opened 2025-08-06
current_snapshot() Function

pdet:
This PR implements a function that returns the id of the last committed snapshot.

The function can be used as :
```sql
ATTACH 'ducklake:__TEST_DIR__/ducklake_list_files.db' AS ducklake;

CREATE TABLE integer(i INTEGER);

FROM ducklake_current_snapshot('ducklake')
----
1
```

or with a macro:
```sql
USE ducklake;

FROM current_snapshot()
```

### DIFF
--- a/src/functions/ducklake_last_snapshot.cpp
+++ b/src/functions/ducklake_last_snapshot.cpp
@@ -1,0 +1,30 @@
+#include "functions/ducklake_table_functions.hpp"
+#include "storage/ducklake_table_entry.hpp"
+#include "storage/ducklake_transaction.hpp"
+#include "common/ducklake_util.hpp"
+#include "storage/ducklake_transaction_changes.hpp"
+#include "duckdb/planner/tableref/bound_at_clause.hpp"
+
+namespace duckdb {
+
+static unique_ptr<FunctionData> DuckLakeLastSnapshotBind(ClientContext &context, TableFunctionBindInput &input,
+                                                         vector<LogicalType> &return_types, vector<string> &names) {
+	auto &catalog = BaseMetadataFunction::GetCatalog(context, input.inputs[0]);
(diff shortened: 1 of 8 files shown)


src/functions/ducklake_last_snapshot.cpp
```diff
@@ -0,0 +1,30 @@
...
+namespace duckdb {
+
+static unique_ptr<FunctionData> DuckLakeLastSnapshotBind(ClientContext &context, TableFunctionBindInput &input,
+                                                         vector<LogicalType> &return_types, vector<string> &names) {
+	auto &catalog = BaseMetadataFunction::GetCatalog(context, input.inputs[0]);
+	auto &transaction = DuckLakeTransaction::Get(context, catalog);
+
+	names.emplace_back("id");
+	return_types.emplace_back(LogicalType::UBIGINT);
+	auto snapshot = transaction.GetSnapshot();
```
Mytherin: This is not the last committed snapshot but the currently active snapshot. We want this functionality as well, but it should be in a separate function that is named appropriately (`ducklake_get_current_snapshot`).

`ducklake_last_committed_snapshot` should refer to the last snapshot committed by the current system (and `NULL` if there is none), i.e. we want to have an `optional_idx last_committed_snapshot` in `DuckLakeCatalog`, which is set by `DuckLakeTransaction::FlushChanges` on a successful commit.

pdet: Aha, got it!

pdet: I'll rename this PR to `ducklake_get_current_snapshot` and will do a different one from the last committed then.

Mytherin: Sounds good

Mytherin:
Thanks!

PR #355, opened 2025-08-07
Add DuckLake DuckDB test config to CI

Mytherin:
CC @pdet


Mytherin:
There's a number of tests that differ only in result ordering - https://github.com/duckdb/duckdb/pull/18542 should solve that once merged

pdet:
@Mytherin I'm picking this up.

PR #356, opened 2025-08-07
Get information on the last committed snapshot

pdet:
This PR implements the `ducklake_last_committed_snapshot(catalog)` and `last_committed_snapshot` macro.

@Mytherin can you double-check the logic when loading an existing ducklake?

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -140,6 +140,25 @@ public:
 		return ++last_uncommitted_catalog_version;
 	}
 
+	void IncreaseCommitedSnapshotId() {
+		if (!last_committed_snapshot.IsValid()) {
+			last_committed_snapshot = 0;
+		} else {
+			last_committed_snapshot = last_committed_snapshot.GetIndex() + 1;
+		}
+	}
+
+	void SetCommitedSnapshotId(idx_t value) {
+		last_committed_snapshot = value;
+	}
+
+	Value GetLastCommitedSnapshotId() const {
+		if (last_committed_snapshot.IsValid()) {
+			return Value::UBIGINT(last_committed_snapshot.GetIndex());
+		}
+		return Value();
+	}
+
 	optional_ptr<const DuckLakeNameMap> TryGetMappingById(DuckLakeTransaction &transaction, MappingIndex mapping_id);
 	MappingIndex TryGetCompatibleNameMap(DuckLakeTransaction &transaction, const DuckLakeNameMap &name_map);
 
@@ -176,6 +195,8 @@ private:
 	string metadata_type;
 	//! Whether or not the catalog is initialized
 	bool initialized = false;
+	//! The id of the last committed snapshot, set at FlushChanges on a successful commit
+	optional_idx last_committed_snapshot;
 };
 
 } // namespace duckdb
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -121,7 +121,20 @@ UPDATE {METADATA_CATALOG}.ducklake_metadata SET value = '0.3-dev1' WHERE key = '
 }
 
 DuckLakeMetadata DuckLakeMetadataManager::LoadDuckLake() {
+	// Get last committed snapshot
 	auto result = transaction.Query(R"(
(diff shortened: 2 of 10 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -121,7 +121,20 @@ UPDATE {METADATA_CATALOG}.ducklake_metadata SET value = '0.3-dev1' WHERE key = '
 }
 
 DuckLakeMetadata DuckLakeMetadataManager::LoadDuckLake() {
+	// Get last committed snapshot
 	auto result = transaction.Query(R"(
+	SELECT MAX(snapshot_id) FROM {METADATA_CATALOG}.ducklake_snapshot
```
Mytherin: Let's skip this and just leave it as `NULL` if the current thread hasn't made any changes

src/include/storage/ducklake_catalog.hpp
```diff
@@ -140,6 +140,25 @@ class DuckLakeCatalog : public Catalog {
 		return ++last_uncommitted_catalog_version;
 	}
 
+	void IncreaseCommitedSnapshotId() {
+		if (!last_committed_snapshot.IsValid()) {
```
Mytherin: This is on the catalog level - so either `last_committed_snapshot` needs to be an atomic, or all of these functions need a lock.

Mytherin: Can we add a test where we call this function in a `concurrentloop` while another thread does commits?

src/storage/ducklake_metadata_manager.cpp
pdet: I'm a bit confused what the last commit should be in this context then. 
Should the last commit be the current snapshot id, if a transaction commits? Otherwise this becomes a commit counter, and it is unrelated to the id right

src/include/storage/ducklake_catalog.hpp
pdet: Do you envision this in a sqltest?
Like a 
```sql
concurrentloop i 0 15

skipif i>7
...

skipif i<7
....
```

 ?

Mytherin: Yes

src/storage/ducklake_transaction.cpp
```diff
@@ -1388,6 +1387,8 @@ void DuckLakeTransaction::FlushChanges() {
 			snapshot.reset();
 		}
 	}
+	// If we got here, this snapshot was successful
+	ducklake_catalog.SetCommittedSnapshotId(GetSnapshot().snapshot_id);
```
Mytherin: This should use `commit_snapshot.snapshot_id` - not call `GetSnapshot` again

test/sql/snapshot_info/ducklake_last_commit.test
```diff
@@ -0,0 +1,127 @@
...
+NULL
+
+# Snapshot 1
+statement ok
+CREATE TABLE ducklake.integer(i INTEGER);
+
+query I
+FROM ducklake_last_committed_snapshot('ducklake')
+----
+0
```
Mytherin: This should return 1 no?

guillesd:
@pdet how is this different from the `current_snapshot()` macro?

pdet:
@guillesd The current snapshot gets the current snapshot id of a transaction, before the commit, while last_committed_snapshot gets the last committed snapshot id. I think I was actually missing a -1 there.

For example, if you have:

```sql
BEGIN TRANSACTION

FROM ducklake_last_committed_snapshot('ducklake')
---- this would return NULL

FROM ducklake_current_snapshot('ducklake')
---- this would return 0

CREATE TABLE ...

COMMIT

FROM ducklake_last_committed_snapshot('ducklake')
---- this would return 0

FROM ducklake_current_snapshot('ducklake')
---- this would return 1

```

src/storage/ducklake_transaction.cpp
```diff
@@ -1388,6 +1388,8 @@ void DuckLakeTransaction::FlushChanges() {
 			snapshot.reset();
 		}
 	}
+	// If we got here, this snapshot was successful
+	ducklake_catalog.SetCommittedSnapshotId(commit_snapshot.snapshot_id - 1);
```
Mytherin: The -1 shouldn't be here, otherwise LGTM

Mytherin:
Thanks!

Issue #336, opened 2025-07-31
`ducklake_merge_adjacent_files` cannot compact files originally created at a now-expired snapshot


dylanzal:
If a data file is associated with a snapshot that has been expired (`ducklake_data_file.begin_snapshot` references a snapshot that no longer exists in `ducklake_snapshot`), it is no longer a viable candidate for compaction. Is this expected behaviour? For my use-case, I have a periodic maintenance job that expires snapshots older than some threshold. An unfortunate consequence of this is that I accumulate many small files, even though I am running frequent compaction operations, because some data files are no longer associated with a valid snapshot.

guillesd:
Could you explain a bit further and maybe give a reproducible example @dylanzal ?

dylanzal:
Sure, here's an example:
```
INSTALL ducklake;
ATTACH 'ducklake:test.ducklake' AS test;
USE test;
CREATE TABLE example (key VARCHAR, value VARCHAR);
INSERT INTO example (key, value) VALUES ('foo', 'bar');
INSERT INTO example (key, value) VALUES ('baz', 'qux');
```

At this point, if I were to run:
```
CALL ducklake_merge_adjacent_files('test');
SELECT count(*) FROM ducklake_list_files('test', 'example');
```
I see there is one data file in the table, the result of the compaction operation. This is what I expect. However, if instead I run the following:
```
CALL ducklake_expire_snapshots('test', older_than => now());
CALL ducklake_merge_adjacent_files('test');
SELECT count(*) FROM ducklake_list_files('test', 'example');
```
I see two data files, and no compaction has occurred. The snapshot associated with the first insert has been expired, and it seems that the associated data file is no longer deemed a viable candidate for compaction. This isn't what I would expect to happen, and results in the accumulation of a large number of pretty small files in my table for a workload where I:
- have frequent inserts
- periodically run compaction
- periodically expire old snapshots (e.g., snapshots older than 1 day)

guillesd:
Hi @dylanzal ! Your example does not work since `CALL ducklake_expire_snapshots('test', older_than => now());` does not really expire anything (otherwise there would be no Snapshots left?). However I did get it to work by removing the specific snapshot associated to the first `INSERT` operation.
```sql
ATTACH 'ducklake:my_ducklake.ducklake' as my_ducklake;
USE my_ducklake;
CREATE TABLE example (key VARCHAR, value VARCHAR);
INSERT INTO example (key, value) VALUES ('foo', 'bar');
[...]

PR #361, opened 2025-08-08
merge_adjacent_files() on expired snapshots

pdet:
This PR changes the compaction query to also be able to compact files from expired snapshots, as long as they fit within the same schema version.

Fix: https://github.com/duckdb/ducklake/issues/336

### DIFF
--- a/test/sql/compaction/merge_files_expired_snapshots.test
+++ b/test/sql/compaction/merge_files_expired_snapshots.test
@@ -55,3 +55,74 @@ FROM example
 ----
 foo	bar
 baz	qux
+
+# Add a few schema changes
+statement ok
+INSERT INTO example (key, value) VALUES ('a', 'b');
+
+statement ok
+ALTER TABLE example ADD COLUMN j INTEGER
+
+statement ok
+INSERT INTO example VALUES ('c', 'd', 1);
+
+statement ok
+INSERT INTO example VALUES ('e', 'f', 2);
+
+statement ok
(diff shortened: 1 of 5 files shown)


test/sql/compaction/merge_files_expired_snapshots.test
```diff
@@ -0,0 +1,128 @@
...
+2	{tables_inserted_into=[1]}
+3	{tables_inserted_into=[1]}
+
+query I
+SELECT count(*) FROM ducklake_list_files('ducklake', 'example');
+----
+2
+
+statement ok
+CALL ducklake_expire_snapshots('ducklake', versions => [2]);
```
Mytherin: This still won't work if we expire `versions => [1, 2]` because then we have lost the schema version. I think the only way to fix this is to track the schema version outside of the `ducklake_snapshot` table, for example we can add this table:


```
CREATE TABLE ducklake_schema_versions(begin_snapshot BIGINT, schema_version BIGINT);
```

This table can be populated whenever `SchemaChangesMade()` is true with the new snapshot id and the new schema version. 

For migration, we could do e.g.:

```
INSERT INTO ducklake_schema_versions SELECT MIN(snapshot_id), schema_version FROM ducklake_snapshot GROUP BY schema_version ORDER BY schema_version;
```

Then we can figure out the schema version associated with a file also after snapshots have been expired.

Mytherin:
Thanks!

Issue #310, opened 2025-07-20
merge_adjacent_files() doesn't use hive-partition folder structure


gacolitti:
I expect that, for a table with partitioning, calling `merge_adjacent_files()` would place new consolidated files in the respective partitioning nested folders. But it appears the new consolidated files are placed in the table's root directory instead. By contrast, inserting data with `insert` puts the new files in the sub directories as expected. Can we place merged files in partitioned directories when using the `merge_adjacent_files()` command?

Sidenote: there's a typo on this page in the **Partitioned Files** section: https://ducklake.select/docs/stable/duckdb/usage/paths.

ben-burwood:
Just tagging on as we are also hitting up against this issue.

Optionally there could be an option to merge_adjacent_files with `respect_partitions` , where files will only be merged within the same Partioned Directories?

gacolitti:
From what I understand, it's already only merging by looking at files within each individual partition. It's just that the new merged files are not placed in the partitions.

guillesd:
@gacolitti can you provide a reproducible example with a public dataset? This seems like something we need to address indeed!

guillesd:
Okey I did reproduce this:

```sql
CREATE TABLE my_ducklake.sales (
      id INTEGER,
      product STRING,
      amount DOUBLE,
      sale_date DATE
  );
ALTER TABLE my_ducklake.sales SET PARTITIONED BY (sale_date);
INSERT INTO my_ducklake.sales
  VALUES
      (1, 'Product A', 100.0, '2023-07-01'),
      (2, 'Product B', 150.0, '2023-07-01');
INSERT INTO my_ducklake.sales
  VALUES
      (3, 'Product A', 100.0, '2023-07-02'),
      (4, 'Product B', 150.0, '2023-07-02');
INSERT INTO my_ducklake.sales
  VALUES
      (5, 'Product C', 300.0, '2023-07-02'),
      (6, 'Product B', 150.0, '2023-07-02');
```
This creates 1 file under `sales/sale_date=2023-07-01` and 2 files under `sales/sale_date=2023-07-02`. Then:
```sql
CALL my_ducklake.merge_adjacent_files();
```
Which creates a new file with all the records from the partition `2023-07-02` under `sales` rather than under `sales/sale_date=2023-07-02`.

yan-hic:
[...]

PR #363, opened 2025-08-11
Ensure that hive partitions are kept after compaction

pdet:
Fix: https://github.com/duckdb/ducklake/issues/310

### DIFF
--- a/src/functions/ducklake_merge_adjacent_files.cpp
+++ b/src/functions/ducklake_merge_adjacent_files.cpp
@@ -222,7 +222,7 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 		candidates[group].candidate_files.push_back(file_idx);
 	}
 
-	// we have gathered all of the candidate files per compaction group
+	// we have gathered all the candidate files per compaction group
 	// iterate over them to generate actual compaction commands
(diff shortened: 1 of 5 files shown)


guillesd:
hey @pdet do you think this will also fix https://github.com/duckdb/ducklake/issues/373? It is basically a similar thing where, in this case, after an `ALTER TABLE` command, the next file added is outside of the hive partition.

pdet:
> hey @pdet do you think this will also fix #373? It is basically a similar thing where, in this case, after an `ALTER TABLE` command, the next file added is outside of the hive partition.

Jup, it fixes it as well, I've added in the test.

src/functions/ducklake_merge_adjacent_files.cpp
```diff
@@ -322,11 +322,31 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
 
 	// generate the LogicalGet
 	auto &columns = table.GetColumns();
+	auto table_path = table.DataPath();
+	string data_path;
+	if (!source_files.empty()) {
```
Mytherin: Instead of looking at the data path directly, can we look at the `partition_id` and `partition_values` of the files

Mytherin:
Thanks!

Issue #330, opened 2025-07-27
table_changes data change feed does not follow the spec


rchui:
**Expected**
The table_changes function should allow you to see changes between any two snapshots and any two points in time.

> In addition to allowing you to query the [state of the database at any point in time](https://ducklake.select/docs/stable/duckdb/usage/time_travel.html), DuckLake allows you to query the changes that were made **between any two snapshots**. This can be done using the table_changes function.

https://ducklake.select/docs/stable/duckdb/advanced_features/data_change_feed

**Actual**

Checking between snapshots [0, 2] vs [0, 4] does not continue to show the changes. I would expect the changes to be cumulative between two snapshots but it appears to only show the literal diff between snapshot 0 and snapshot 4 not all changes between snapshot 0 and snapshot 4.

```sql
D from table_changes('stg_sales_orders', 0, 2);
┌─────────────┬───────┬─────────────┬───┬──────────┬───────────────┬────────────┬──────────────────────┐
│ snapshot_id │ rowid │ change_type │ … │ quantity │   unit_cost   │ order_date │      updated_at      │
│    int64    │ int64 │   varchar   │   │  double  │ decimal(18,6) │    date    │ timestamp with tim…  │
├─────────────┴───────┴─────────────┴───┴──────────┴───────────────┴────────────┴──────────────────────┤
│                                                0 rows                                                │
└──────────────────────────────────────────────────────────────────────────────────────────────────────┘
D from table_changes('stg_sales_orders', 0, 3);
┌─────────────┬───────┬─────────────┬───┬───────────────┬────────────┬──────────────────────┐
│ snapshot_id │ rowid │ change_type │ … │   unit_cost   │ order_date │      updated_at      │
│    int64    │ int64 │   varchar   │   │ decimal(18,6) │    date    │ timestamp with tim…  │
├─────────────┼───────┼─────────────┼───┼───────────────┼────────────┼──────────────────────┤
│           3 │     0 │ insert      │ … │      0.050000 │ 2025-07-27 │ 2025-07-27 17:02:0…  │
│           3 │     1 │ insert      │ … │      0.070000 │ 2025-07-27 │ 2025-07-27 17:02:0…  │
│           3 │     2 │ insert      │ … │      7.890000 │ 2025-07-27 │ 2025-07-27 17:02:0…  │
├─────────────┴───────┴─────────────┴───┴───────────────┴────────────┴──────────────────────┤
│ 3 rows                                                               11 columns (6 shown) │
[...]

PR #366, opened 2025-08-11
The source snapshot should find the snapshot even if the timestamp is lower than all snapshots within the catalog

pdet:
Fix: https://github.com/duckdb/ducklake/issues/330

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -95,7 +95,7 @@ public:
 	virtual void UpdateGlobalTableStats(const DuckLakeGlobalStatsInfo &stats);
 	virtual SnapshotChangeInfo GetChangesMadeAfterSnapshot(DuckLakeSnapshot start_snapshot);
 	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot();
-	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot(BoundAtClause &at_clause);
+	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot(BoundAtClause &at_clause, bool from = false);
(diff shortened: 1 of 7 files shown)


src/include/storage/ducklake_metadata_manager.hpp
```diff
@@ -95,7 +95,7 @@ class DuckLakeMetadataManager {
 	virtual void UpdateGlobalTableStats(const DuckLakeGlobalStatsInfo &stats);
 	virtual SnapshotChangeInfo GetChangesMadeAfterSnapshot(DuckLakeSnapshot start_snapshot);
 	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot();
-	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot(BoundAtClause &at_clause);
+	virtual unique_ptr<DuckLakeSnapshot> GetSnapshot(BoundAtClause &at_clause, bool from = false);
```
Mytherin: Can we turn this into a descriptive enum - e.g. `LOWER_BOUND/UPPER_BOUND`

test/sql/table_changes/ducklake_lower_timestamp.test
```diff
@@ -0,0 +1,39 @@
...
+statement ok
+insert into repro by name select 1 as foo, '1' as bar, current_timestamp as update_ts;
+
+statement ok
+insert into repro by name select 2 as foo, '2' as bar, current_timestamp as update_ts;
+
+statement ok
+insert into repro by name select 3 as foo, '3' as bar, current_timestamp as update_ts;
+
+# statement ok
```
Mytherin: nit is this commented out on purpose

Mytherin:
Thanks!

PR #368, opened 2025-08-12
add issue template and discussion redirect

guillesd:
I think it is nice if we get the right information from the start!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,164 @@
+name: Bug report
+description: Create a report to help us improve
+labels:
+  - needs triage
+body:
+  - type: markdown
+    attributes:
+      value: >
+        DuckDB has several repositories for different components, please make sure you're raising your issue in the correct one:
+
+          * [Documentation/website](https://github.com/duckdb/ducklake-web/issues/new)
+          * [Core DuckDB](https://github.com/duckdb/duckdb/issues/new)
+          * APIs:
+            * [duckdb-java](https://github.com/duckdb/duckdb-java/issues/new)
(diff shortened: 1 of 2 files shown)


guillesd:
@pdet this should be it!

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,164 @@
...
+            * [MySQL extension](https://github.com/duckdb/duckdb-mysql/issues/new)
+            * [Postgres scanner](https://github.com/duckdb/duckdb-postgres/issues/new)
+            * [Spatial extension](https://github.com/duckdb/duckdb-spatial/issues/new)
+            * [SQLite scanner](https://github.com/duckdb/duckdb-sqlite/issues/new)
+            * [UI](https://github.com/duckdb/duckdb-ui/issues/new)
+            * [VSS extension](https://github.com/duckdb/duckdb-vss/issues/new)
+          * Connectors:
+            * [dbt-duckdb](https://github.com/duckdb/dbt-duckdb)
+
+        If none of the above repositories are applicable, feel free to raise it in this one
```
pdet: This is a bit of an overkill, no? I think that for duckdb it makes sense since it might be the first point where issues generally happen, my guess is that issues opened here would be sth like 80% related to ducklake, 10% duckdb and 10% either related to postgres,sqlite,mysql scanner, secrets, and so on.

guillesd: Dale! I will just reference the main DuckDB repo and the docs

guillesd:
@hannes if you can merge this one that would be great! Pedro already reviewed

hannes:
Thanks!

Issue #281, opened 2025-07-11
INTERNAL Error: Attempted to access index 0 within vector of size 0


erikcw:
Just got this traceback when attempting to call `duckdb.sql("CALL cc_ducklake.merge_adjacent_files();", connection=conn)` (from Python).  I'm guessing its because there are no files to merge as the command ran sucessfully earlier and the vector size is being reported as `0`.  Seems like some internal error handling is missing for this edge case.

```python
duckdb.sql("CALL cc_ducklake.merge_adjacent_files();", connection=conn)

duckdb.duckdb.InternalException: INTERNAL Error: Attempted to access index 0 within vector of size 0                                                                                                                                                                     [15/42]

Stack Trace:

/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(+0x7660bc) [0xfffef9ba60bc]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(+0x766188) [0xfffef9ba6188]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(+0x767adc) [0xfffef9ba7adc]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(_ZN6duckdb17InternalExceptionC1IJmmEEERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEDpT_+0x148) [0xfffef9634a68]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(_ZN6duckdb6vectorINS_27DuckLakeCompactionFileEntryELb1EE19AssertIndexInBoundsEmm+0x58) [0xfffef9655998]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(_ZN6duckdb17DuckLakeCompactor25GenerateCompactionCommandENS_6vectorINS_27DuckLakeCompactionFileEntryELb1EEE+0x5c) [0xfffef9655a3c]
/root/.duckdb/extensions/v1.3.2/linux_arm64/ducklake.duckdb_extension(_ZN6duckdb17DuckLakeCompactor19GenerateCompactionsERNS_18DuckLakeTableEntryERNS_6vectorINS_10unique_ptrINS_15LogicalOperatorESt14default_deleteIS5_ELb1EEELb1EEE+0x844) [0xfffef9659bc4]
[...]

PR #369, opened 2025-08-12
Compaction Bug Fix

pdet:
Fix: https://github.com/duckdb/ducklake/issues/281

This PR fixes a compaction bug that would get triggered when the file size of the newly compacted file would be bigger than the max file size setting.


arouel:
@pdet please review our simplified test setup to reproduce this issue at https://github.com/duckdb/ducklake/issues/281#issuecomment-3182608563. I think you can use "sqltest" to do the same.

pdet:
> @pdet please review our simplified test setup to reproduce this issue at [#281 (comment)](https://github.com/duckdb/ducklake/issues/281#issuecomment-3182608563). I think you can use "sqltest" to do the same.

You are absolutely right, this test case worked like a charm! Many thanks!

guillesd:
@pdet I ran the Python reproducer with this branch and DuckDB's 1.4Dev and it worked great! Thanks

PR #371, opened 2025-08-13
Bind DuckLakeViewEntry when first scanned

pdet:


src/storage/ducklake_schema_entry.cpp
```diff
@@ -242,6 +242,16 @@ void DuckLakeSchemaEntry::Scan(ClientContext &context, CatalogType type,
 			// this entry exists in both the local and global set - emit only the transaction-local entry
 			continue;
 		}
+		if (entry.second->type == CatalogType::VIEW_ENTRY) {
+			auto view_entry = dynamic_cast<DuckLakeViewEntry *>(entry.second.get());
```
Mytherin: nit: `entry.second->Cast<DuckLakeViewEntry>()`

Mytherin:
Thanks!

PR #380, opened 2025-08-15
Fix Internal Error issues with column dropping.

pdet:
This PR fixes the following tests:
* test/sql/catalog/case_insensitive_caps.test
* test/sql/catalog/case_insensitive_operations.test
* test/sql/alter/map/drop_column_in_struct.test
* test/sql/alter/list/drop_column_in_struct.test

### DIFF
--- a/src/storage/ducklake_field_data.cpp
+++ b/src/storage/ducklake_field_data.cpp
@@ -174,7 +174,7 @@ unique_ptr<DuckLakeFieldId> DuckLakeFieldId::AddField(const vector<string> &colu
 	vector<unique_ptr<DuckLakeFieldId>> new_children;
 	if (depth >= column_path.size()) {
 		// leaf - add the column at this level
-		// copy over all of the other columns as-is
+		// copy over all the other columns as-is
 		for (auto &child : children) {
 			new_children.push_back(child->Copy());
 		}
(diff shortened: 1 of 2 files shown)


src/storage/ducklake_field_data.cpp
```diff
@@ -207,6 +209,13 @@ unique_ptr<DuckLakeFieldId> DuckLakeFieldId::RemoveField(const vector<string> &c
 	for (idx_t child_idx = 0; child_idx < children.size(); child_idx++) {
 		auto &child = *children[child_idx];
 		if (StringUtil::CIEquals(child.Name(), column_path[depth])) {
+			if (depth == 1 && column_path.size() == 2 &&
+			    ((type.id() == LogicalTypeId::MAP &&
```
Mytherin: We only need to check the type id here - we can never drop anything from a map or a list

pdet: I think we still need to check `column_path.size() == 2` to ensure our delete is in the correct layer

OW,
```sql
statement ok
ALTER TABLE test DROP COLUMN s.value.j
```
fails.

Mytherin: Makes sense

Mytherin:
Thanks!

PR #387, opened 2025-08-19
Fix test/sql/catalog/test_schema.test

pdet:
This PR has 3 fixes for a DuckDB test that is throwing an internal error.

1. Not add dropped elements to dependent
2. Make error message of dependents match duckdb
3. Get snapshot of transaction if Immediate Transaction Mode is set.

@Mytherin, I reckon that 3. is not the correct fix for it, I think you recommended starting a transaction there? But I'm not exactly sure what you meant.

### DIFF
--- a/src/storage/ducklake_transaction_manager.cpp
+++ b/src/storage/ducklake_transaction_manager.cpp
@@ -1,5 +1,7 @@
 #include "storage/ducklake_transaction_manager.hpp"
 
+#include "duckdb/main/settings.hpp"
+
 namespace duckdb {
 
 DuckLakeTransactionManager::DuckLakeTransactionManager(AttachedDatabase &db_p, DuckLakeCatalog &ducklake_catalog)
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_transaction_manager.cpp
```diff
@@ -9,6 +11,12 @@ DuckLakeTransactionManager::DuckLakeTransactionManager(AttachedDatabase &db_p, D
 Transaction &DuckLakeTransactionManager::StartTransaction(ClientContext &context) {
 	auto transaction = make_shared_ptr<DuckLakeTransaction>(ducklake_catalog, *this, context);
 	transaction->Start();
+	if (DBConfig::GetSetting<ImmediateTransactionModeSetting>(context) && get_snapshot) {
+		get_snapshot = false;
+		// no snapshot loaded yet for this transaction - load it
+		transaction->GetSnapshot();
```
Mytherin, line 17: Can we do `transaction->GetConnection().BeginTransaction()` here or does that not solve the issue?

pdet: I'll get a
`Startup queries provided via on_init failed: TransactionContext Error: cannot
  start a transaction within a transaction`

Mytherin: I think it's fine as-is, it's a debug flag anyway

Mytherin:
Thanks!

PR #393, opened 2025-08-25
Introduction of ducklake_rewrite_data_files function

pdet:
This PR introduces the `ducklake_rewrite_data_files` function. This function rewrites and merges data files with deletions, as long as they exceed a given threshold.

By default, the threshold, is 95% of the file, which means that up to that point, at least 95% of the file must have been deleted. If multiple files of the same table are being rewritten, these files are also merged into one file.

It is important to notice that currently, this function only works on the latest snapshot. When this function is called, the deletion file of the last delete is also marked to be removed, since that is not necessary for snapshot reconstruction, as it is applied to the new rewritten dataset. Previous files, will have their `end_snapshot` written accordingly to take into account the new file.

## Usage
This function comes in two flavours.
1. `ducklake_rewrite_data_files('ducklake')` where the function will be executed on all tables from a database.
2. `ducklake_rewrite_data_files('ducklake', 'table')` where the function will be executed only in a specified table.

It also takes two named parameters. One of them is the `schema` and the other the `delete_threshold`. The `delete_threshold` goes from [0,1] and informs the % of the data that must be deleted for a file to be considered for rewriting.

It is also possible to change the default value of the threshold with a ducklake option:
`CALL ducklake.set_option('compaction_delete_threshold_rewrite', 0.1)`
Notice that this will change the default value for the whole ducklake. However if the threshold option is set when calling the function, that value will have priority over the set default.

## Benchmark
The following is a simple benchmark where a table with `100_000_000` values, has all values equal or over `50` deleted.

```sql
ATTACH 'ducklake:ducklake_deletes.db' AS ducklake (DATA_PATH 'ducklake_delete_files/tmp');
use ducklake
create or replace table delete_test as select range a from range(100_000_000);
.timer on
delete from delete_test where a >= 50;
create or replace table copied_delete as select range a from range(50);

-- 50 rows
-- Run Time (s): real 0.824 user 0.808596 sys 0.209569
select * from delete_test;
-- also 50 rows
-- Run Time (s): real 0.002 user 0.001487 sys 0.000606
select * from copied_delete;

CALL ducklake_rewrite_data_files('ducklake', 'delete_test', delete_threshold => 0);

-- Rewrite the deletes

```


src/storage/ducklake_transaction.cpp
```diff
@@ -55,6 +55,7 @@ void DuckLakeTransaction::Commit() {
 	} else if (connection) {
 		connection->Commit();
 	}
+	// connection->Query("");
```
pdet: Ah, I still have to run this test

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -21,6 +21,15 @@
 
 namespace duckdb {
 
+//===--------------------------------------------------------------------===//
+// Compaction Type
+//===--------------------------------------------------------------------===//
+
+enum CompactionType {
```
Mytherin: `enum class` to avoid polluting the global namespace

As a general rule you should never use C-style enums

src/storage/ducklake_transaction.cpp
```diff
@@ -1237,25 +1237,44 @@ void DuckLakeTransaction::CommitChanges(DuckLakeCommitState &commit_state,
...
 		auto &table_changes = entry.second;
 		for (auto &compaction : table_changes.compactions) {
+			if (type != compaction.type) {
+				continue;
+			}
 			auto new_file = GetNewDataFile(compaction.written_file, commit_snapshot, table_id, compaction.row_id_start);
-			new_file.begin_snapshot = compaction.source_files[0].file.begin_snapshot;
+			switch (type) {
+			case REWRITE_DELETES:
+				new_file.begin_snapshot = compaction.source_files[0].delete_files.back().begin_snapshot;
```
Mytherin, line 1266: Are the delete files guaranteed to be ordered?

test/sql/rewrite_data_files/test_last_snapshot_rewrite.test
```diff
@@ -0,0 +1,218 @@
...
+query I
+SELECT delete_file_id FROM ducklake_metadata.ducklake_delete_file
+----
+1
+2
+3
+
+# This won't do much, since our default threshold is too high
+statement ok
+CALL ducklake_rewrite_data_files('ducklake', 'test');
```
Mytherin, line 59: Can we add a test where we run `ducklake_rewrite_data_files` followed by resuming regular operations? e.g. maybe in a loop, something like:

```
loop i 0 10

statement ok
INSERT INTO tbl FROM range(${i} * 1000, ${i} * 1000 + 1000);

statement ok
DELETE FROM tbl WHERE i BETWEEN ${i} * 1000 AND ${i} * 1000 + 950

statement ok
CALL ducklake_rewrite_data_files('ducklake', 'tbl');

query I
SELECT COUNT(*) = ${i} * 50 FROM tbl
----
true

endloop


```

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2194,6 +2203,113 @@ WHERE data_file_id IN (%s);
...
+	if (compactions.empty()) {
+		return;
+	}
+	// Delete Rewrites only deletes the deletion files.
+	string deleted_file_ids;
+	string scheduled_deletions;
+	set<idx_t> files_to_remove;
+	unordered_map<idx_t, idx_t> table_idx_last_snapshot;
+	// We can start by figuring out the files we can actually remove
+	for (int64_t i = static_cast<int64_t>(compactions.size()) - 1; i >= 0; i--) {
```
Mytherin: nit, but for these loops I think the following style is cleaner
```cpp
for(idx_t i = compactions.size(); i > 0; i--) {
   auto &compaction = compactions[i - 1];
}
```

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2194,6 +2203,113 @@ WHERE data_file_id IN (%s);
...
+	for (idx_t i = 0; i < compactions.size(); ++i) {
+		auto &compaction = compactions[i];
+		D_ASSERT(!compaction.path.empty());
+		auto path = GetRelativePath(compaction.delete_file_path);
+		if (files_to_remove.find(i) != files_to_remove.end()) {
+			// We only delete deletion files if they are part of the last snapshot, as they won't be required for
+			// time travel
+			scheduled_deletions += StringUtil::Format("(%d, %s, %s, NOW())", compaction.delete_file_id.index,
+			                                          SQLString(path.path), path.path_is_relative ? "true" : "false");
+			scheduled_deletions += ", ";
```
Mytherin: Also a nit, but I think this is nicer / less error prone:

```cpp
if (!scheduled_deletions.empty()) {
    scheduled_deletions += ", ";
}
scheduled_deletions += StringUtil::Format(...);
```

Then the `.pop_back()` at the end is not required

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2194,6 +2203,113 @@ WHERE data_file_id IN (%s);
...
+		                                              table_idx_last_snapshot[compaction.table_index.index],
+		                                              compaction.new_id.index));
+
+		if (result->HasError()) {
+			result->GetErrorObject().Throw("Failed to update snapshot end file information in DuckLake: ");
+		}
+	}
+	if (StringUtil::EndsWith(scheduled_deletions, ", ")) {
+		for (idx_t i = 0; i < 2; ++i) {
+			scheduled_deletions.pop_back();
```
Mytherin: This can go if the commas are set correctly initially

src/functions/ducklake_set_option.cpp
```diff
@@ -71,6 +71,12 @@ static unique_ptr<FunctionData> DuckLakeSetOptionBind(ClientContext &context, Ta
 		value = to_string(data_inlining_row_limit);
 	} else if (option == "require_commit_message") {
 		value = val.GetValue<bool>() ? "true" : "false";
+	} else if (option == "compaction_delete_threshold_rewrite") {
```
Mytherin: Maybe `compaction_delete_threshold` or `rewrite_delete_threshold`? Both `compaction` and `rewrite` seems unnecessary

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -295,15 +318,34 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
...
+					continue;
+				}
+				result.delete_file = source.delete_files.back().data;
+			}
+			break;
+		}
+		case MERGE_ADJACENT_TABLES: {
+			if (!source.delete_files.empty() && type == MERGE_ADJACENT_TABLES) {
+				// Merge Adjacent Tables does not support compaction
+				throw InternalException("FIXME: compact deletions");
```
Mytherin: Maybe the FIXME can now go since this is intended to stay, e.g. this should be `merge_adjacent_files should not be used to rewrite files with deletes`

src/include/storage/ducklake_metadata_info.hpp
pdet: Yeah, my bad

src/storage/ducklake_transaction.cpp
pdet: It should be ordered by the query that generates it, no?
`ORDER BY data.begin_snapshot, data.row_id_start, data.data_file_id, del.begin_snapshot` in `DuckLakeMetadataManager::GetFilesForCompaction(DuckLakeTableEntry &table,
                                                                                   CompactionType type,
                                                                                   double deletion_threshold)`

Mytherin:
Thanks!

PR #397, opened 2025-08-27
Add support for per thread output

J-Meyers:
Some of my usecases see very significant performance improvements when using `PER_THREAD_OUTPUT`, I wanted to migrate to using ducklake rather than just naively partitioned parquet files, but being > 4 times slower for writes wasn't acceptable.

This tries to add that option to ducklake, for my test I had to create a sample table to read from which was a bit funky but I found just using a `range` led to single threaded execution.

Renaming the option or allowing it to be more tied to the individual insertion both seem reasonable to me, but I wasn't sure what name you would prefer (either with the parquet prefix like the parquet specific ones, or without like the hive_partitioning one), or any other way of configuring options.

### DIFF
--- a/src/functions/ducklake_options.cpp
+++ b/src/functions/ducklake_options.cpp
@@ -12,7 +12,7 @@ struct DuckLakeOptionMetadata {
 	const char *description;
 };
 
-using ducklake_option_array = std::array<DuckLakeOptionMetadata, 14>;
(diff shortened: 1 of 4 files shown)


pdet:
Thanks for the PR! Code-wise LGTM, indeed there is the trade-off of potentially generating many small files, but as long as the default is false, I think it should be fine...

Mytherin:
Thanks!

Issue #300, opened 2025-07-17
Orphaned files on failed inserts


renre-axw:
Whilst performing an INSERT INTO FROM statement, the query apparently completes then when trying to COMMIT I get the following error.

100% ▕████████████████████████████████████████████████████████████▏ 
D COMMIT;
TransactionContext Error:
Failed to commit: Failed to execute query "ROLLBACK": 

I am then left with orphaned files in the S3 store and running cleanup CALL ducklake_cleanup_old_files('ducklake', cleanup_all => true); does not address these.

Mytherin:
Thanks for the report!

Oprhaned files are inevitable if insertions are cancelled before interacting with the catalog. We plan to add a method to remove orphaned files in the near future.

That said, in this case it seems like a problem that should be resolvable. It would help if you have a full reproducer so we could investigate further.

guillesd:
Hi @renre-axw if you have time to provide a reproducer I think this is a nice one to pick up!

ironman5366:
hi! I'm experiencing the same issues. In my case, I see it in extremely long inserts (~a few hours). It slowly climbs to 100% and then I see this exact message - `Failed to commit: Failed to execute query "ROLLBACK":`. 

In the most recent case this is happening when a table is being created, so I wouldn't expect there to be any logical conflict with existing data. I'm not sure whether just other operations run on other tables at the same time might cause issues.

renre-axw:
I experienced this issue whilst also performing a long insert.
I would be unable to provide the data used but suspect I can create fake data of similar size/nature and hopefully reproduce.
The setup was inserting records to S3, using an Amazon RDS Aurora DB for the Ducklake.

guillesd:
Hi @renre-axw if you can provide a reproducer with a very long failed insert, please provide it!

steven-luabase:
We also just experienced this while running a long insert (failed after ~40 minutes):
`TransactionContext Error: Failed to commit: Failed to execute query \"ROLLBACK\"`

I don't have a reproducible example just yet but we were running a DELETE then INSERT in a single transaction.

garyz-inspirum:
we are also seeing this error in our production logs from time to time. i don't really know how to re-produce it as it seems it is quite random. this error happens every a couple of hours. doesn't seem to be following a fix pattern. 

[...]

PR #398, opened 2025-08-28
Implementation of ducklake_delete_orphaned_files cleanup function

pdet:
This PR implements the `ducklake_delete_orphaned_files` function.

Orphaned files are files that might be created but not tracked by the metadata catalog (e.g., due to a system failure). The syntax is consistent with `ducklake_cleanup_old_files`. It can take up to three named parameters:
1. `older_than`: A value that defines how old the files must be to be taken into consideration for removal
2. `dry_run`:  Runs the algorithm but instead of removing the files returns a list so the user can double-check them.
3. `cleanup_all`: If all orphaned files should be removed regardless of how old they are.

Example call:
```sql
CALL ducklake_delete_orphaned_files('ducklake', older_than => now() - interval '1 day', dry_run => True);
```

The `older_than` default value is set by the `orphan_file_delete_older_than` option. If a default value is set, any option set manually with have higher precedence than the default.

Example usage:
```sql
CALL ducklake.set_option('orphan_file_delete_older_than', '1 week')
```
Fix (partially): https://github.com/duckdb/ducklake/issues/300

Fix: https://github.com/duckdblabs/duckdb-internal/issues/5693


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,97 @@ ORDER BY snapshot_id
...
+		DuckLakeFileForCleanup info;
+		info.path = row.GetValue<string>(0);
+		result.push_back(std::move(info));
+	}
+	return result;
+}
+
+vector<DuckLakeFileForCleanup> DuckLakeMetadataManager::GetFilesForCleanup(const string &filter, CleanupType type,
+                                                                           const string &data_path) {
+	string query;
```
Mytherin: nit: unused

Mytherin: We should enable clang-tidy and format-check CI on DuckLake - see https://github.com/duckdblabs/duckdb-internal/issues/5100

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,97 @@ ORDER BY snapshot_id
...
+	auto query = R"(SELECT filename
+FROM read_blob('{DATA_PATH}**')
+WHERE filename NOT IN (
+SELECT
+    CASE
+        WHEN NOT f.path_is_relative THEN f.path
+        ELSE
+            CASE
+                WHEN NOT t.path_is_relative THEN t.path || f.path
+                ELSE '{DATA_PATH}' || s.path || t.path || f.path
```
Mytherin: This seems to be missing a check for `s.path_is_relative`

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,97 @@ ORDER BY snapshot_id
...
+            END
+    END AS full_path
+FROM ducklake_delete_file f
```
Mytherin: Maybe we can run the union of `ducklake_delete_file` and `ducklake_data_file` before the join so we don't have to duplicate these joins and this path handling?

Mytherin (commented):
Thanks for the PR! Looks good - some minor comments

src/storage/ducklake_metadata_manager.cpp
pdet: I'll add the CI in a following PR, I think that's a good idea.

```
  code-quality-check:
    name: Code Quality Check
    uses: duckdb/extension-ci-tools/.github/workflows/_extension_code_quality.yml@main
    with:
      duckdb_version: v1.3.0
      ci_tools_version: main
      extension_name: quack
      format_checks: 'format;tidy'
```

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,101 @@ ORDER BY snapshot_id
...
+               ELSE CASE
+                        WHEN NOT table_relative THEN table_path || file_path
+                        ELSE CASE
+                                 WHEN NOT schema_relative THEN schema_path || table_path || file_path
+                                 ELSE '{DATA_PATH}' || schema_path || table_path || file_path
+                             END
+                   END
+           END,
+           '/',
+           '{SEPARATOR}'
```
Mytherin, line 2140: Separator handling shouldn't happen in the SQL query. All paths are defined to have `/` as a separator in DuckLake. You can use the `LoadPath` method to get the correct path when reading the result.

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,101 @@ ORDER BY snapshot_id
...
+        WHEN NOT f.path_is_relative THEN f.path
+        ELSE '{DATA_PATH}' || f.path
+    END ,
+           '/',
+           '{SEPARATOR}'
+) AS full_path
+FROM ducklake_files_scheduled_for_deletion f
+)
+)" + filter;
+	query = StringUtil::Replace(query, "{DATA_PATH}", data_path);
```
Mytherin: `{DATA_PATH}` is already handled in `DuckLakeTransaction::Query` - we shouldn't need this here separately

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,101 @@ ORDER BY snapshot_id
...
+UNION ALL
+SELECT REPLACE(
+    CASE
+        WHEN NOT f.path_is_relative THEN f.path
+        ELSE '{DATA_PATH}' || f.path
+    END ,
+           '/',
+           '{SEPARATOR}'
+) AS full_path
+FROM ducklake_files_scheduled_for_deletion f
```
Mytherin, line 2161: I'm missing a `{METADATA_CATALOG}` for all the table accesses here - but maybe that's no longer necessary now that we set the search path of the DuckLake transaction to only include the metadata catalog

src/storage/ducklake_metadata_manager.cpp
Mytherin: Actually instead of pushing `DATA_PATH` into the query at all - can't you just call `FromRelativePath` on the result?

pdet: Oh indeed `{DATA_PATH}` is handled in the `Query`, that simplifies things a bit.
I'm not sure I follow the `FromRelativePath` on the result, I need the correct paths properly set during query execution for the matcharoo of paths during query execution time, otherwise they won't match the filter properly.

Unless you are suggesting doing that in the cpp code?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2097,32 +2097,101 @@ ORDER BY snapshot_id
...
+                                 ELSE '{DATA_PATH}' || schema_path || table_path || file_path
+                             END
+                   END
+           END,
+           '/',
+           '{SEPARATOR}'
+       ) AS full_path
+FROM
+  (SELECT s.path AS schema_path, t.path AS table_path, f.path AS file_path, s.path_is_relative AS schema_relative, t.path_is_relative AS table_relative, f.path_is_relative AS file_relative
+   FROM ducklake_data_file f
```
Mytherin: Can we get rid of the two join chains and just union the data file and delete file?

Mytherin:
Thanks!

PR #406, opened 2025-09-01
DuckLake Checkpoint Function

pdet:
This PR introduces a DuckLake checkpoint function `CALL ducklake_checkpoint();` that calls all our compaction and cleanup functions on the current snapshot.

The functions are called in the following order:
1. `ducklake_flush_inlined_data`
2. `ducklake_expire_snapshots`
3. `ducklake_merge_adjacent_files`
4. `ducklake_rewrite_data_files`
5. `ducklake_cleanup_old_files`
6. `ducklake_delete_orphaned_files`

To use the `ducklake_checkpoint()` option is also important to set the default parameters that will be used for these functions. The following list of options should be set:
*  `rewrite_delete_threshold` : A threshold that determines the minimum amount of data that must be  removed from a file before a rewrite is warranted. From 0 - 1. Related to `ducklake_rewrite_data_files`.
* `delete_older_than`: How old unused files must be to be removed by the `ducklake_delete_orphaned_files` and  `ducklake_cleanup_old_files` cleanup functions.
* `expire_older_than`: "How old snapshots must be, by default, to be expired by: `ducklake_expire_snapshots`
* `compaction_schema`: "Pre-defined schema used as a default value for the following compaction functions  `ducklake_flush_inlined_data`,`ducklake_merge_adjacent_files`,  `ducklake_rewrite_data_files`, `ducklake_delete_orphaned_files`
* `compaction_table`: Pre-defined table used as a default value for the following compaction functions `ducklake_flush_inlined_data`,`ducklake_merge_adjacent_files`, `ducklake_rewrite_data_files`, `ducklake_delete_orphaned_files`

This PR also added some of the necessary options above.


pdet:
I've also added the `CHECKPOINT;` functionality.

src/storage/ducklake_transaction.cpp
```diff
@@ -361,9 +361,9 @@ void DuckLakeTransaction::WriteSnapshotChanges(DuckLakeCommitState &commit_state
 	AddChangeInfo(commit_state, change_info, changes.tables_inserted_inlined, "inlined_insert");
 	AddChangeInfo(commit_state, change_info, changes.tables_deleted_inlined, "inlined_delete");
 	AddChangeInfo(commit_state, change_info, changes.tables_flushed_inlined, "flushed_inlined");
-	if (!changes.tables_compacted.empty() && !change_info.changes_made.empty()) {
-		throw InvalidInputException("Transactions can either make changes OR perform compaction - not both");
-	}
+	// if (!changes.tables_compacted.empty() && !change_info.changes_made.empty()) {
+	// 	throw InvalidInputException("Transactions can either make changes OR perform compaction - not both");
```
Mytherin: nit: Can we fully remove this instead of just commenting it out?

Mytherin: Since we can now compact and do operations in the same transaction, there's also a bunch of tests that don't exist for this that should likely be written where these are interleaved, e.g.:

```sql
 BEGIN;
[INSERT]
[COMPACT]
COMMIT;
```

```sql
 BEGIN;
[COMPACT]
[INSERT]
[COMPACT]
[INSERT]
COMMIT;
```

```sql
 BEGIN;
[DELETE]
[COMPACT]
[INSERT]
[COMPACT]
COMMIT;
```

PR #409, opened 2025-09-03
Make DuckLake tests run with SQLite and Postgres as catalogs.

pdet:
Requires: https://github.com/duckdb/duckdb/pull/18847

### DIFF
--- a/test/configs/sqlite.json
+++ b/test/configs/sqlite.json
@@ -108,6 +108,14 @@
       ]
     },
     {
+      "reason": "FIXME:Cannot compare values of type VARCHAR and type TIMESTAMP WITH TIME ZONE ",
+      "paths": [
+        "test/sql/compaction/cleanup_old_files_global_option.test",
+        "test/sql/compaction/compaction_cleanup_global.test",
+        "test/sql/compaction/expire_snapshot_global_option.test"
+      ]
+    },
+    {
       "reason": "DuckDB as a catalog specific",
       "paths": [
         "test/sql/autoloading/autoload_data_path.test",


pdet:
@Maxxen This is the PR where we test Ducklake with postgres/sqlite catalogs, and have them as extension dependencies.
Maybe we can also add the spatial tests/dependency here?

.github/workflows/Catalogs.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: Setup Ccache
+      uses: hendrikmuhs/ccache-action@main
+      with:
+        key: ${{ github.job }}
+        save: ${{ github.ref == 'refs/heads/main' || github.repository != 'duckdb/duckdb-postgres' }}
+
+    - name: Setup vcpkg
+      uses: lukka/run-vcpkg@v11.1
+      with:
+        vcpkgGitCommitId: a42af01b72c28a8e1d7b48107b33e4f286a55ef6
```
carlopi:
```suggestion
        vcpkgGitCommitId: ce613c41372b23b1f51333815feb3edd87ef8a8b
```

We usually keep this commit everywhere

.github/workflows/Catalogs.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: Test Postgres
+      env:
+        PGSERVICE: postgres
+      run: |
+        createdb ducklakedb
+        build/release/test/unittest --test-config test/configs/postgres.json
+
+    - name: Test SQLite
+      run: |
+        build/release/test/unittest --test-config test/configs/sqlite.json
```
carlopi: Should we invert them? I feel like postgres does pollute the environment via `createdb`

carlopi (commented):
Minor tweaks, but this looks great

pdet:
@Mytherin maybe you want to have a quick look, before I press the button?

vcpkg.json
```diff
@@ -1,4 +1,5 @@
 {
   "dependencies": [
+    "openssl"
```
Mytherin: Why is this added?

test/configs/sqlite.json
```diff
@@ -0,0 +1,106 @@
+{
+  "description": "Run DuckLake tests using SQLite as a catalog DBMS.",
+  "on_init": "SET extension_directory = 'build/release/repository';",
```
Mytherin: Why is the `extension_directory` necessary?

test/sql/add_files/add_files_type_check_timestamp.test
```diff
@@ -11,8 +11,12 @@ require icu
 statement ok
 SET TimeZone='UTC';
 
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/
+
+test-env DUCKLAKE_CATALOG_DB ducklake_add_files_type_timestamp.db
+
 statement ok
-ATTACH 'ducklake:__TEST_DIR__/ducklake_add_files_type_timestamp.db' AS ducklake
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}${DUCKLAKE_CATALOG_DB}' AS ducklake  (DATA_PATH '__TEST_DIR__/ducklake_add_files_type_timestamp')
```
Mytherin: Is there a point to splitting this up into two separate env parameters? Wouldn't `${DUCKLAKE_CONNECTION}` be sufficient?

test/sql/add_files/add_files_type_check_string_blob.test
```diff
@@ -8,8 +8,12 @@ require parquet
 
 require json
 
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/
+
+test-env DUCKLAKE_CATALOG_DB ducklake_add_files_string_blob.db
+
 statement ok
-ATTACH 'ducklake:__TEST_DIR__/ducklake_add_files_string_blob.db' AS ducklake
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}${DUCKLAKE_CATALOG_DB}' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_add_files_string_blob')
```
Mytherin: Could be a future PR - but I think it might be useful to parameterize the `DATA_PATH` as well so we can more easily test e.g. data paths on S3

vcpkg.json
pdet: Because of postgres being a dependency for the tests, IIRC we need to combine the dependencies down, not sure if it's possible to wrap this in a parameter with our current infrastructure.

test/configs/sqlite.json
pdet: So DuckDB knows where to load extensions from, otherwise the setup fails.

test/sql/add_files/add_files_type_check_timestamp.test
pdet: Postgres sets them up as:
```
 "test_env": [{
      "env_name":  "DUCKLAKE_CONNECTION",
      "env_value": "postgres:"
   },
    {
      "env_name":  "DUCKLAKE_CATALOG_DB",
      "env_value": "dbname=ducklakedb"
   }],
``` 

SQLite only sets 
```
 "test_env": [{
      "env_name":  "DUCKLAKE_CONNECTION",
      "env_value": "sqlite:"
   },
```
since it can reuse the default `DUCKLAKE_CATALOG_DB` name.

test/sql/add_files/add_files_type_check_string_blob.test
pdet: Yeah, I think this is out of scope of this PR,

Got an issue to track this: https://github.com/duckdblabs/duckdb-internal/issues/5863

test/sql/add_files/add_files_type_check_timestamp.test
Mytherin: What we do in [the other config](https://github.com/duckdb/ducklake/blob/main/test/configs/attach_ducklake.json) is to use `{TEST_DIR}/{BASE_TEST_NAME}.db` as the database path. It might be worth using that here as well so we don't need to come up with a new database file for every single test we write (and risk forgetting to change it which breaks some CI sometimes).

vcpkg.json
Mytherin: I think we need to avoid this, I think this will be shipped in the final extension. We shouldn't ship what is effectively test code into the final extension that gets distributed.

test/configs/sqlite.json
Mytherin: That means that we can't run these tests in debug or reldebug mode, since we have hard-coded the extension repository to `build/release`, right? I think we need to figure out a better way to do this as well. Maybe @carlopi has some ideas.

carlopi: Could you try swapping:
```
  "on_init": "SET extension_directory = 'build/release/repository';"
```
with
```
   "autoloading": "available"
```
and see if that picks up the local extension directory auto-magically?

vcpkg.json
carlopi: This is a crappy problem (that we have also elsewhere, we basically just pile dependecy all together manually at the moment, like https://github.com/duckdb/duckdb-iceberg/blob/main/vcpkg.json, see dependence on avro is questionable).

@samansmink worked at some point on a solution.

What I do wasm side on the other hand is beautiful: override vcpkg.json, filtering out stuff. That is also not really OK.

I need to think  about this for a moment, but maybe openssl goes to a `test/vcpkg.json`, and as part of the test setup you do point to the other one (I am not really sure how, since it's hardoded, but maybe this demonstrate we need support for changing vcpkg.json location).

I need to think for a second what's the way out.

pdet: I've yolo-ed the CI by removing it, and it seems to work without it actually?

https://github.com/duckdb/ducklake/pull/409/commits/3732d8ad125eefac5dbd4280d317a20dc4c1d1c4

https://github.com/duckdb/ducklake/actions/runs/17638564235/job/50119993397?pr=409

I'm not a devops guy, but maybe this only really mettahs if more extensions dependencies have vcpkg dependencies?

test/configs/sqlite.json
pdet: You mean sth like:
```
{
  "description": "Run DuckLake tests using SQLite as a catalog DBMS.",
  "autoloading": "available",
...
``` 

?

If so, this unfortunately does not werk.

vcpkg.json
carlopi: If it works it works, solved

test/configs/sqlite.json
carlopi: autoloading all it is, let's see CI

vcpkg.json
samansmink: > I think we need to avoid this, I think this will be shipped in the final extension. We shouldn't ship what is effectively test code into the final extension that gets distributed.

Adding a dependency to vcpkg shouldn't mean that its included in the extension? Only if the extension links with it, I would say?

Extension CI tools does have support to specify test dependencies, see the delta extension here https://github.com/duckdb/duckdb-delta/blob/c249eaf529cb3d5ceb5b94c597b366783ed69385/Makefile#L25
which was added here https://github.com/duckdb/duckdb-delta/pull/222

pdet:
@Mytherin can you have another look?
I used the UUID trick for now, and some tests are not yet covered (e.g., if you have multiple attaches for the same db) which I think is fine, ~ 15 tests.

When we get the `{BASE_TEST_NAME}` in we can change that. 

Otherwise I think it uses all your feedback.

PR #415, opened 2025-09-05
Verify data_path fits the one in the catalog, unless override_data_path is set

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/5801


Mytherin:
Thanks! LGTM - just the Windows scenario needs some work

pdet:
> Thanks! LGTM - just the Windows scenario needs some work

I think is the path issue we talked about before.

PR #416, opened 2025-09-05
Rename ducklake_file_column_statistics TO ducklake_file_column_stats so it follows the same std of other stats tables.

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/5658


Mytherin:
Thanks!

PR #420, opened 2025-09-06
 Fix view aliases getting cleared out every time

NiclasHaderer:
The current behavior clears out the aliases every time, because the dummy view, does not contain any aliases. This fixes this by adding the aliases to the view creation. Alternatively I could also imagine not overwriting the aliases, that would also fix things for us.

---

Not having this fix will break aliased views for MotherDuck


Mytherin:
Thanks!

Issue #404, opened 2025-08-29
Suboptimal performance with `ducklake_add_data_files`


J-Meyers:
### What happens?

Within ducklake is there a planned method of paralellizing execution of CALL operations like `ducklake_add_data_files`?

While the function is way, way, way faster than a true INSERT operation, it still seems like it's running significantly more slowly than it could be.

Observing CPU usage and the code shows that all the core logic is single threaded and relies heavily on `parquet_metadata` which is only single threaded as well, even when providing a glob. I get that there are some IO limitations present here, but I'd still be surprised if there was no benefit to parallelizing it

In my operations (total query time 94 seconds) trying to add 1280 files:
- ReadParquetStats is the major bottleneck: 64 seconds (68% of total time)
  - Query: 36 seconds
  - Processing: 28 seconds
- ReadParquetSchema: 5 seconds
- ReadParquetFileMetadata: 5 seconds
- File processing phase: 20 seconds

Viewing the CPU usage with htop shows that it's not effectively using the resources on a 32 core machine. 

To me it seems that within this extension the capability to parallelize the basic processing operations would have a significant benefit, on top of the benefits of parallelizing various underlying duckdb operations. But I don't see this done at all so far in the extension so hope that y'all can lend some guidenance for how you want it organized and what patterns you expect for parallelism here.

The ability of ducklake to ingest extremely large data quantities (in the above case just 680 GB) in a reasonable timespan is an important part of being able to adopt ducklake and migrate from previously existing large datalakes of several hundred TB. Thinking through how to parallelize functions within ducklake, but outside of the database becomes relevant in these migrations. Additionally it is relevant with duckdb's suboptimal performance writing to s3 or nfs.

This issue has a larger affect the larger the dataset, so it's not as obvious with these examples because it's a bit difficult to artifically create an appropriately large dataset.

### To Reproduce

```sql
CALL dbgen(sf=<insertscalefactorhere you can withstand>);
COPY (SELECT *, EXTRACT(YEAR FROM o_orderdate) AS year, EXTRACT(MONTH FROM o_orderdate) AS month FROM orders) TO 'partitioned_orders.parquet' (PARTITION_BY (year, month), ROW_GROUP_SIZE 2_048);
```

```sql
INSTALL ducklake;
LOAD ducklake;
ATTACH 'ducklake:metadata.ducklake' AS ducklake;
USE ducklake;
CREATE TABLE orders_2 AS (FROM 'partitioned_orders.parquet' LIMIT 0);

[...]

PR #423, opened 2025-09-08
`ducklake_add_data_files` Simplify column stats processing

J-Meyers:
Replaces `ParquetColumnStats` with `DuckLakeColumnStats`, previously ducklake would convert from separate stats to a list of key value pairs back to separate stats, not doing that and instead just keeping as separate led to a significant speedup

On my sample query of 1280 files 680 GB which previously took 94 seconds
- Reduction by 35 seconds by switching from `ParquetColumnStats` to `DuckLakeColumnStats`
- Reduction by an additional 35 seconds possible by running with duckdb PR [parallel](https://github.com/duckdb/duckdb/pull/18854)

Works on https://github.com/duckdb/ducklake/issues/404

More speedups possible by doing the parquet file min max calculations in a duckdb native function, there's no real need to get the metadata for each row group to ducklake, doing the comparisons for min and max in a vectorized way would definitely help, also adding exporting the leaf_column_id within the `parquet_schema` function would make some logic easier over here I think)

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -284,13 +284,42 @@ FROM parquet_metadata(%s)
 				auto &name = bbox_child_types[child_idx].first;
 				auto &value = bbox_child_values[child_idx];
 				if (!value.IsNull()) {
-					stats.column_stats.push_back(GetStatsValue("bbox_" + name, value));
+					if (name == "xmax") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.xmax = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "xmin") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.xmin = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "ymax") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.ymax = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "ymin") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.ymin = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "zmax") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.zmax = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "zmin") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.zmin = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "mmax") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.mmax = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else if (name == "mmin") {
+						auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+						geo_stats.mmin = value.DefaultCastAs(LogicalType::DOUBLE).GetValue<double>();
+					} else {
+						throw InternalException("Unknown bbox child name %s", name);
+					}
 				}
 			}
 		}
 		if (!row.IsNull(8)) {
 			auto list_value = row.iterator.chunk->GetValue(8, row.row);
-			stats.column_stats.push_back(GetStatsValue("geo_types", std::move(list_value)));
+			auto &geo_stats = stats.extra_stats->Cast<DuckLakeColumnGeoStats>();
+			for (const auto &child : ListValue::GetChildren(list_value)) {
+				geo_stats.geo_types.insert(StringValue::Get(child));
+			}
 		}
 
 		column.column_stats.push_back(std::move(stats));
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -120,7 +120,6 @@ private:
 	unique_ptr<DuckLakeNameMapEntry> MapHiveColumn(ParquetFileMetadata &file_metadata, const DuckLakeFieldId &field_id,
 	                                               DuckLakeDataFile &result, const Value &hive_value);
 
-	Value GetStatsValue(string name, Value val);
 	void CheckMatchingType(const LogicalType &type, ParquetColumn &column);
 
 private:
@@ -227,13 +226,6 @@ FROM parquet_schema(%s)
 		auto &filename = file->filename;
 		parquet_files.emplace(filename, std::move(file));
 	}
-}
-
-Value DuckLakeFileProcessor::GetStatsValue(string name, Value val) {
-	child_list_t<Value> children;
-	children.emplace_back("key", Value(std::move(name)));
-	children.emplace_back("value", std::move(val));
-	return Value::STRUCT(std::move(children));
 }
 
 void DuckLakeFileProcessor::ReadParquetStats(const string &glob) {
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -132,7 +132,7 @@ private:
 	unordered_map<string, unique_ptr<ParquetFileMetadata>> parquet_files;
 };
 
-inline DuckLakeColumnGeoStats& MaybeInitializeStats(DuckLakeColumnStats &stats) {
+inline DuckLakeColumnGeoStats &MaybeInitializeStats(DuckLakeColumnStats &stats) {
 	if (!stats.extra_stats) {
 		stats.extra_stats = make_uniq<DuckLakeColumnGeoStats>();
 	}
@@ -141,8 +141,15 @@ inline DuckLakeColumnGeoStats& MaybeInitializeStats(DuckLakeColumnStats &stats) 
 
 void DuckLakeFileProcessor::ReadParquetSchema(const string &glob) {
 	auto result = transaction.Query(StringUtil::Format(R"(
-SELECT file_name, name, type, num_children, converted_type, scale, precision, field_id, logical_type
-FROM parquet_schema(%s)
+WITH base AS (
+  SELECT file_name, name, type, num_children, converted_type, scale, precision, field_id, logical_type
+  FROM parquet_schema(%s)
+),
+ordered AS (SELECT *, row_number() OVER () AS rn FROM base),
+partitioned AS (SELECT * EXCLUDE (rn), row_number() OVER (PARTITION BY file_name ORDER BY rn) - 1 AS flattened_column_id FROM ordered)
+SELECT * EXCLUDE (flattened_column_id)
+FROM partitioned
+ORDER BY file_name, flattened_column_id;
 )",
 	                                                   SQLString(glob)));
 	if (result->HasError()) {
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -230,7 +230,7 @@ FROM parquet_schema(%s)
 
 void DuckLakeFileProcessor::ReadParquetStats(const string &glob) {
 	auto result = transaction.Query(StringUtil::Format(R"(
-SELECT file_name, column_id, num_values, coalesce(stats_min, stats_min_value), coalesce(stats_max, stats_max_value), stats_null_count, total_compressed_size, geo_bbox, geo_types
+SELECT file_name, column_id, coalesce(stats_min, stats_min_value), coalesce(stats_max, stats_max_value), stats_null_count, total_compressed_size, geo_bbox, geo_types
 FROM parquet_metadata(%s)
 )",
 	                                                   SQLString(glob)));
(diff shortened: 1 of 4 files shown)


src/functions/ducklake_add_data_files.cpp
```diff
@@ -834,10 +850,24 @@ vector<DuckLakeDataFile> DuckLakeFileProcessor::AddFiles(const vector<string> &g
 	for (auto &glob : globs) {
 		// query the parquet_schema to figure out the schema for each of the columns
 		ReadParquetSchema(glob);
+
+		// turn off preserve_insertion_order to speed up the metadata queries, parquet_schema currently relies on
+		// ordered outputs
+		auto pres_false_res = transaction.Query("SET preserve_insertion_order=false");
```
pdet: @Mytherin is this a setting we could have by default?

src/functions/ducklake_add_data_files.cpp
```diff
@@ -264,22 +252,25 @@ FROM parquet_metadata(%s)
 			throw InvalidInputException("Column id not found in Parquet map?");
 		}
 		auto &column = column_entry->second.get();
-		ParquetColumnStats stats;
+		DuckLakeColumnStats stats(LogicalType::SQLNULL); // FIXME: Use the correct type here
```
pdet: Can't we set the correct type here?

src/functions/ducklake_add_data_files.cpp
```diff
@@ -264,22 +252,25 @@ FROM parquet_metadata(%s)
...
 
 		if (!row.IsNull(2)) {
 			// stats.column_stats.push_back(GetStatsValue("value_count", row.GetValue<string>(2)));
```
pdet: Why is this commented out? If not necessary maybe should just be removed?

src/functions/ducklake_add_data_files.cpp
```diff
@@ -834,10 +850,24 @@ vector<DuckLakeDataFile> DuckLakeFileProcessor::AddFiles(const vector<string> &g
...
 		// query the parquet_metadata to get the stats for each of the columns
 		ReadParquetStats(glob);
 		// read parquet file metadata
 		ReadParquetFileMetadata(glob);
+
+		// reset the setting
+		auto pres_true_res = transaction.Query("SET preserve_insertion_order=true");
+		if (pres_true_res->HasError()) {
+			pres_true_res->GetErrorObject().Throw("Failed to add data files to DuckLake: ");
+		}
```
pdet: I think we should probably verify what is the configuration set to see if we need to change it at all.

src/functions/ducklake_add_data_files.cpp
Mytherin: I don't think we should be changing this here, no. I wonder why this makes such a large impact on performance? It should not have that big of an impact.

pdet: According to OP is 35 out of 94 sec

J-Meyers: Without this it doesn't run the `parquet_metadata` and `parquet_file_metadata` queries in their parallel version for running over many files. That the queries were running single threaded was the initial explanation for the speed difficulties within `ducklake_add_data_files`. I should be clear that most of the speedup is in `parquet_metadata`, much more than `parquet_file_metadata`, but both do see a speedup. I don't know of another way to hint that it should run in paralllel. This speedup IS NOT actually currently observed unless running with my PR into duckdb to allow them to run in parallel: https://github.com/duckdb/duckdb/pull/18854

From my testing for the duckdb PR for the metadata query (there is addational overhead in the test, but you can see the majority of those 35 seconds here with the parquet_metadata):
Results prior: Run Time (s): real 34.280 user 33.362211 sys 0.893388
Results post: Run Time (s): real 2.938 user 77.055939 sys 4.093423

src/functions/ducklake_add_data_files.cpp
J-Meyers: I think right now `DeriveLogicalType` isn't the cheapest, so we try to only do it once per column per file (instead of per column per file per row group) and only do it when it goes to the `DuckLakeParquetTypeChecker` then we just cast all results to the actual logical type we are assigning the column, not these intermediate ones. This type (and the whole class column stats class) isn't used until `MergeStats` is actually called, so we set it right before that, at that point they all use the same type.

Although there is some trickiness in the constructor that initializes the extra_stats to geometry with specific inputs

src/functions/ducklake_add_data_files.cpp
Mytherin: I think this is something we should investigate separately rather than adding in this work-around. There is no reason for the system not to be able to run this in parallel also with `preserve_insertion_order = true`.

src/functions/ducklake_add_data_files.cpp
J-Meyers: I'm not sure why it was there or what the thinking was, it's been commented out for 3 months, presumably because the current `DuckLakeColumnStats` doesn't have a field for it. 

I'm happy to delete and remove from the query as well if that's what's wanted, but I wasn't sure if it was a part of some other future plan piece

src/functions/ducklake_add_data_files.cpp
J-Meyers: ~~It turns out I was definitely being dumb and got confused with some of my tests and recompilations, while the parallelism does have that effect, the `preserve_insertion_order = false` was not necessary (and was slightly counterproductive)~~

src/functions/ducklake_add_data_files.cpp
J-Meyers: Removed here: d113a46

src/functions/ducklake_add_data_files.cpp
J-Meyers: Removed this configuration setting

~~I was confused with some tests I had run, it is not necessary for the speedup~~ 
Update: https://github.com/duckdb/ducklake/pull/423#discussion_r2342746638

J-Meyers:
> Hi, thanks for the PR it is looking good, (also thanks for all the recent issues and PRs you have been opening, is much appreciated :-).
> 
> I had some small questions/requests.
> 
> Could we also add a micro-benchmark that benchmarks the parquet stats? (See `benchmarks/tpch` as an example). I think we should add regression tests in the near future, so it would be good to have :-)

@pdet

I tried adding a basic benchmark, but was having some difficulty with a proper cleanup or how to take advantage of caching. Right now both the .db files created and the exported parquets remain, so repeated runs would get upset, so I added OVERWRITE, but that seems more like a patch. Subesequent runs try and fully regenerate the tcph dataset.

I used some weird export settings to try and generate more row groups with a smaller dataset size, but didn't want insanely small files

Although the tests run quickly getting sufficient datasize is difficult and generating the dataset with `dbgen(sf=10)` takes a while

### Benchmark Results (scale factor 10)
Results with this branch + parquet_column_metadata branch:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.353124
benchmark/ingest/add_files_lineitem.benchmark	2	0.325152
benchmark/ingest/add_files_lineitem.benchmark	3	0.336028
benchmark/ingest/add_files_lineitem.benchmark	4	0.334177
benchmark/ingest/add_files_lineitem.benchmark	5	0.331770
```

Results with this branch + paralellism branch + `column_id` [handling in ducklake](https://github.com/duckdb/duckdb/pull/18854#issuecomment-3286314539):
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	1.085267
benchmark/ingest/add_files_lineitem.benchmark	2	1.046103
benchmark/ingest/add_files_lineitem.benchmark	3	1.047027
benchmark/ingest/add_files_lineitem.benchmark	4	1.035097
benchmark/ingest/add_files_lineitem.benchmark	5	1.039195
```

Results with this branch + paralellism branch:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	1.514555
benchmark/ingest/add_files_lineitem.benchmark	2	1.499116
benchmark/ingest/add_files_lineitem.benchmark	3	1.537237
benchmark/ingest/add_files_lineitem.benchmark	4	1.543666
benchmark/ingest/add_files_lineitem.benchmark	5	1.518092
```

Results with this branch alone:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	3.418199
benchmark/ingest/add_files_lineitem.benchmark	2	3.332479
benchmark/ingest/add_files_lineitem.benchmark	3	3.279108
benchmark/ingest/add_files_lineitem.benchmark	4	3.340530
benchmark/ingest/add_files_lineitem.benchmark	5	3.279764
```

Results with main for both:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	6.033699
benchmark/ingest/add_files_lineitem.benchmark	2	6.448958
benchmark/ingest/add_files_lineitem.benchmark	3	6.140097
benchmark/ingest/add_files_lineitem.benchmark	4	6.229260
benchmark/ingest/add_files_lineitem.benchmark	5	6.252945
```

### Benchmark Results (scale factor 1)
Results with this branch + parquet_column_metadata branch:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.064677
benchmark/ingest/add_files_lineitem.benchmark	2	0.064069
benchmark/ingest/add_files_lineitem.benchmark	3	0.064884
benchmark/ingest/add_files_lineitem.benchmark	4	0.059230
benchmark/ingest/add_files_lineitem.benchmark	5	0.063079
```

Results with this branch + paralellism branch + `column_id` [handling in ducklake](https://github.com/duckdb/duckdb/pull/18854#issuecomment-3286314539):
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.141883
benchmark/ingest/add_files_lineitem.benchmark	2	0.141867
benchmark/ingest/add_files_lineitem.benchmark	3	0.162295
benchmark/ingest/add_files_lineitem.benchmark	4	0.162835
benchmark/ingest/add_files_lineitem.benchmark	5	0.161237
```

Results with this branch + paralellism branch:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.213125
benchmark/ingest/add_files_lineitem.benchmark	2	0.202000
benchmark/ingest/add_files_lineitem.benchmark	3	0.211920
benchmark/ingest/add_files_lineitem.benchmark	4	0.204872
benchmark/ingest/add_files_lineitem.benchmark	5	0.203443
```

Results with this branch alone:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.349897
benchmark/ingest/add_files_lineitem.benchmark	2	0.362635
benchmark/ingest/add_files_lineitem.benchmark	3	0.366433
benchmark/ingest/add_files_lineitem.benchmark	4	0.366985
benchmark/ingest/add_files_lineitem.benchmark	5	0.363661
```

Results with main for both:
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.652084
benchmark/ingest/add_files_lineitem.benchmark	2	0.652727
benchmark/ingest/add_files_lineitem.benchmark	3	0.624346
benchmark/ingest/add_files_lineitem.benchmark	4	0.635499
benchmark/ingest/add_files_lineitem.benchmark	5	0.632442
```

src/functions/ducklake_add_data_files.cpp
J-Meyers: Okay, there's definetly a relation although in the benchmarks explicitly setting it hurt, in my real world use it helped, I think there is something more funky with the benchmark and it's sensitivity to the size of outputs
image
The incrementing of column_id within ReadParquetSchema relies on the ordering so that may cause issues. I'm not sure what guarentees there are with how it actually returns them.

I put a potential patch here: https://github.com/duckdb/duckdb/pull/18854#issuecomment-3283600103 which would resolve it by adding two new columns

Whether the setting is actually changed within the extension, having sensitivity to that variable means it should be able to be handled regardless.

This may only be an issue if duckdb when combining datachunks combines them in a non serial way, or if there are more than 2048 flat columns in a single file

J-Meyers: In addition to the very small additions mentioned in that patch I've also created a new parquet metadata function which does the aggregation itself:  
image
Right now for `ReadParquetStats` we have to transfer a lot of rows, and do the aggregation serially, pushing the aggregation into the parquet metadata functions both prevents parsing a lot of rows and parallelizes their aggregations, which you can see from the screenshot in the comment above still make up a significant amount of time, despite the optimizations in this PR.

Although relying on this new function has some backwards compatibility issues and adds another function to maintain, the code for it is all working (and wasn't that complex) and may be useful regardless.

Execution time with the optimized functions (and preserve_insertion_order false):
image

3.5 seconds

Execution time with the optimized functions (and preserve_insertion_order true):
image
13.6 seconds


Duckdb additional changes:
https://github.com/J-Meyers/duckdb/compare/parquet_metadata_multifile...J-Meyers:duckdb:parquet_column_metadata?expand=1
Ducklake additional changes:
https://github.com/J-Meyers/ducklake/compare/add_data_files_remove_parquet_column_stats...J-Meyers:ducklake:parquet_column_metadata?expand=1

Using only those in this PR:
image

Still reducing to ~24 seconds

Further weirdness with `preserve_insertion_order` I'm not sure what settings lead to the choices within duckdb deciding whether to parallelize or not, but it seems to be needed to be set to False to consistently parallelize the operations, although it happens to also paralellize operations on the further optimized branch, without setting it to false.

Without setting `preserve_insertion_order` to false on this branch:
image
It instead takes 1 minute

The query used:
```sql
ATTACH 'ducklake:metadata.ducklake' AS ducklake;
USE ducklake;
CREATE TABLE test0 AS (FROM 'data_source.parquet' LIMIT 0);
SET preserve_insertion_order = False;
.timer on
CALL ducklake_add_data_files('ducklake', 'test0', 'data_source.parquet');
```

J-Meyers:
CI failure seems unrelated [131/211] (62%): test/sql/concurrent/concurrent_insert_conflict.test            
Does not involve `ducklake_add_data_files`

src/functions/ducklake_add_data_files.cpp
pdet: @Maxxen would this be problematic for geo?

pdet:
I think is looking good, just want our geo specialist to check if the stats change won't impact his code.

Indeed the test failing seems unrelated, might be a spurious failure, I'll check if I can reproduce it on main.

w.r.t the benchmark, I believe that load is only executed once for the 5 runs, so having overwrite there seems fine if that's necessary, I would assume that the benchmark dir does get cleaned-up after runs, but in anycase having overwrite there doesn't hurt imo.:

```cpp
string InterpretedBenchmark::GetDatabasePath() {
    auto fs = FileSystem::CreateLocal();
    if (!cache_db.empty()) {
        return fs->JoinPath(BenchmarkRunner::DUCKDB_BENCHMARK_DIRECTORY, cache_db);
    }
    if (in_memory) {
        return "";
    }
    auto db_path = fs->JoinPath(BenchmarkRunner::DUCKDB_BENCHMARK_DIRECTORY, DEFAULT_DB_PATH);
    DeleteDatabase(db_path);
    return db_path;
}
```

src/functions/ducklake_add_data_files.cpp
```diff
@@ -137,10 +132,24 @@ struct DuckLakeFileProcessor {
 	unordered_map<string, unique_ptr<ParquetFileMetadata>> parquet_files;
 };
 
+inline DuckLakeColumnGeoStats &MaybeInitializeStats(DuckLakeColumnStats &stats) {
+	if (!stats.extra_stats) {
+		stats.extra_stats = make_uniq<DuckLakeColumnGeoStats>();
+	}
```
Maxxen: This seems a bit hacky... don't we know if the column is a geometry type at this point?

Maxxen: Ah I guess not... still I don't know if I like that this is hardcoded here, (although I guess geometry is the only thing that uses extra_stats). Id think its was cleaner if we initialized the geo stats explicitly when we parse them.

src/functions/ducklake_add_data_files.cpp
Maxxen: Not sure... there are others places (like when merging) where we assume the extra geo stats are instantiated. If they aren't, what happens if we add a parquet file which don't have geo stats set (for some reason?).

src/functions/ducklake_add_data_files.cpp
J-Meyers: Okay, it took a bit of work, but now I've moved when we know what type it is much earlier c1cdeeb.

This change should also make it so that we stop loading in files that are not compatible early on, rather than loading all of their metadata, then realizing that their schema was incompatible

src/functions/ducklake_add_data_files.cpp
J-Meyers: Changed to set correct type here as requested: c1cdeeb

J-Meyers:
Updated the benchmark results for my last two commits of changes and added in a field for what the performance could be with an additional function within duckdb `parquet_column_metadata` https://github.com/duckdb/ducklake/pull/423#discussion_r2343277922

While that function would massively speed this up, it isn't just made for ducklake and includes some columns to try and make it more generally useful to anyone

It should also be noted that the benchmark performance with paralellism is **very sensitive** to the decisions of duckdb to actually paralellize things, otherwise it gives no advantage, in real world use via cli I've found that I have to set `preserve_insertion_order` to false

Now that the parallesim has been added to main in duckdb, can I rely on the column `column_id` being present here?

pdet:
Hi @J-Meyers, 

LGTM, I ran it locally on 1.4, and there is a constant `40%` improvement.

Can you re-tag the PR to 1.4?

pdet:
Thanks!

PR #425, opened 2025-09-08
Revert {METADATA_CATALOG} removal

pdet:
Fix: https://github.com/duckdb/ducklake/issues/418


Mytherin:
We should then also add `{METADATA_CATALOG}.` to the new functions that did not have them - e.g. [GetOrphanFilesForCleanup](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_metadata_manager.cpp#L2178)

pdet:
> We should then also add `{METADATA_CATALOG}.` to the new functions that did not have them - e.g. [GetOrphanFilesForCleanup](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_metadata_manager.cpp#L2178)

That test seems to pass with postgres, maybe is just the creation of the tables that gets fuzzy? I'll add them anyway.

PR #426, opened 2025-09-08
Only allow the data_inlining_row_limit option if DuckDB is the catalog.

pdet:


src/include/storage/ducklake_catalog.hpp
```diff
@@ -186,7 +187,7 @@ class DuckLakeCatalog : public Catalog {
 	//! A unique tracker for catalog changes in uncommitted transactions.
 	atomic<idx_t> last_uncommitted_catalog_version;
 	//! The metadata server type
-	string metadata_type;
+	DuckLakeCatalogDBMS metadata_type;
```
Mytherin: This should not be an enum

PR #428, opened 2025-09-08
Format Python script

esadek:
Format Python script with default Ruff configuration.


pdet:
Hi @esadek thanks for the PR.

This script is actually not necessary anymore since we now run tests on the CI.

I've pushed a PR removing it from the repo:
https://github.com/duckdb/ducklake/pull/440

Cheers,

PR #429, opened 2025-09-09
Add ducklake_static_backup

carlopi:
Example, you can attach a given ducklake in read/write mode like:
```sql
ATTACH 'ducklake:postgres:<CONNECTION_STRING>' AS my_ducklake (DATA_PATH 's3://some_bucket', STATIC_BACKUP 's3://some_bucket/backup.ducklake');
--- some operations
CALL ducklake_static_backup('my_ducklake');
--- this will copy metadata to a duckdb file at s3://some_bucket/backup.ducklake
--- some more operations
CALL ducklake_static_backup('my_ducklake');
```
And at the same time, a fully static read-only backup will be accessible like:
```sql
ATTACH 'ducklake:s3://some_bucket/backup.ducklake' (READ_ONLY); ---- DATA_PATH will be the same as the main ducklake instance
```
(note that backup will be up to date to the most recent `ducklake_static_backup` call that happened BEFORE attaching it)

### DIFF
--- a/src/functions/ducklake_static_backup.cpp
+++ b/src/functions/ducklake_static_backup.cpp
@@ -1,0 +1,95 @@
+#include "functions/ducklake_table_functions.hpp"
+#include "duckdb/common/types/uuid.hpp"
+#include "duckdb/main/attached_database.hpp"
+#include "duckdb/main/database_manager.hpp"
+#include "storage/ducklake_catalog.hpp"
+#include "storage/ducklake_transaction.hpp"
+
+namespace duckdb {
+
+struct BackupBindData : public TableFunctionData {
+
+	explicit BackupBindData(Catalog &catalog) : catalog(catalog) {
+	}
+
+	Catalog &catalog;
+	string backup_location;
+};
+
+static unique_ptr<FunctionData> DuckLakeStaticBackupBind(ClientContext &context, TableFunctionBindInput &input,
+                                                         vector<LogicalType> &return_types, vector<string> &names) {
+	auto &catalog = BaseMetadataFunction::GetCatalog(context, input.inputs[0]);
+	auto result = make_uniq<BackupBindData>(catalog);
+
+	auto &ducklake_catalog = reinterpret_cast<DuckLakeCatalog &>(catalog);
+	string backup_location = ducklake_catalog.GetStaticBackup();
+
+	if (backup_location.empty()) {
+		throw InvalidInputException("static_backup not specified as attach option");
+	}
+
+	result->backup_location = backup_location;
+
+	return_types.emplace_back(LogicalType::VARCHAR);
+	names.emplace_back("errors");
+
+	return std::move(result);
+}
+
+struct DuckLakeBackupData : public GlobalTableFunctionState {
+	DuckLakeBackupData() : offset(0), executed(false) {
+	}
+
+	idx_t offset;
+	bool executed;
+};
+
+unique_ptr<GlobalTableFunctionState> DuckLakeStaticBackupInit(ClientContext &context, TableFunctionInitInput &input) {
+	auto result = make_uniq<DuckLakeBackupData>();
+	return std::move(result);
+}
+
+void DuckLakeStaticBackupExecute(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
+	auto &data = data_p.bind_data->Cast<BackupBindData>();
+	auto &state = data_p.global_state->Cast<DuckLakeBackupData>();
+
+	if (!state.executed) {
+		auto &transaction = DuckLakeTransaction::Get(context, data.catalog);
+
+		auto tmp_uuid = "ducklake_backup_file." + UUID::ToString(UUID::GenerateRandomUUID());
+
+		auto &fs = FileSystem::GetFileSystem(context);
+
+		if (fs.FileExists(tmp_uuid) || fs.FileExists(tmp_uuid + ".wal")) {
+			throw BinderException(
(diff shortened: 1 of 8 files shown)


src/functions/ducklake_static_backup.cpp
```diff
@@ -0,0 +1,95 @@
+#include "functions/ducklake_table_functions.hpp"
+#include "duckdb/common/types/uuid.hpp"
+#include "duckdb/main/attached_database.hpp"
+#include "duckdb/main/database_manager.hpp"
+#include "storage/ducklake_catalog.hpp"
+#include "storage/ducklake_transaction.hpp"
+
+namespace duckdb {
+
+struct BackupBindData : public TableFunctionData {
```
Tishj, line 10: please wrap these in an anonymous namespace to avoid name collisions
```
namespace {

...

} // namespace
```

src/functions/ducklake_static_backup.cpp
```diff
@@ -0,0 +1,95 @@
...
+	if (!state.executed) {
+		auto &transaction = DuckLakeTransaction::Get(context, data.catalog);
+
+		auto tmp_uuid = "ducklake_backup_file." + UUID::ToString(UUID::GenerateRandomUUID());
+
+		auto &fs = FileSystem::GetFileSystem(context);
+
+		if (fs.FileExists(tmp_uuid) || fs.FileExists(tmp_uuid + ".wal")) {
+			throw BinderException(
+			    "Temporary file \"%s\" is already in use, please cleanup files in the form \"ducklake_backup_file.*\"",
```
Tishj, line 65: This is generated, we can just regenerate if this is the case, no?

src/functions/ducklake_static_backup.cpp
```diff
@@ -0,0 +1,95 @@
...
+static unique_ptr<FunctionData> DuckLakeStaticBackupBind(ClientContext &context, TableFunctionBindInput &input,
+                                                         vector<LogicalType> &return_types, vector<string> &names) {
+	auto &catalog = BaseMetadataFunction::GetCatalog(context, input.inputs[0]);
+	auto result = make_uniq<BackupBindData>(catalog);
+
+	auto &ducklake_catalog = reinterpret_cast<DuckLakeCatalog &>(catalog);
+	string backup_location = ducklake_catalog.GetStaticBackup();
+
+	if (backup_location.empty()) {
+		throw InvalidInputException("static_backup not specified as attach option");
```
Tishj, line 28: Is this strictly necessary? Is it possible to just accept a second parameter to the function and use that if it's not defined on the catalog?

src/functions/ducklake_static_backup.cpp
```diff
@@ -0,0 +1,95 @@
...
+		auto &fs = FileSystem::GetFileSystem(context);
+
+		if (fs.FileExists(tmp_uuid) || fs.FileExists(tmp_uuid + ".wal")) {
+			throw BinderException(
+			    "Temporary file \"%s\" is already in use, please cleanup files in the form \"ducklake_backup_file.*\"",
+			    tmp_uuid);
+		}
+
+		auto result = transaction.Query(
+		    string("") + "ATTACH IF NOT EXISTS '" + tmp_uuid +
```
Tishj, line 70: Can we use:
```c++
StringUtil::Format(R"(
ATTACH IF NOT EXISTS '%s' AS {METADATA_CATALOG_NAME_IDENTIFIER_BACKUP} (STORAGE_VERSION 'v1.4.0');
COPY FROM DATABASE {METADATA_CATALOG_NAME_IDENTIFIER} TO {METADATA_CATALOG_NAME_IDENTIFIER_BACKUP};
DETACH {METADATA_CATALOG_NAME_IDENTIFIER_BACKUP};
COPY (SELECT content FROM read_blob('%s')) TO '%s' (FORMAT BLOB);
COPY (SELECT content FROM read_blob('%s.wal')) TO '%s.wal' (FORMAT BLOB);
)", ...);
```

src/functions/ducklake_static_backup.cpp
```diff
@@ -0,0 +1,95 @@
...
+	names.emplace_back("errors");
+
+	return std::move(result);
+}
+
+struct DuckLakeBackupData : public GlobalTableFunctionState {
+	DuckLakeBackupData() : offset(0), executed(false) {
+	}
+
+	idx_t offset;
```
Tishj, line 43: unused

Tishj:
This looks like it's missing the second half, a method to restore from the backup?

carlopi:
> This looks like it's missing the second half, a method to restore from the backup?

Thanks for the review!

This is on purpose, since the semantic it's not super clear, like what happens to stuff you have inserted or removed in the meantime?
That can always be performed out-of-band, but I think offering a functionally for that allows more misuse.

The backup it's intentended to simplify cases where the metadata catalog might for various reasons be non available to end-users, while the backup can be simply a couple of files on remote storage.
For example, a postgres metadata catalog it's great for parallel read/write, but a read-only duckdb file it's more portable and more accessible (for example, the DuckDB-Wasm client), so if one accept to be somewhat out-of-sync, the tradeoff might be worth it.

This is also meant to simplify quasi-frozen ducklake architectures, where one might add data with some cadence, but most workloads are read heavy.

Tishj:
I think if it's called `backup`, I would expect it to be a proper backup.
This sounds like it's only a metadata backup, and like you said, performing deletes could cause the backup to get entirely broken, if the deleted data gets garbage collected.

I think this would require proper branching support first, to be able to freeze a state of the table, preventing garbage collection of the data referenced by the backup

pdet:
Closing it for now due to the lack of activity.

PR #433, opened 2025-09-10
Use the transaction snapshot unless the file snapshot id differs from it

pdet:
Fixes a small issue found by the duckdb tests and the inclusion of the checkpoint functionality.

Test:
`test/sql/storage/compression/rle/rle_nulls_edge_case.test`

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -300,8 +300,10 @@ unique_ptr<LogicalOperator>
 DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry> source_files) {
 	// get the table entry at the specified snapshot
 	auto snapshot_id = source_files[0].file.begin_snapshot;
-	DuckLakeSnapshot snapshot(snapshot_id, source_files[0].schema_version, 0, 0);
-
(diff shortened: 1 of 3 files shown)


src/functions/ducklake_compaction_functions.cpp
```diff
@@ -300,8 +300,10 @@ unique_ptr<LogicalOperator>
 DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry> source_files) {
 	// get the table entry at the specified snapshot
 	auto snapshot_id = source_files[0].file.begin_snapshot;
-	DuckLakeSnapshot snapshot(snapshot_id, source_files[0].schema_version, 0, 0);
-
+	auto snapshot = transaction.GetSnapshot();
+	if (snapshot_id != snapshot.snapshot_id) {
+		snapshot = DuckLakeSnapshot(snapshot_id, source_files[0].schema_version, 0, 0);
```
Mytherin: I'm a bit puzzled as to what this solves - wouldn't we still always use `snapshot_id` here?

Does `GetSnapshot` force starting a snapshot and is that what solves the issue? (i.e. `(void) transaction.GetSnapshot()` would also solve the bug?)

PR #450, opened 2025-09-17
Check conflicts on a file-level basis instead of a table-level basis

pdet:


src/storage/ducklake_transaction.cpp
```diff
@@ -104,6 +104,7 @@ struct TransactionChangeInformation {
 	set<TableIndex> dropped_views;
 	set<TableIndex> tables_inserted_into;
 	set<TableIndex> tables_deleted_from;
+	set<DataFileIndex> files_deleted_from;
```
Mytherin: This could potentially grow to be a large list if we are deleting from many files. We should be able to figure out which files a transaction has deleted from by querying the tables - which doesn't require us to add more data to the change feed. Could we rework this to do that instead?

pdet: absolutely

src/storage/ducklake_transaction.cpp
```diff
@@ -476,9 +477,24 @@ void DuckLakeTransaction::CheckForConflicts(const TransactionChangeInformation &
 	for (auto &table_id : changes.tables_deleted_from) {
 		ConflictCheck(table_id, other_changes.dropped_tables, "delete from table", "dropped it");
 		ConflictCheck(table_id, other_changes.altered_tables, "delete from table", "altered it");
-		ConflictCheck(table_id, other_changes.tables_deleted_from, "delete from table", "deleted from it");
 		ConflictCheck(table_id, other_changes.tables_compacted, "delete from table", "compacted it");
 	}
+	if (!changes.tables_deleted_from.empty()) {
+		// If we have deletes on the tables check for files being deleted
+		const auto deleted_files = metadata_manager->GetFilesDeletedOrDroppedAfterSnapshot(transaction_snapshot);
```
Mytherin: Before we run this query can we first check if we have any potential conflicts based on the `tables_deleted_from`? If there is no overlap we can skip this query.

djouallah:
@pdet this is very big deal, does it mean it coming to 1.4 series :) say yes please

pdet:
> @pdet this is very big deal, does it mean it coming to 1.4 series :) say yes please

It will!

Not sure yet when we will do a release of a new DuckLake version for 1.4, but the basic idea is that if a feature or fix doesn't require a change in a feature level at DuckDB core, it will eventually go out for the 1.4. We will also have proper nightly builds of the extension.

PR #454, opened 2025-09-18
Unify data_inlining check

philippmd:
`DuckLakeInitializer::Initialize` checks whether the Catalog implements a DuckCatalog (`IsDuckCatalog()`); This check is a bit more permissive than checking whether the META_TYPE is empty.

In this PR we unify those checks to not run into situations where the DATA_INLINING_ROW_LIMIT can be set on startup but then not dynamically modified later.


pdet:
Thanks!

PR #458, opened 2025-09-19
Use Client Context FIleSystem in DuckLake transaction

Tmonster:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/5937

In a DuckLake transaction, there are multiple places where we attempt to remove files since their contents is invalidated by other statements within the transaction. We would use the database file system to do this. The database file system does not have access to the context, and therefore cannot access any secrets, which meant communication with s3 storage would fail. 

By using the Client Context file system, we now have access to the secrets, and can therefore delete the file. 
The test requires s3 communication, unsure if that is set up yet.

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -377,7 +377,9 @@ void DuckLakeTransaction::WriteSnapshotChanges(DuckLakeCommitState &commit_state
 
 void DuckLakeTransaction::CleanupFiles() {
 	// remove any files that were written
-	auto &fs = FileSystem::GetFileSystem(db);
+	// auto &fs = FileSystem::GetFileSystem(db);
+	auto context_ref = context.lock();
+	auto &fs = FileSystem::GetFileSystem(*context_ref);
 	for (auto &entry : table_data_changes) {
 		auto &table_changes = entry.second;
 		for (auto &file : table_changes.new_data_files) {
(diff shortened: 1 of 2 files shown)


src/storage/ducklake_transaction.cpp
```diff
@@ -377,6 +377,8 @@ void DuckLakeTransaction::WriteSnapshotChanges(DuckLakeCommitState &commit_state
 
 void DuckLakeTransaction::CleanupFiles() {
 	// remove any files that were written
+	// auto &fs = FileSystem::GetFileSystem(db);
```
pdet: nit: remove comment

src/storage/ducklake_transaction.cpp
```diff
@@ -1811,7 +1815,9 @@ void DuckLakeTransaction::TransactionLocalDelete(TableIndex table_id, const stri
 		if (file.file_name == data_file_path) {
 			if (file.delete_file) {
 				// this file already has a transaction-local delete file - delete it
-				auto &fs = FileSystem::GetFileSystem(db);
+				// auto &fs = FileSystem::GetFileSystem(db);
```
pdet: same

pdet (commented):
Hi @Tmonster Thanks! LGTM, could you re-tag it to the v1.4 branch?

Tmonster:
Updated the base branch. I thought we were still targeting main on extensions for weird CI/automatic extension building/distribution reasons?

pdet:
> Updated the base branch. I thought we were still targeting main on extensions for weird CI/automatic extension building/distribution reasons?

We had a quick meeting about this last week, so it seems that was a red herring.

Tmonster:
There isn't a CI test to this yet, so the `require-env` statements will automatically skip the test in this PR. Might be a good idea to test yourself before updating the extension. Is there a ticket for cloud testing for DuckLake (or any open table format)?

pdet:
> There isn't a CI test to this yet, so the `require-env` statements will automatically skip the test in this PR. Might be a good idea to test yourself before updating the extension. Is there a ticket for cloud testing for DuckLake (or any open table format)?

I'm going to update the data path of ducklake to be configurable, probably this week https://github.com/duckdblabs/duckdb-internal/issues/5863 .

Then it's a matter of setting up CI with MinIO.

No idea what is the plan for the other OTF extensions.

PR #465, opened 2025-09-23
Add Remote FS CI + Test Fixes

pdet:
This PRs changes our tests to execute with a configurable `DATA_PATH`. This allows all our tests to be executed with a remote file storage.

In addition to that, we now have a new CI with MinIO that simulates an S3 storage, executing all DuckLake tests there. This PR also has a few fixes for tests that were not working properly when executed in this format.

### DIFF
--- a/scripts/start-docker.sh
+++ b/scripts/start-docker.sh
@@ -30,5 +30,5 @@ for i in {1..30}; do
     break
   fi
   echo "Still waiting for MinIO..."
-  sleep 1
+  sleep 10
 done


pdet:
@carlopi could you have a quick look on the CI/MinIo setup?
I've also had to add `curl` as vcpkg dependency to get `httpfs` to work, I think this does not get bundled on the extension? (At least I remember @samansmink saying something similar to that, but maybe i got it wrong?) If so I might have to figure out some other way to get that going.

pdet:
cc @Tmonster now your test will run ;-)

vcpkg.json
```diff
@@ -1,4 +1,5 @@
 {
   "dependencies": [
+    "curl"
```
samansmink: Assuming curl is a dependency you need for httpfs, the cleaner approach would be to add to your makefile:

```Makefile
# Core extensions that we need for crucial testing
DEFAULT_TEST_EXTENSION_DEPS=
# For cloud testing we also need these extensions
FULL_TEST_EXTENSION_DEPS=httpfs

# Aws and Azure have vcpkg dependencies and therefore need vcpkg merging
ifeq (${BUILD_EXTENSION_TEST_DEPS}, full)
	USE_MERGED_VCPKG_MANIFEST:=1
endif
```

Then in CI jobs where you want httpfs built, you would set 
```sh
BUILD_EXTENSION_TEST_DEPS= full
```

This we automatically pull in the vcpkg manifest of httpfs, but only do so if we are actually building it for testing

carlopi:
Looks good.

Comments, possibly for follow ups: looking into the connection errors, hard to tell whether expected or not.
See the line like:
```
skip on error_message matching 'Unable to connect': 19
```
This is because by default we assume 'Unable to connect' to be an OK failure more (say networking), but on a local minio setup I would expect they should be accounted for, you can override default (that is skip on 'Unable to connect' or HTTP errors) by adding: `"skip_error_messages": []` to the configuration.

Second comment, this for later: do the same for duckdb tests.

.github/workflows/MinIO.yml
```diff
@@ -0,0 +1,93 @@
...
+      VCPKG_TARGET_TRIPLET: 'x64-linux-release'
+      GEN: ninja
+      VCPKG_FEATURE_FLAGS: "-binarycaching"
+      VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
+      PIP_BREAK_SYSTEM_PACKAGES: 1
+      BUILD_EXTENSION_TEST_DEPS: full
+
+    steps:
+      - name: Install required ubuntu packages
+        run: |
```
Tmonster, line 27: was this taken from DuckDB-Iceberg? I think a lot of this can be simplified. DuckDB-Iceberg CI also needs some cleaning up tbh

Tmonster: `wget`, `zip`, `openjdk-8-jdk` etc. I don't think are needed

pdet:
> Looks good.
> 
> Comments, possibly for follow ups: looking into the connection errors, hard to tell whether expected or not. See the line like:
> 
> ```
> skip on error_message matching 'Unable to connect': 19
> ```
> 
> This is because by default we assume 'Unable to connect' to be an OK failure more (say networking), but on a local minio setup I would expect they should be accounted for, you can override default (that is skip on 'Unable to connect' or HTTP errors) by adding: `"skip_error_messages": []` to the configuration.
> 
> Second comment, this for later: do the same for duckdb tests.

These haven't happened locally, but it's also a weird that this option is not opt-in

PR #469, opened 2025-09-23
Fix updating/merging into tables with geometries

### LINKED ISSUE #453: ALTER TABLE with GEOMETRY column fails with "Unsupported user-defined type" error in DuckLake
### What happens?

When attempting to add a GEOMETRY column to an existing table in DuckLake using ALTER TABLE ... ADD COLUMN, the operation fails with the error "Unsupported user-defined type". This occurs even when the spatial extension is loaded and the table already contains geometry columns.

### To Reproduce

Sample Code to Reproduce

```SQL
-- Setup: Install and load required extensions
INSTALL spatial; LOAD spatial;

-- Create a temporary directory for DuckLake metadata (required)
-- Note: In practice, you'd use a persistent path or PostgreSQL
ATTACH 'ducklake:temp_metadata.db' AS testlake;

-- Create schema and sample table with geometry data
CREATE SCHEMA IF NOT EXISTS testlake.test_schema;

-- Create initial table with geometry column
CREATE TABLE testlake.test_schema.buildings AS 
SELECT 
    'building_' || row_number() OVER () as id,
    ST_Point(random() * 1000000 + 500000, random() * 1000000 + 8000000) as geom,
    'residential' as building_type,
    floor(random() * 1000)::INTEGER as building_number
FROM range(10);

-- Verify table exists and has geometry data
SELECT * FROM testlake.test_schema.buildings LIMIT 5;

-- This fails with "Unsupported user-defined type" error:
ALTER TABLE testlake.test_schema.buildings ADD COLUMN geom_4326 GEOMETRY;
```

Expected Behavior: The ALTER TABLE statement should successfully add a new GEOMETRY column to the existing table.

Actual Behavior: The operation fails with "Invalid Input Error: Unsupported user-defined type".

Workaround: Currently, the only way to add geometry columns appears to be recreating the table with the desired schema.

### OS:

Windows 11 

### DuckDB Version:

1.4.0 (Andium)

### DuckLake Version:

0.3

### DuckDB Client:

CLI, Python and DBeaver

### Hardware:

_No response_

### Full Name:

Roman Slobodchikov

### Affiliation:

Norkart AS

I have not tested with any build

- [x] Yes, I have

- [x] Yes, I have

### LINKED ISSUE #461: MERGE INTO crashes when using ducklake 0.3 with geometry columns
### What happens?

When I try to update table containing `geometry` columns in ducklake it fails with:
```
Invalid Input Error:
Unexpected end of WKB data
```

Tables which don't have geometry columns work just fine 👍

### To Reproduce

```sql
ATTACH 'ducklake:my_ducklake.ducklake'
AS my_ducklake;
USE my_ducklake;

INSTALL spatial; LOAD spatial;
LOAD httpfs;

-- Initiate the monaco data from January 2014
CREATE TABLE monaco AS
SELECT
    ST_Point(lat, lon) as geometry,
    * EXCLUDE (lat, lon) REPLACE (kind::varchar AS kind, ref_types::varchar AS ref_types)
FROM ST_ReadOSM('https://download.geofabrik.de/europe/monaco-140101.osm.pbf')
WHERE kind = 'node';

-- Try to update the monaco with data from September 2025
WITH new_monaco AS (
    SELECT
        ST_Point(lat, lon) as geometry,
        * EXCLUDE (lat, lon) REPLACE (kind::varchar AS kind, ref_types::varchar AS ref_types)
    FROM ST_ReadOSM('https://download.geofabrik.de/europe/monaco-250901.osm.pbf')
)
MERGE INTO monaco
    USING new_monaco
    USING (kind, id)
WHEN MATCHED THEN UPDATE
WHEN NOT MATCHED THEN INSERT;
```

### OS:

MacOS 15.6.1 aarch64

### DuckDB Version:

v1.4.0 (Andium) b8a06e4a22

### DuckLake Version:

45788f0

### DuckDB Client:

CLI

### Hardware:

Macbook Pro M1 Pro

### Full Name:

Onni Hakala

### Affiliation:

Freelancer

Yes

- [x] Yes, I have

- [x] Yes, I have

Maxxen:
Closes #453 
Closes #461 

This PR fixes `MERGE INTO`/`UPDATE` of tables with geometry columns by properly casting geometry inputs to WKB_BLOB that the parquet extension can understand. Because there operators don't actually create a proper physical plan, but instead "emulate" the executor by just inserting directly into internal physical copy/insert/delete operators, managing their states manually, its not as easy as just inserting a new projection into the plan. Instead I've added intermediate "cast chunks' where necessary where the operators can cast their input to the types required by the physical copy-to-file, should a cast be necessary. 

Ideally this sort of casting would be done internally by the copy operator, or controlled by the copy function. In fact it is (its called `copy_to_select_t`, but since we don't go through a normal logical->physical planning stage pipeline the copy function never gets a chance to execute this callback, so we have to do this workaround instead.

I've also tried to generalize the notion of "types that require an intermediate cast" instead of hard coding checks for `GEOMETRY`. Maybe that could be useful for the future. 

I've also added a check that errors when trying to write geometry types nested within lists/structs. This doesn't currently work properly because we don't return the require geometry stats correctly when they're nested in the parquet extension. Im not sure why, but I think this is a remnant from the original GeoParquet implementation (since GeoParquet only supports geometry types as top-level columns). I'll look into fixing this, but it needs to be fixed in duckdb/duckdb.

### DIFF
--- a/src/include/common/ducklake_types.hpp
+++ b/src/include/common/ducklake_types.hpp
@@ -22,7 +22,7 @@ public:
 	static bool IsGeoType(const LogicalType &type);
 	static bool RequiresCast(const LogicalType &type);
 	static bool RequiresCast(const vector<LogicalType> &types);
-	// If this type requires a cast, return the type to cast to
+	//! If this type requires a cast, return the type to cast to
 	static LogicalType GetCastedType(const LogicalType &type);
 };
 
--- a/test/sql/cloud/test_cloud_cases.test
+++ b/test/sql/cloud/test_cloud_cases.test
@@ -2,32 +2,16 @@
 # description: test ducklake extension
 # group: [cloud]
 
-require-env CLOUD_TESTS_READY
-
-require-env AWS_KEY_ID
-
-require-env AWS_SECRET_KEY
-
-require-env AWS_REGION
-
 require ducklake
 
 require parquet
 
-require httpfs
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
 
 statement ok
-set s3_region='${AWS_REGION}';
-
-statement ok
-create secret foo (
-    type s3,
-    key_id '${AWS_KEY_ID}',
-    secret '${AWS_SECRET_KEY}'
-);
-
-statement ok
-attach 'ducklake:duckdb:' as lake (data_path 's3://test-ducklake-pedro/');
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS lake (DATA_PATH '${DATA_PATH}/test_cloud_cases/', METADATA_CATALOG 'metadata');
 
 statement ok
 use lake;
@@ -47,7 +31,7 @@ DELETE FROM test where i = 2;
 statement ok
 COMMIT;
 
-statement ok
+query I
 select * from test order by all;
 ----
 0
--- a/test/sql/geo/ducklake_geometry_merge.test
+++ b/test/sql/geo/ducklake_geometry_merge.test
@@ -9,6 +9,8 @@ require spatial
 
 test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
 
+test-env DATA_PATH __TEST_DIR__
+
 statement ok
 ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_files', METADATA_CATALOG 'ducklake_meta')
 
--- a/test/sql/geo/ducklake_geometry_nested.test
+++ b/test/sql/geo/ducklake_geometry_nested.test
@@ -9,6 +9,8 @@ require spatial
 
 test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
 
+test-env DATA_PATH __TEST_DIR__
+
 statement ok
 ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_files', METADATA_CATALOG 'ducklake_meta')
 


src/include/common/ducklake_types.hpp
```diff
@@ -18,6 +18,12 @@ class DuckLakeTypes {
 	static LogicalType FromString(const string &str);
 	static string ToString(const LogicalType &str);
 	static void CheckSupportedType(const LogicalType &type);
+
+	static bool IsGeoType(const LogicalType &type);
+	static bool RequiresCast(const LogicalType &type);
+	static bool RequiresCast(const vector<LogicalType> &types);
+	// If this type requires a cast, return the type to cast to
```
pdet: nit //!

test/sql/cloud/test_cloud_cases.test
```diff
@@ -31,7 +47,7 @@ DELETE FROM test where i = 2;
 statement ok
 COMMIT;
 
-query I
+statement ok
```
pdet: Was this an accident? And I mean all changes in this test file.

test/sql/geo/ducklake_geometry_merge.test
```diff
@@ -0,0 +1,61 @@
...
+require ducklake
+
+require parquet
+
+require spatial
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_files', METADATA_CATALOG 'ducklake_meta')
```
pdet, line 15: Can you also add test-env DATA_PATH __TEST_DIR__ ?

As this will allow us to one day in the future run geo tests on a different storage.

pdet (changes requested):
Hey Max, thanks for the PR! The code changes all LGTM.

Just have some minor requests on the testing side, and I think is good to go.

Also a good idea to split the PR in two since one will still need changes in main  (I guess V1.5?)

test/sql/cloud/test_cloud_cases.test
Maxxen: Yeah probably... my formatter added the `group: [cloud]` on a later commit but I must have ended up with a wonky merge when rebasing. Will revert!

pdet (approved):

PR #471, opened 2025-09-24
Adds `OR` and `IN` pushdown

### LINKED ISSUE #470: Filter pushdown with `OR` or `IN` fails
### What happens?

When ducklake is given a filter with an `OR` (or equivalent) it doesn't pushdown the predicate to limit the number of files read, and instead reads all files, this becomes a big issue with large highly partitioned datasets

`IN` pushdown also doesn't work, but that isn't as breaking as `OR` not working.

#### 1. Create partitoned dataset
```sql
ATTACH 'ducklake:bug_report_pushdown.db' AS dl;
USE dl;

CREATE TABLE partitioned_table (
  id BIGINT,
  data_column VARCHAR,
  partition_col VARCHAR,
  random_value DOUBLE
);

ALTER TABLE partitioned_table SET PARTITIONED BY (partition_col);

-- Insert data with 100 different partition values
INSERT INTO partitioned_table (
  SELECT 
      i as id,
      'value_' || (i % 1000) as data_column,
      'partition_' || LPAD((i % 100)::VARCHAR, 3, '0') as partition_col,  -- partition_000 to partition_099
      random() as random_value
  FROM range(50000) t(i)  -- 500 rows per partition
);
```

#### 2. Examine files
```sql
SELECT COUNT(*) as total_files FROM ducklake_list_files('dl', 'partitioned_table');
-- Expected: 100
```

##### Test 1: Single Partition Query (Expected: Scan only relevant files)
```sql
EXPLAIN ANALYZE 
SELECT COUNT(*) 
FROM partitioned_table 
WHERE partition_col = 'partition_001';
```
- Total Files Read: 1

##### Test 2: OR Query (Bug: Scans entire dataset)
```sql
EXPLAIN ANALYZE 
SELECT COUNT(*) 
FROM partitioned_table 
WHERE partition_col = 'partition_001' OR partition_col = 'partition_002';
```
- Expected Total Files Read: 2, actual: 100

##### Test 3: IN Query (Bug: Scans entire dataset)
```sql
EXPLAIN ANALYZE 
SELECT COUNT(*) 
FROM partitioned_table 
WHERE partition_col IN ('partition_001', 'partition_002', 'partition_003', 'partition_004', 'partition_005')
```
- Expected Total Files Read: 5, actual: 100

### OS:

linux x86_64

### DuckDB Version:

1.4.0

### DuckLake Version:

e986000

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Julian Meyers

### Affiliation:

Independent

Tested with main

- [x] Yes, I have

- [x] Yes, I have

J-Meyers:
Fixes #470

This failure to pushdown seemed significant enough that I wanted to target v1.4, but I'm not sure if that's correct
It had a disastrous performance impact on big datasets with restrictive filters.

While OR pushdown was technically handled before, it wasn't in practice because it was wrapped in a different filter first that wasn't handled.

The key to the OR being pushed down was that it was wrapped in an OPTIONAL_FILTER, which wasn't being processed

I don't know the significance of what makes some filters optional, but actually pushing them down really helps.

### DIFF
--- a/src/storage/ducklake_multi_file_list.cpp
+++ b/src/storage/ducklake_multi_file_list.cpp
@@ -16,6 +16,8 @@
 #include "duckdb/planner/filter/constant_filter.hpp"
 #include "duckdb/planner/filter/conjunction_filter.hpp"
 #include "duckdb/planner/filter/null_filter.hpp"
+#include "duckdb/planner/filter/optional_filter.hpp"
+#include "duckdb/planner/filter/in_filter.hpp"
 #include "storage/ducklake_table_entry.hpp"
 
 namespace duckdb {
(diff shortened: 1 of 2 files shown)


guillesd:
@pdet can you take a look? The issue is reproduced https://github.com/duckdb/ducklake/pull/471/files

pdet:
Thanks!

In general, we want to only have PRs to main if they either require a change in DuckDB core or necessitate a change in another extension that necessitates a change in DuckDB Core, so that was the right call.

PR #480, opened 2025-09-26
Minor fixup ci: attempt at having core_nightly automatically be supported also on nightly duckdb versions from v1.4-andium

carlopi:


pdet:
@carlopi, is this good to go?

carlopi:
I think so, but there is a CI failure that is unexpected. Could you retrigger to see if failures goes away? Or I can merge again? Or maybe it's transient, have not really seen something like that before.

carlopi:
Failure is quite weird, and likely due to an innocuous looking first commit bumping extension-ci-tools. I will reopen with CI changes only.

PR #485, opened 2025-09-29
Bump extension-ci-tools to main

carlopi:
Other half of https://github.com/duckdb/ducklake/pull/480, this I think will fail CI, and I have no idea how this is possible.


pdet:
Hi, the CI issue should be fixed, can you merge with V1.4 again?

PR #492, opened 2025-10-01
Acquire correct snapshot for inline tables depending on a table schema.

pdet:
cc @guillesd could you check if this solve your issue?


guillesd:
Yes will do this tomorrow morning!

PR #497, opened 2025-10-06
Fixing assertion errors and add CI run on debug.

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/6156
Fix: https://github.com/duckdblabs/duckdb-internal/issues/6154

### DIFF
--- a/.github/workflows/Debug.yml
+++ b/.github/workflows/Debug.yml
@@ -1,4 +1,4 @@
-name: RelAssert Mode Tests
+name: Debug Mode Tests
 on: [push, pull_request,repository_dispatch]
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || '' }}-${{ github.base_ref || '' }}-${{ github.ref != 'refs/heads/main' || github.sha }}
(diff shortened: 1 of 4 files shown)


pdet:
@carlopi I'm trying to run ducklake on debug mode on CI, but it seems that the machine can't handle the build process.
Is there any trick I'm missing here? That's already an ubuntu-latest, so I imagine that's prolly the more flexible machine that GH offers in the free tier?

I guess the other option would be to have a release run with `CRASH_ON_ASSERT` and `FORCE_ASSERT` set?

carlopi:
I am not aware of any easy fix out of the top of my head.
I think lowering parallelism (at compile or possibly at link time) do make sense, but it's a bit of trial and error.
Not super fun.

Also the problem I'd say is generic, you just got unlucky

pdet:
> I am not aware of any easy fix out of the top of my head. I think lowering parallelism (at compile or possibly at link time) do make sense, but it's a bit of trial and error. Not super fun.
> 
> Also the problem I'd say is generic, you just got unlucky

Yeah reducing parallelism only gets me to a no space left error, I've tried making a build with no extension dependencies, and also a relassert, but no luck, I'm wondering if you have any ideas of something else I could try? If that's a "we need a machine large enough" kind of problem, I guess we will have to make do without assertion checking on CI for now.

carlopi:
can you just try to swap linux-latest with macos-latest ?

samansmink:
> can you just try to swap linux-latest with macos-latest

I think using macos runners should not be taken lightly. We have only very few runners available

samansmink:
I think in the extension build someone added https://github.com/duckdb/extension-ci-tools/blob/cd519f5ba5707887f0f5282999d12ef32d1fe773/.github/workflows/_extension_distribution.yml#L242 at some point. Maybe that helps?

pdet:
Im giving a try to the removal of gunk in the image, let's see if that flows.

PR #500, opened 2025-10-07
Fix ducklake table rename

NiclasHaderer:
`ScanSchema` does not check for renamed tables. As a result it is possible for `schema.Scan` to return a table that does no longer exist


pdet:
Hi @NiclasHaderer, thanks for the PR! Couldn't the same happen for renamed schemas and views? Also, could you add a test for that?

NiclasHaderer:
Schemas cannot be renamed as far as I know ?? and views [work differently](https://github.com/NiclasHaderer/ducklake/blob/afacc42e3d60a2fb365fd8d7c8c9716e1b43edb1/src/storage/ducklake_transaction.cpp#L2013-L2033), so they are good

pdet:
LTGM! Thanks!

PR #501, opened 2025-10-08
Only support native geometry types

Maxxen:


pdet:
Hi @Maxxen can you merge with `v1.4-andium` again?

PR #502, opened 2025-10-08
1.4 -> Main

pdet:


pdet:
@Maxxen it seems we are getting tons of:
`INTERNAL Error: Unrecognized option for PARQUET: geoparquet_version`, is that because DuckDB main is not yet up to date with `V1.4.1` or am I missing something?

PR #506, opened 2025-10-09
Macro support

pdet:
This draft is a work in progress for macro support in DuckLake.

It currently implements 3 tables to represent a macro in the catalog, with multiple implementations and multiple parameters.
```sql
CREATE TABLE {METADATA_CATALOG}.ducklake_macro(schema_id BIGINT, macro_id BIGINT, macro_name VARCHAR, begin_snapshot BIGINT, end_snapshot BIGINT);
CREATE TABLE {METADATA_CATALOG}.ducklake_macro_impl(macro_id BIGINT, impl_id BIGINT, dialect VARCHAR, sql VARCHAR, type VARCHAR);
CREATE TABLE {METADATA_CATALOG}.ducklake_macro_parameters(macro_id BIGINT, impl_id BIGINT,column_id BIGINT, parameter_name VARCHAR, parameter_type VARCHAR, default_value VARCHAR, default_value_type VARCHAR);

``` 

The PR supports Scalar and Table Macros. Macros can also have multiple implementations. Default parameters and type definitions are also supported.

Time travel is possible when attaching, as macros will be properly set to a given snapshot.

Schema dependency on macro is maintained, but this PR does not track other dependencies yet (e.g., views, other macros, ...).

A new ducklake version is also created in this PR `0.4-dev1`.

DuckDB Macro tests are also enabled for ducklake.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -36,7 +36,7 @@ DuckLakeCompaction::DuckLakeCompaction(PhysicalPlan &physical_plan, const vector
 // GetData
 //===--------------------------------------------------------------------===//
 SourceResultType DuckLakeCompaction::GetDataInternal(ExecutionContext &context, DataChunk &chunk,
-                                             OperatorSourceInput &input) const {
+                                                     OperatorSourceInput &input) const {
 	return SourceResultType::FINISHED;
 }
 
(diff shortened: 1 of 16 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1168,6 +1250,50 @@ void DuckLakeMetadataManager::WriteNewInlinedTables(DuckLakeSnapshot commit_snap
 	ExecuteInlinedTableQueries(commit_snapshot, inlined_tables, inlined_table_queries);
 }
 
+void DuckLakeMetadataManager::CreateMacro(DuckLakeSnapshot snapshot, MacroIndex index) {
+	// auto result = transaction.Query(snapshot,"SELECT name")
```
Mytherin: Probably unused?

Issue #509, opened 2025-10-11
Alter table rename fails if table used to exist


Alex-Monahan:
### What happens?

Howdy folks! 

When using dbt, DuckLake 0.3 runs into an error after the second run. (As far as I know, this makes dbt effectively broken for DuckLake 0.3). 

It is a similar error message to https://github.com/duckdb/ducklake/issues/508, but the error should not occur (and does not occur on DuckDB 1.3.2 and DuckLake 0.2). 

### To Reproduce

This is a pure SQL reproducer of the issue. It occurs on DuckDB 1.4.1, with both the released version of DuckLake and one compiled from the latest main of DuckLake. 

```sql
install ducklake;
load ducklake;
ATTACH 'ducklake:ducklake_catalog_8.duckdb' as my_ducklake (data_path 'ducklake_duckdb_files_8/');
use my_ducklake;
create schema main_raw;
use main_raw;
create table if not exists "my_ducklake"."main_raw"."customer" as (
  select * from range(100)
);
BEGIN TRANSACTION;
create table "my_ducklake"."main_raw"."customer__dbt_tmp" as (
  select * from range(100)
);
alter table "my_ducklake"."main_raw"."customer" rename to "customer__dbt_backup";
alter table "my_ducklake"."main_raw"."customer__dbt_tmp" rename to "customer";
commit;
```
This results in:
```sql
Binder Error:
Cannot rename table customer__dbt_tmp to customer, since customer__dbt_tmp already exists.
```

### OS:

MacOS

### DuckDB Version:

1.4.1

### DuckLake Version:

0.3 and d7f7908099985117dce3e3e858791571454e762b

### DuckDB Client:

CLI and dbt/Python

### Hardware:

N/A, but M1 Macbook Pro 16GB RAM

### Full Name:

Alex Monahan

### Affiliation:

MotherDuck

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

guillesd:
Let's see if #510 is a quick fix! I'm just looking for a second pair of eyes to see if this fixes it.

slygren:
[...]

PR #510, opened 2025-10-13
Fixed more rename errors

NiclasHaderer:
Currently it is not possible to do atomic table name swap operations:

```sql
BEGIN
CREATE TABLE my_table_tmp AS FROM range(84);
ALTER TABLE my_table RENAME TO my_table_backup;
ALTER TABLE my_table_tmp RENAME TO my_table;
-- Above query results in: "Cannot rename table my_table_tmp to my_table, since my_table already exists."
COMMIT
```

this fixes this, by filtering out renamed tables on catalog lookups

---

Fixes https://github.com/duckdb/ducklake/issues/509

### DIFF
--- a/.github/workflows/Debug.yml
+++ b/.github/workflows/Debug.yml
@@ -1,0 +1,73 @@
+name: Debug Mode Tests
+on: [push, pull_request,repository_dispatch]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || '' }}-${{ github.base_ref || '' }}-${{ github.ref != 'refs/heads/main' || github.sha }}
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ducklake-config-tests:
+    name: DuckLake (Debug)
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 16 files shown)


guillesd:
Hi @NiclasHaderer, thanks for the PR. A couple of points:
1. Maybe you could also fix the error trace, since this is what you already complain about in your issue: (so change `alter.table` for `new_table->name` in the last string interpolation)
2. The test seems overly long and unnecessary, no? In my opinion this issue repro https://github.com/duckdb/ducklake/issues/509 is all you need. Not really follow why you are doing the counts.

Mytherin:
Thanks! LGTM

slygren:
Not sure if the merged code is out and made into an installable ducklake extension. If so - the reported issue is very much live and still existing for 1.4.1

NiclasHaderer:
The latest extension was released from https://github.com/duckdb/ducklake/commit/f134ad8, so this fix has not been released yet.

You can always find out which extension version you have installed by force installing the extension you're looking for and then checking the version using the `duckdb_extensions` table function.

guillesd:
Hi @slygren @NiclasHaderer ! There will be a patch release in 3 weeks. In the meantime you can `force install ducklake from core_nightly` where the fix is already up!

PR #477, opened 2025-09-25
Add `ComplexFilterPushdown` support

J-Meyers:
This adds support for relatively complex filters being pushed down via `ComplexFilterPushdown`
Followup to #471

### Motivation:
As mentioned in my previous PR even relatively simple queries currently can't be pushed down without this
```sql
FROM partitoned_table
WHERE (partition_col_0 = 0 AND partition_col_1 = 1) OR partition_col_1 = 2
```

Where the manifesto says
```
Hidden Partitioning and Pruning: DuckLake is aware of data partitioning and table- and file-level statistics, allowing for early pruning of scans for maximum efficiency.
```
But it doesn't currently actually apply in a variety of cases, to the point of actually worse performance than hive structured files (I think) in some cases because the base duckdb will pushdown complex filters with hive

With optimizations like this I don't think any fair/realistic benchmarks can be made since it's just 1/however big N you want

### Design:

The general design is for each expression try and pass them through the FilterCombiner since that can do a lot of work, then if there are leftovers recurse into the ORs and ANDs, if the combiner was able to extract useful groups of functions then parse them as we did before. Propogate when pieces are unsatisfiable, but in general keep as many restrictions as possible. Since we're interacting in SQL we can use that to handle whatever arbitrary expression trees we want rather than just being restricted to single column filters as in TableFilterSet.

The guiding principle is to have the catalog do as much work as possible given the expected balance between the size of ducklake (massive) and the size of data for common queries (small enough for a single machine).

This will likely add some tiny overhead on queries where the filters don't actually eliminate any files from being included but are complicated queries, particularly when the data in ducklake is of comparable size to the metadata (which seems incredibly rare).

Annoyingly, the actual creation of the queries has to be delayed a bit because we may later receive a dynamic pushdown, we want that to still keep all our old more complex filters and not add new filters unless otherwise necessary.

### Performance:
When running the generated query via CLI on realistic datasets just running the query via cli takes < .05 seconds
**However** instead when logging the time within `DuckLakeTransaction::Query` when running the relevant user query the generated query takes ~0.25 seconds I'm not sure why there is this disconnect, but it is pretty significant

These numbers are when running with duckdb as the metadata database, the performance may be different on different metadata stores, but still should be faster than reading hundreds of parquet files.

[...]

PR #513, opened 2025-10-14
Add basic != pushdown and parens

J-Meyers:
Takes some changes from #477 as getting it to be reasonable and generic within duckdb is proving to be difficult rather than just simple sql handling. Trying to break that down into small PRs like this as much as I can, but keeping it up for history.

This adds != filtering, which should be useful on partitioned datasets where this is the partition column (or was at the time of writing), adds parenthesis for some clarity and to prevent any additional issues arising from order of operations as the function is eventually used in more complex contexts.

I checked and support for pushing down not equals predicates was explicitly added to duckdb-delta

Adds support for only parsing one side of an AND expression, I'm not sure how needed it is without the complex parsing (I couldn't find a repro without it), but logically it seems more consistent to always try and filter out as much as possible, rather than rely on subsequent calls from the generate expressions which parses each of them out.

### DIFF
--- a/test/sql/stats/filter_pushdown.test
+++ b/test/sql/stats/filter_pushdown.test
@@ -157,11 +157,11 @@ FROM range(501000, 501001) t(i);
 
 # != (not equal) pushdown - should prune file with constant value
 query I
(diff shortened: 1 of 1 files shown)


pdet:
Thanks! LGTM!

Issue #504, opened 2025-10-09
order by queries on inlined data failing


jamesmstone:
### What happens?

Trying to query data that has been inlined returns an error : `IO Error: Cannot open file "ducklake_inlined_data_1_1": No such file or directory  `

### To Reproduce

```
SQL_CMDS=$(cat <<'EOF'  
ATTACH 'ducklake:duck.duckdb' AS data ( DATA_INLINING_ROW_LIMIT 10000);  
select version(), extension_version as ducklake FROM duckdb_extensions() where extension_name ='ducklake'; ;  
  
CREATE TABLE IF NOT EXISTS data.example_table (  
created_at TIMESTAMPTZ,  
version INTEGER  
);  
INSERT INTO data.example_table (created_at, version) VALUES (now(),1);  
  
SELECT version FROM data.example_table ORDER BY created_at DESC LIMIT 1;  
EOF  
)  
  
echo "$SQL_CMDS" | duckdb :memory: -bail -echo  
```
I get
```
ATTACH 'ducklake:duck.duckdb' AS data ( DATA_INLINING_ROW_LIMIT 10000);  
select version(), extension_version as ducklake FROM duckdb_extensions() where extension_name ='ducklake'; ;  
┌─────────────┬──────────┐  
│ "version"() │ ducklake │  
│ varchar │ varchar │  
├─────────────┼──────────┤  
│ v1.4.1 │ f134ad8 │  
└─────────────┴──────────┘  
CREATE TABLE IF NOT EXISTS data.example_table (  
created_at TIMESTAMPTZ,  
version INTEGER  
);  
INSERT INTO data.example_table (created_at, version) VALUES (now(),1);  
IO Error: Cannot open file "ducklake_inlined_data_1_1": No such file or directory  
SELECT version FROM data.example_table ORDER BY created_at DESC LIMIT 1;  
```

### OS:

linux x86_64

### DuckDB Version:

1.4.1

### DuckLake Version:

f134ad8

### DuckDB Client:

duckdb-cli ( and also go client)

### Full Name:

James Stone

### Affiliation:

-

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

[...]

PR #514, opened 2025-10-14
Filter CTE Generation

J-Meyers:
This modifies the `GetFilesForTable` and `DuckLakeMultiFileList` to first generate CTEs for each of the specific column stats.

This does not change the actual filters that are being run, it only changes the shape of the query. It includes logic to explicitly `materialize` the CTE if there are multiple references, and explicitly says `not materialized` when it is only 1 rather than leaving it up to the optimizer. Since it is currently run on a single TableFilter the reference count is always 1 and they are never marked as materialized.

Part of the continuation of breaking up #477 into more reviewable chunks as suggested

This makes a material difference on query time only when the same column is referenced multiple times, as can happen with complex filters (which are not yet processed) because in ducklakes with lots of table with lots of columns a very significant portion of the total time is just looking up the column stats, not comparing them, so doing that multiple times really hurts query execution time.

Sample user query:
```sql
SELECT COUNT(*) FROM ducklake.filter_pushdown WHERE i IN (500, 600, 700);
```

Sample generated query:
```sql
WITH col_2_stats AS NOT MATERIALIZED (
  SELECT data_file_id, max_value, min_value
  FROM {METADATA_CATALOG}.ducklake_file_column_stats
  WHERE column_id = 2 AND table_id = 1
)

SELECT data.path, data.path_is_relative, data.file_size_bytes, data.footer_size, data.row_id_start, data.begin_snapshot, data.partial_file_info, data.mapping_id, del.path, del.path_is_relative, del.file_size_bytes, del.footer_size
FROM {METADATA_CATALOG}.ducklake_data_file data
LEFT JOIN (
    SELECT *
    FROM {METADATA_CATALOG}.ducklake_delete_file
    WHERE table_id=1  AND 4 >= begin_snapshot
          AND (4 < end_snapshot OR end_snapshot IS NULL)
    ) del USING (data_file_id)
WHERE data.table_id=1 AND 4 >= data.begin_snapshot AND (4 < data.end_snapshot OR data.end_snapshot IS NULL)

AND data_file_id IN (SELECT data_file_id FROM col_2_stats WHERE max_value IS NULL OR min_value IS NULL OR (500 BETWEEN TRY_CAST(min_value AS INTEGER) AND TRY_CAST(max_value AS INTEGER) OR 600 BETWEEN TRY_CAST(min_value AS INTEGER) AND TRY_CAST(max_value AS INTEGER) OR 700 BETWEEN TRY_CAST(min_value AS INTEGER) AND TRY_CAST(max_value AS INTEGER)))
```


J-Meyers:
I'm a bit uncertain about what you want here. Should all of the core logic of converting a `TableFilterSet` to SQL (and the associated CTEs to SQL) be moved into the metadata manager, including all the logic of `GenerateFilterPushdown`?

If so, I'll wait for the other simple PR to get in otherwise I'll have merge conflicts moving all 300 lines.

Mytherin:
Yes, ideally no SQL gets generated outside of the metadata manager, and we just move structs describing the operations into the metadata manager. This rule is broken somewhat through the filter already - ideally that is not a SQL string but a set of structs as well.

src/storage/ducklake_multi_file_list.cpp
```diff
@@ -395,11 +192,18 @@ OpenFileInfo DuckLakeMultiFileList::GetFile(idx_t i) {
...
+	}
+
+	auto result = make_uniq<DuckLakeMultiFileList>(read_info, transaction_local_files, transaction_local_data,
+	                                               std::move(filter_copy));
 	result->files = GetFiles();
 	result->read_file_list = read_file_list;
 	result->delete_scans = delete_scans;
-	return std::move(result);
+	result->inlined_data_tables = inlined_data_tables;
+	return result;
```
J-Meyers, line 206: This copy here I believe was key in fixing https://github.com/duckdb/ducklake/issues/504
It is resolved with this branch as far as I can tell

pdet: Can you add a test to represent #504, if this fixes it already?

pdet (commented):
Thanks for the contribution, that looks great!

One extra request, could you get the CI we discussed in #477 for testing the filters, where we store 1 row per file and run the duckdb filter tests on it?

J-Meyers:
Current failure is unrelated caused separately here: https://github.com/duckdb/ducklake/pull/535#issuecomment-3519512380

pdet:
Thanks!

PR #518, opened 2025-10-15
Enable postgres inlining and add tests

aclowes:
Edit: I think this is a better approach: https://github.com/duckdb/ducklake/pull/526

Instead of a blanket disable, can we enable it with caveats?

I'd be interested in fixing the data type issues but it seems an easy solution isn't available. I wonder if writing parquet files in bytea columns to the postgres database is a reasonable approach as data types won't be a problem and the scan methods could be adapted in a relatively easy manner? Just an idea and wondering if you all have another approach you are planning.


test/sql/transaction/transaction_conflict_inlining_postgres.test
```diff
@@ -0,0 +1,102 @@
...
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION} options=''-c lock_timeout=1''' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_transaction_conflict_files', ENCRYPTED, DATA_INLINING_ROW_LIMIT 10000)
```
aclowes, line 15: I can't figure out how to have the "default" test configuration skip these tests, ie when duckdb is being used as the metadata catalog.

pdet:
Hi @aclowes, thanks for the PR.

Supporting other DBMSs for inlining is challenging when data types do not exist in the catalog system. That's why the `test/sql/data_inlining/data_inlining_types.test` test still fails.

I believe that the correct way of doing this is to store types that do not exist as a string, and then converting it to the right type when flushing/querying.

aclowes:
@pdet thanks for the feedback! I considered that approach, in particular using JSON serialization, but it seems it would require a lot of custom work and a separate interface for DuckDB vs Postgres (and future catalog stores), so instead I serialized the whole row in the internal format here: https://github.com/duckdb/ducklake/pull/526. I'll close this PR.

Issue #293, opened 2025-07-15
Data Inlining for non-DuckDB metadata catalogs


zwhitten:
I was trying to set up a DuckLake setup with Postgres as the Metadata DB and ran across an error saying data inlining is currently only supported for DuckDB ? 
Is there a plan to expand this capability to other metadata databases in the future?

Thanks

guillesd:
Hi @zwhitten this is planned to be supported. I can't give you a timeline yet though!

zwhitten:
Ok understood. Thanks @guillesd !

geoHeil:
Intereesting observation.

```sql
ATTACH 'ducklake:duckpond_ducklake' AS duckpond (DATA_INLINING_ROW_LIMIT 10);
```

indeed fails as explained before.

But this combination:

```sql
ATTACH 'ducklake:duckpond_ducklake' AS duckpond;
CALL duckpond.set_option('data_inlining_row_limit', '10');
```

i.e. first attach without inlining and then store persistently to catalog seems to work.

I do not know though how well/production grade this works.

guillesd:
Hi @geoHeil ! This would be a different issue. This issue is concretely about data inlining support for other catalogs (not only DuckDB). What you are raising is regarding the initializing parameter not working https://ducklake.select/docs/preview/duckdb/usage/connecting#parameters, but the option does work http://ducklake.select/docs/preview/duckdb/usage/configuration#setting-config-values-1. If this is the case could you create a separate issue including the reproducible example?

geoHeil:
Now I am confused. As you say `ATTACH 'ducklake:duckpond_ducklake' AS duckpond (DATA_INLINING_ROW_LIMIT 10);` should work - which (reproducibly) fails for me. However, when using 


```sql
ATTACH 'ducklake:duckpond_ducklake' AS duckpond;
CALL duckpond.set_option('data_inlining_row_limit', '10');
```

things seem to work fine. So is this now a problem only with the parameter handling? Or as originally deemed something which still has to be implemented?

See:


```sql
CREATE OR REPLACE SECRET duckpond_secret_catalog (
    TYPE postgres,
    HOST 'db.mycorp.com,
    PORT 5433,
    DATABASE ducklake_catalog,
    USER getenv('CATALOG_DB_USER'),
    PASSWORD getenv('CATALOG_DB_PASSWORD')
);

CREATE OR REPLACE SECRET duckpond_secret_storage (
  TYPE S3,
[...]

PR #526, opened 2025-10-20
Switch inline storage to serialization

aclowes:
This switches the inlined data storage method from native tables to serialized DataChunks. This should improve Postgres compatibility and performance.

Performance: inserting 100k rows similar to `test/sql/data_inlining/data_inlining_large.test` took 3.6s with the existing code and 0.2s with the new code.

Postgres compatibility: all tests pass when running `build/release/test/unittest --test-config test/configs/postgres.json`

Fixes https://github.com/duckdb/ducklake/issues/293

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1324,7 +1324,9 @@ WHERE table_id = %d AND schema_version=(
 			// Create column names array for SQL
 			string column_names_sql = "ARRAY[";
 			for (idx_t i = 0; i < column_names.size(); i++) {
-				if (i > 0) column_names_sql += ", ";
+				if (i > 0) {
+					column_names_sql += ", ";
+				}
 				column_names_sql += "'" + column_names[i] + "'";
 			}
 			column_names_sql += "]";
@@ -1432,9 +1434,6 @@ shared_ptr<DuckLakeInlinedData> DuckLakeMetadataManager::TransformInlinedData(Qu
 		}
 	}
 	
-	// For deletion scans, we need to reconstruct the original data that was deleted
-	// We'll allow all columns but need to handle them specially
-	
 	// First pass: collect deleted row IDs and their end_snapshots from delete chunks
 	map<idx_t, int64_t> deleted_row_end_snapshots; // Map row_id to end_snapshot
 	
@@ -1454,23 +1453,17 @@ shared_ptr<DuckLakeInlinedData> DuckLakeMetadataManager::TransformInlinedData(Qu
 			// Convert from base64 to binary data
 			auto blob_data = Blob::FromBase64(string_t(serialized_blob_base64));
 			
+			// Create owning memory stream and write blob data directly to it
+			MemoryStream mem_stream(blob_data.size());
+			mem_stream.WriteData(const_data_ptr_cast(blob_data.data()), blob_data.size());
+			mem_stream.Rewind(); // Reset position to beginning for reading
+			
 			// Deserialize the DataChunk directly from the blob data
 			auto chunk_ptr = make_uniq<DataChunk>();
-			
-			// Copy blob data to a writable buffer for MemoryStream (needs to outlive deserializer)
-			auto blob_size = blob_data.size();
-			auto blob_buffer = make_unsafe_uniq_array<data_t>(blob_size);
-			memcpy(blob_buffer.get(), blob_data.data(), blob_size);
-			
-			{
-				// Create memory stream from the blob data
-				MemoryStream mem_stream(blob_buffer.get(), blob_size);
-				BinaryDeserializer deserializer(mem_stream);
-				
-				deserializer.Begin();
-				chunk_ptr->Deserialize(deserializer);
-				deserializer.End();
-			}
+			BinaryDeserializer deserializer(mem_stream);			
+			deserializer.Begin();
+			chunk_ptr->Deserialize(deserializer);
+			deserializer.End();
 			
 			// Parse column names from the value
 			vector<string> chunk_column_names;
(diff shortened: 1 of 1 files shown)


test/configs/postgres.json
```diff
@@ -15,33 +15,9 @@
...
-        "test/sql/data_inlining/data_inlining_types.test",
-        "test/sql/data_inlining/data_inlining_update.test",
-        "test/sql/transaction/transaction_conflict_inlining.test",
-        "test/sql/data_inlining/inlining_global_options.test",
-        "test/sql/checkpoint/checkpoint_updates_interleaved.test",
-        "test/sql/checkpoint/checkpoint_ducklake.test",
-        "test/sql/checkpoint/many_inserts_transaction.test",
-        "test/sql/data_inlining/insert_inlining_concurrent.test",
-        "test/sql/transaction/transaction_inlining.test"
+        "test/sql/transaction/transaction_conflict_inlining.test"
```
aclowes, line 20: This is a tricky one. Instead of throwing an exception immediately, Postgres hangs on the conflicting second statement until the first statement is committed. I think the intent of the test still passes, but I don't know how to write a test that works for both behaviors.

szarnyasg:
Hi @aclowes, thanks a lot for the PR! I build this PR for DuckDB v1.4.1 and tested it using a local Postgres installation with 1000 inserts. For some reason, it was slow (25 seconds!) but it did the inlining and flushing upon `CHECKPOINT`, producing a single file as expected.

Here's my script: `demo-ducklake.py`

```python
import duckdb

con = duckdb.connect(config={"allow_unsigned_extensions": "true"})

print(duckdb.__version__)
print(duckdb.sql("PRAGMA version"))
print(duckdb.sql("select extension_version from duckdb_extensions() where extension_name = 'ducklake';"))

con.sql("ATTACH 'ducklake:postgres:dbname=postgres' AS my_ducklake (DATA_INLINING_ROW_LIMIT 20_000, DATA_PATH 'my_ducklake_dir);")
con.sql("USE my_ducklake")
con.sql("CREATE TABLE t(i INT)")

for idx in range(0, 1000):
    con.sql(f"INSERT INTO t VALUES ({idx})")
con.sql("CHECKPOINT")
```
```bash
time python demo-ducklake.py
```
```console
1.4.1
┌─────────────────┬────────────┬──────────┐
│ library_version │ source_id  │ codename │
│     varchar     │  varchar   │ varchar  │
├─────────────────┼────────────┼──────────┤
│ v1.4.1          │ b390a7c376 │ Andium   │
└─────────────────┴────────────┴──────────┘

┌───────────────────┐
│ extension_version │
│      varchar      │
├───────────────────┤
│ a26a769           │
└───────────────────┘

python demo-ducklake.py  10.81s user 1.65s system 49% cpu 24.938 total
```

aclowes:
Hi @szarnyasg, many thanks for taking a look. I'm not surprised 1k inserts in a loop is slow, but I think it's in line with the current performance.

I'm not sure how you are loading the extension in python, when I try something like this it fails, having run `pip install duckdb` to install the regular build:
```
con.execute("INSTALL '../ducklake/build/release/extension/ducklake/ducklake.duckdb_extension';") 

_duckdb.IOException: IO Error: Failed to install '../ducklake/build/release/extension/ducklake/ducklake.duckdb_extension'
The file was built specifically for DuckDB version '8f43b9862d' and can only be loaded with that version of DuckDB. (this version of DuckDB is 'v1.4.1')
```

Instead, I tested using the sql tests. I modified the existing test at `test/sql/data_inlining/data_inlining_large.test` to be as follows, starting at line 17 after the ATTACH statement:
```
statement ok
CALL dbgen(sf = 0.1)

statement ok
CREATE TABLE ducklake.customer AS FROM customer LIMIT 0

loop i 0 1000

statement ok
INSERT INTO ducklake.customer FROM customer LIMIT 10 OFFSET ${i} * 10

endloop

query I
SELECT COUNT(*) FROM ducklake.customer
----
10000
```

To mimic larger insert batch sizes, I also tested this, which inserts the data in batches of 1000 rows instead: 
```
statement ok
CALL dbgen(sf = 0.1)

statement ok
CREATE TABLE ducklake.customer AS FROM customer LIMIT 0

loop i 0 10

statement ok
INSERT INTO ducklake.customer FROM customer LIMIT 1000 OFFSET ${i} * 1000

endloop

query I
SELECT COUNT(*) FROM ducklake.customer
----
10000
```

I'm running these two test cases using `time build/release/test/unittest test/sql/data_inlining/data_inlining_large.test` for duckdb as a catalog and `time build/release/test/unittest --test-config test/configs/postgres.json test/sql/data_inlining/data_inlining_large.test` to use postgres as a catalog. 

The second axis is the inlining method. The base case is at git hash `c39873b4676c125770ee98ce72f401d8256aff7c` where I've re-enabled Postgres as a metadata catalog but before the serialization implementation. The serialization test is at the head of this PR which currently `a26a7698efc3b6b0738464e7ba5f3fc2149473e2`.

Here is the grid of the results:

| Inlining method | Test case | DuckDB | Postgres |
|---|---|---|---|
| As-is | Batches of 10 | 9 sec | 12 sec |
| As-is | Batches of 1000 | 0.9 sec | 1.2 sec |
| Serialization | Batches of 10 | 10 sec | 11 sec |
| Serialization | Batches of 1000 | 0.6 sec | 0.9 sec |

I think maybe the benefit of not parsing a large SQL statement means it's faster most of the time. And it allows postgres to be used as a catalog again.

pdet:
Hi @aclowes thanks for your PR, we really appreciate outside contributions! Unfortunately we wanted to go in a different direction, where we store the inlined types that do not exist in the catalog system either by approximation (e.g., if it's a uint8 but the catalog only supports int64, we go with int64) and in the worst case as a varchar.

This allows us to skip serialization/deserialization costs, while still being sufficiently agnostic about the catalog system.

You can see the PR here:
https://github.com/duckdb/ducklake/pull/556

Hence, I'm then closing this PR, in the future, please feel free to contact me directly if you have any question on what to contribute to DuckDB to ensure that your contribution also fits our designs/goals!

aclowes:
@pdet thanks for the feedback! I'm not surprised you took another approach as serialization seemed hacky, although it was versatile.

PR #535, opened 2025-10-30
Use `parquet_full_metadata`

J-Meyers:
Continuation of https://github.com/duckdb/duckdb/pull/19148 and https://github.com/duckdb/ducklake/pull/423

This speeds up the operations noticeably, though not as far as a dedicated `parquet_column_metadata` function, lots of time is spent in basically moving the data in and out of `DataChunks` in particular for the `parquet_metadata`.

Key optimizations:
- Use `parquet_full_metadata` for a single parsing of the metadata, this is benificial when the files are remote and the cache is not enabled, when the files are local, filesystem caching reducesthe significance of this unless you have a truly huge data size dominated by metadata
- Switch to direct access of the data through the `DataChunk` rather than calling `GetValue` this made a very significant reduction, bringing the time to read and unpack the data to be usable from 8.5 seconds to 3.3 seconds
- Improved column stat aggregation calculations, basically moved the code from `parquet_column_metadata` to here which caches conversion to `Value` for comparisons and processes all the mins, then all the maxs etc for easier short circuiting (change from 4.6 seconds to 2.4 seconds) (this could be trivially parallelized if there was an interface within ducklake for that)

### New timing
Time breakdown(preserve insertion order false required):
```
Parquet full metadata query took 2201 ms
Parquet file metadata processing took 0 ms
Parquet schema processing took 0 ms
Parquet metadata processing took 156 ms
Reading parquet file metadata took 3210 ms
Mapping and adding files took 2410 ms
Total execution time: 6.762 seconds
```
### Previous:
Total execution time: 13.642 seconds
### With `column_metadata`:
3.551

Times are taken from the .timer on, not the time from the progress bar which tends to under-rerport by a few seconds

[Old time reference](https://github.com/duckdb/ducklake/pull/423#discussion_r2343277922)

I suspect this changes together should significantly decrease the total execution time to create frozen ducklakes as described in a recent blog (particularly relative with the cache off, or in other scenarios of loading more than just a few files): https://ducklake.select/2025/10/24/frozen-ducklake/#performance

Note that in this blog it was neglected to turn preserve_insertion_order to false which makes a massive difference in execution time in most cases

#### This branch
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.095825
benchmark/ingest/add_files_lineitem.benchmark	2	0.095197
benchmark/ingest/add_files_lineitem.benchmark	3	0.089405
benchmark/ingest/add_files_lineitem.benchmark	4	0.094142
benchmark/ingest/add_files_lineitem.benchmark	5	0.092781
```
#### Reference previous sf1 best possible (take advantage of added column_id)
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.141883
benchmark/ingest/add_files_lineitem.benchmark	2	0.141867
benchmark/ingest/add_files_lineitem.benchmark	3	0.162295
benchmark/ingest/add_files_lineitem.benchmark	4	0.162835
benchmark/ingest/add_files_lineitem.benchmark	5	0.161237
```
#### Reference current
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.212660
benchmark/ingest/add_files_lineitem.benchmark	2	0.213065
benchmark/ingest/add_files_lineitem.benchmark	3	0.215464
benchmark/ingest/add_files_lineitem.benchmark	4	0.217041
benchmark/ingest/add_files_lineitem.benchmark	5	0.217098
```
#### This branch
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	0.602087
benchmark/ingest/add_files_lineitem.benchmark	2	0.596061
benchmark/ingest/add_files_lineitem.benchmark	3	0.600540
benchmark/ingest/add_files_lineitem.benchmark	4	0.578238
benchmark/ingest/add_files_lineitem.benchmark	5	0.577673
```
#### Reference previous sf10 best possible (take advantage of added column_id)
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	1.085267
benchmark/ingest/add_files_lineitem.benchmark	2	1.046103
benchmark/ingest/add_files_lineitem.benchmark	3	1.047027
benchmark/ingest/add_files_lineitem.benchmark	4	1.035097
benchmark/ingest/add_files_lineitem.benchmark	5	1.039195
```
#### Reference current
```
name	run	timing
benchmark/ingest/add_files_lineitem.benchmark	1	1.618826
benchmark/ingest/add_files_lineitem.benchmark	2	1.585683
benchmark/ingest/add_files_lineitem.benchmark	3	1.589488
benchmark/ingest/add_files_lineitem.benchmark	4	1.587740
benchmark/ingest/add_files_lineitem.benchmark	5	1.576632
```
Note that the optimization for column_id is a relatively simple change, although it's not implemented here because this is now better and since their fully associated it's no longer necessary.

Reference times from last pr:
https://github.com/duckdb/ducklake/pull/423#issuecomment-3282525122

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -211,39 +211,29 @@ FROM parquet_full_metadata(%s)
 		auto parquet_schema_offset = parquet_schema_entry.offset;
 		auto parquet_schema_length = parquet_schema_entry.length;
 
-		ParquetFileMetadata *file_ptr = nullptr;
+		// Extract filename from the file metadata struct
+		auto &struct_children = StructVector::GetEntries(file_metadata_list_entries);
+		idx_t struct_idx = file_metadata_offset;
 
-		// File metadata - access struct fields directly
-		{
-			// file_metadata_entries is a struct vector, get its child vectors
-			auto &struct_children = StructVector::GetEntries(file_metadata_list_entries);
+		auto filename =
+		    FlatVector::GetData<string_t>(*struct_children[0])[struct_idx].GetString(); // struct field: file_name
 
-			// Access individual fields directly from the struct child vectors
-			// We only process the first entry (index 0 within the list)
-			idx_t struct_idx = file_metadata_offset; // First entry in the list
-
-			auto &filename_vec = *struct_children[0];
-			auto filename = FlatVector::GetData<string_t>(filename_vec)[struct_idx].GetString();
-
-			auto &found_file_ptr = parquet_files[filename];
-			if (!found_file_ptr) {
-				found_file_ptr = make_uniq<ParquetFileMetadata>();
-			} else {
-				// Already initialized, this file must be found in multiple globs
-				continue;
-			}
-			file_ptr = found_file_ptr.get();
-			file_ptr->filename = filename;
-
-			auto &num_rows_vec = *struct_children[1];
-			auto &file_size_vec = *struct_children[2];
-			auto &footer_size_vec = *struct_children[3];
-
-			file_ptr->row_count = FlatVector::GetData<int64_t>(num_rows_vec)[struct_idx];
-			file_ptr->file_size_bytes = FlatVector::GetData<uint64_t>(file_size_vec)[struct_idx];
-			file_ptr->footer_size = FlatVector::GetData<uint64_t>(footer_size_vec)[struct_idx];
+		// Check if we've already processed this file (can happen with overlapping globs)
+		auto &file_metadata_ptr = parquet_files[filename];
+		if (file_metadata_ptr) {
+			// File already processed in a previous glob, skip
+			continue;
 		}
-		auto &file = *file_ptr;
+
+		// Initialize new file metadata entry
+		file_metadata_ptr = make_uniq<ParquetFileMetadata>();
+		auto &file = *file_metadata_ptr;
+		file.filename = filename;
+
+		file.row_count = FlatVector::GetData<int64_t>(*struct_children[1])[struct_idx]; // struct field: num_rows
+		file.file_size_bytes =
+		    FlatVector::GetData<uint64_t>(*struct_children[2])[struct_idx]; // struct field: file_size_bytes
+		file.footer_size = FlatVector::GetData<uint64_t>(*struct_children[3])[struct_idx]; // struct field: footer_size
 
 		bool saw_root = false;
 		vector<idx_t> child_counts;
--- a/src/include/storage/ducklake_inlined_data_reader.hpp
+++ b/src/include/storage/ducklake_inlined_data_reader.hpp
@@ -30,8 +30,8 @@ public:
 public:
 	bool TryInitializeScan(ClientContext &context, GlobalTableFunctionState &gstate,
 	                       LocalTableFunctionState &lstate) override;
-	AsyncResult Scan(ClientContext &context, GlobalTableFunctionState &global_state, LocalTableFunctionState &local_state,
-	          DataChunk &chunk) override;
+	AsyncResult Scan(ClientContext &context, GlobalTableFunctionState &global_state,
+	                 LocalTableFunctionState &local_state, DataChunk &chunk) override;
 
 	string GetReaderType() const override;
 
--- a/src/storage/ducklake_inlined_data_reader.cpp
+++ b/src/storage/ducklake_inlined_data_reader.cpp
@@ -155,7 +155,7 @@ bool DuckLakeInlinedDataReader::TryInitializeScan(ClientContext &context, Global
 }
 
 AsyncResult DuckLakeInlinedDataReader::Scan(ClientContext &context, GlobalTableFunctionState &global_state,
-                                     LocalTableFunctionState &local_state, DataChunk &chunk) {
+                                            LocalTableFunctionState &local_state, DataChunk &chunk) {
 	if (!virtual_columns.empty()) {
 		scan_chunk.Reset();
 		data->data->Scan(state, scan_chunk);


J-Meyers:
Current actions failure is unrelated, it's not on code that runs ducklake_add_data_files or even runs any sort of ducklake function:
```
1. test/sql/copy/csv/code_cov/csv_disk_reload.test:21
Query unexpectedly failed (test/sql/copy/csv/code_cov/csv_disk_reload.test:21)
 (test/sql/copy/csv/code_cov/csv_disk_reload.test:21)!
select count(*) from read_csv_auto('duckdb_unittest_tempdir/8626/t1.csv',buffer_size = 262144, sample_size=-1);
Actual result:
Out of Memory Error: could not allocate block of size 260.0 KiB (1.8 MiB/1.9 MiB used)

Possible solutions:
* Reducing the number of threads (SET threads=X)
* Disabling insertion-order preservation (SET preserve_insertion_order=false)
* Increasing the memory limit (SET memory_limit='...GB')

See also https://duckdb.org/docs/stable/guides/performance/how_to_tune_workloads

LINE 1: select count(*) from read_csv_auto('duckdb_unittest_tempdir/8626/t1.csv',buffer_...
```

J-Meyers:
Inconsistent failure of this test I've observed elsewhere as well, I've been unable to reproduce it locally.
```
1. test/sql/data_inlining/inlining_global_options.test:107
Wrong row count in query! (test/sql/data_inlining/inlining_global_options.test:107)!
Expected 4 rows, but got 3 rows
SELECT data_file_id, table_id FROM __ducklake_metadata_ducklake.ducklake_data_file;
Expected result:

10	2
11	5
12	4
13	6
Actual result:
data_file_id	table_id	
BIGINT	BIGINT	
[ Rows: 3]
10	2
11	5
12	6
```
https://github.com/duckdb/ducklake/actions/runs/19281301857/job/55132813801?pr=535

src/include/storage/ducklake_inlined_data_reader.hpp
```diff
@@ -30,7 +30,7 @@ class DuckLakeInlinedDataReader : public BaseFileReader {
 public:
 	bool TryInitializeScan(ClientContext &context, GlobalTableFunctionState &gstate,
 	                       LocalTableFunctionState &lstate) override;
-	void Scan(ClientContext &context, GlobalTableFunctionState &global_state, LocalTableFunctionState &local_state,
+	AsyncResult Scan(ClientContext &context, GlobalTableFunctionState &global_state, LocalTableFunctionState &local_state,
```
pdet: This come from the patch right?
Could you also PR to duckdb to remove the patch?

src/functions/ducklake_add_data_files.cpp
```diff
@@ -147,227 +148,323 @@ struct DuckLakeFileProcessor {
...
+
+		auto &parquet_metadata_entry = parquet_metadata_list_data[row_idx];
+		auto parquet_metadata_offset = parquet_metadata_entry.offset;
+		auto parquet_metadata_length = parquet_metadata_entry.length;
+
+		auto &parquet_schema_entry = parquet_schema_list_data[row_idx];
+		auto parquet_schema_offset = parquet_schema_entry.offset;
+		auto parquet_schema_length = parquet_schema_entry.length;
+
+		ParquetFileMetadata *file_ptr = nullptr;
```
pdet: Can this `file_ptr` ever stay null? I understand that you either skip it in the continue or you dereference it, but I guess it would always be a reference, no?

src/functions/ducklake_add_data_files.cpp
```diff
@@ -147,227 +148,323 @@ struct DuckLakeFileProcessor {
...
+			if (child_count > 0) {
+				column_stack.push_back(column_ptr);
+				child_counts.push_back(child_count);
 			}
 		}
 
-		column.column_stats.push_back(std::move(stats));
-	}
-}
+		DetermineMapping(file);
```
pdet, line 345: Out of curiosity, could the mapping also be achieved via the `parquet_full_metadata` function?

pdet (commented):
Hi @J-Meyers, sorry that this took a while, I was on a long-vacation.

Thanks for the contribution, I think this looks really good, just a couple minor comments.

src/functions/ducklake_add_data_files.cpp
J-Meyers: Not easily, since this is about the explicit relationships between the parquet columns and the `DuckLakeFieldId`s and does type checking to make sure that the type in the parquet file is consistent with the type already specified in ducklake

src/functions/ducklake_add_data_files.cpp
J-Meyers: I changed how it's initalized to be clearer, previously it was all wrapped in { } to make clear what variables may be used later, and file_ptr only had the purpose of leaking the file so that it could then be assigned to the reference, but that was still messy, now I've instead reduced the number of variables used in collecting the metadata and eliminated the {} eliminating the need for file_ptr

src/include/storage/ducklake_inlined_data_reader.hpp
J-Meyers: I don't think I can until this PR is in, when removing the patch I need to point to a different commit on main for ducklake that doesn't need the patch, and that wont exist until this PR is in

pdet: Yes, absolutely, I meant more as something to keep in mind!

src/functions/ducklake_add_data_files.cpp
pdet: Yeah, I think we have the `duckdb_type` in the `parquet_schema`, maybe something we could use instead of having that mapping code, it seems to be a bit of a pitfall of bugs right now.

Unrelated to this PR ofc.

pdet:
Thanks, I'll merge it as soon as all tests pass!

PR #541, opened 2025-11-02
Formatting nits

szarnyasg:


pdet:
Hey @szarnyasg thanks for the PR, LGTM.

Since you are changing the error messages, you also have to modify the tests that check them, otherwise, the tests will fail.

PR #555, opened 2025-11-11
Fix issue with row-lineage over merge into updates

pdet:
Fix: https://github.com/duckdblabs/liquid-analytics/issues/49
Fix: https://github.com/duckdblabs/liquid-analytics/issues/45


Mytherin:
LGTM - I think the encryption test needs a `require httpfs` after a recent upstream change

PR #556, opened 2025-11-12
Insertion Inline Support for Portgres/SQLite

pdet:
This PR adds support for insertion inlining for postgres and sqlite.
If the type is a nested type we store it as a string.

For non-nested types, the type stored will depend on what is available in the catalog. For example in SQLite:

```cpp
LogicalType SQLiteUtils::ToSQLiteType(const LogicalType &input) {
	switch (input.id()) {
	case LogicalTypeId::BOOLEAN:
	case LogicalTypeId::TINYINT:
	case LogicalTypeId::SMALLINT:
	case LogicalTypeId::INTEGER:
	case LogicalTypeId::BIGINT:
	case LogicalTypeId::UTINYINT:
	case LogicalTypeId::USMALLINT:
	case LogicalTypeId::UINTEGER:
		return LogicalType::BIGINT;
	case LogicalTypeId::FLOAT:
	case LogicalTypeId::DOUBLE:
		return LogicalType::DOUBLE;
	case LogicalTypeId::BLOB:
		return LogicalType::BLOB;
	default:
		return LogicalType::VARCHAR;
	}
```

Since the types stored might mismatch the actual data type, when reading from inlined data we always perform a cast. Note that if the cast is not necessary it will be removed by the optimizer.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1466,7 +1466,8 @@ string DuckLakeMetadataManager::GetColumnType(const DuckLakeColumnInfo &col) {
 		return StringUtil::Format("MAP(%s, %s)", GetColumnType(col.children[0]), GetColumnType(col.children[1]));
 	}
 	default:
-		if (col.children.empty()) {
+		if (!col.children.empty()) {
+			// This is a nested structure that we currently do not support.
 			throw NotImplementedException("Unsupported nested type %s in DuckLakeMetadataManager::GetColumnType",
 			                              col.type);
 		}


Alex-Monahan:
Super exciting!!! Thanks Pedro!!

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1443,6 +1443,11 @@ string DuckLakeMetadataManager::GetColumnType(const DuckLakeColumnInfo &col) {
 	if (col.children.empty()) {
 		return DuckLakeTypes::FromString(col.type).ToString();
 	}
+	auto &ducklake_catalog = transaction.GetCatalog();
+	if (!ducklake_catalog.IsDuckCatalog() && (col.type == "struct" || col.type == "map" || col.type == "list")) {
```
Mytherin: It would be nicer if this was a virtual method on the metadata manager, e.g.:

```
virtual bool TypeIsNativelySupported(const LogicalType &type) {
    ...
}
```

We could then create a `DuckDBMetadataManager` (similar to how we now have `PostgresMetadataManager`) that would override this method and just return true for every type.

Mytherin (commented):
Thanks! LGTM - one minor comment

yan-hic:
Somewhat related. Since both are file-based, why would one want/prefer SQLite over duckdb for ducklake catalog?

pdet:
You can use multiple clients and connect concurrently with sqlite as a catalog, while with duckdb you are limited to one client.

Sh4yy:
THIS IS AWESOME

guitcastro:
Duckdb 1.4.2 didn't ship this, version. Any changes that 1.4.3 will ship it?

Code42Cate:
Really looking forward to this, hope it gets shipped in 1.4.3 👀

pdet:
> Duckdb 1.4.2 didn't ship this, version. Any changes that 1.4.3 will ship it?

Yes, it will go to `1.4.3`.

djouallah:
do you need to turn it on or it is by default, as it is not working with 1.4.3

guillesd:
@djouallah  The docs need to be updated, but this is how you set it up: https://ducklake.select/docs/stable/duckdb/advanced_features/data_inlining
```sql
ATTACH 'ducklake:postgres:...' (DATA_INLINING_ROW_LIMIT 10);
```

djouallah:
was hoping for better performance to be honest :) every insert has one row

============================================================
SUMMARY STATISTICS
============================================================

HEAP Table:
  Average throughput: 486.83 inserts/sec
  Maximum throughput: 593.44 inserts/sec (at 8 workers)

ICEBERG Table:
  Average throughput: 1.75 inserts/sec
  Maximum throughput: 2.13 inserts/sec (at 2 workers)

DUCKLAKE Table:
  Average throughput: 10.51 inserts/sec
  Maximum throughput: 12.64 inserts/sec (at 4 workers)

Mytherin:
If you're using a remote Postgres table network round-trips on commits are probably dominating that time. The extension hasn't really optimized network round-trips at this point so there's a lot more round-trips happening than necessary.

djouallah:
@Mytherin  PG is in my laptop !! that's why I was surprised

Mytherin:
Hm, we should investigate this then. I'm guessing there's a lot of conflicts happening on insertion which is causing the slowdown. Could you perhaps re-run with a single worker and check performance then?

djouallah:
same issue even with 1 worker, I added a bug report here

https://github.com/duckdb/ducklake/issues/615

Code42Cate:
I had a similar issue with super slow queries, which I traced back to an excessive amount of schemas, which created many different inlining tables. Have you checked the amount of inlining tables/schemas?

PR #559, opened 2025-11-13
Ensure entry type before compaction

jeewonhh:
Fixes an issue when checkpointing DuckLake that has views. Discussed in https://github.com/duckdblabs/motherduck/issues/415


jeewonhh:
in favor of #558

PR #571, opened 2025-11-21
Expressions as Column Default Values

pdet:
This PR enables expressions to be used as column default values.
e.g.

``` sql
CREATE TABLE t_1 (id INTEGER, id_plus INTEGER DEFAULT 1);
ALTER TABLE t_1 ALTER id_plus SET DEFAULT round(pi());
```

should now be possible.

Additionally, I've removed the `"Only literals (e.g. 42 or 'hello world') are supported as default values"` error message skipping, so existing DuckDB tests related to default value usage are now executed as well.

### Specification Changes

The following columns are added to the specification:

``` sql
ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} default_value_qualifier VARCHAR DEFAULT NULL;
ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} default_value_dialect VARCHAR DEFAULT NULL;
```

They are used to track the qualifier (e.g., `literal`, `expression`) and the dialect (e.g., `duckdb`). We add these for both `initial_value` and `default_value`.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1524,13 +1524,10 @@ string GetExpressionType(ParsedExpression &expression) {
 		}
 		return "expression";
 	}
-	case ExpressionType::FUNCTION:
-	case ExpressionType::COLUMN_REF:
-		return "expression";
 	case ExpressionType::VALUE_CONSTANT:
 		return "literal";
 	default:
-		throw NotImplementedException("Expression type not implemented for default column value");
+		return "expression";
 	}
 }
 


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -70,7 +72,7 @@ CREATE TABLE {METADATA_CATALOG}.ducklake_column_tag(table_id BIGINT, column_id B
...
 CREATE TABLE {METADATA_CATALOG}.ducklake_delete_file(delete_file_id BIGINT PRIMARY KEY, table_id BIGINT, begin_snapshot BIGINT, end_snapshot BIGINT, data_file_id BIGINT, path VARCHAR, path_is_relative BOOLEAN, format VARCHAR, delete_count BIGINT, file_size_bytes BIGINT, footer_size BIGINT, encryption_key VARCHAR);
-CREATE TABLE {METADATA_CATALOG}.ducklake_column(column_id BIGINT, begin_snapshot BIGINT, end_snapshot BIGINT, table_id BIGINT, column_order BIGINT, column_name VARCHAR, column_type VARCHAR, initial_default VARCHAR, default_value VARCHAR, nulls_allowed BOOLEAN, parent_column BIGINT);
+CREATE TABLE {METADATA_CATALOG}.ducklake_column(column_id BIGINT, begin_snapshot BIGINT, end_snapshot BIGINT, table_id BIGINT, column_order BIGINT, column_name VARCHAR, column_type VARCHAR, initial_default VARCHAR, default_value VARCHAR, nulls_allowed BOOLEAN, parent_column BIGINT, initial_value_qualifier VARCHAR, initial_value_dialect VARCHAR, default_value_qualifier VARCHAR, default_value_dialect VARCHAR);
```
Mytherin: The `initial_value` should probably only be allowed to be a literal

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -161,6 +163,10 @@ void DuckLakeMetadataManager::MigrateV03(bool allow_failures) {
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro(schema_id BIGINT, macro_id BIGINT, macro_name VARCHAR, begin_snapshot BIGINT, end_snapshot BIGINT);
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro_impl(macro_id BIGINT, impl_id BIGINT, dialect VARCHAR, sql VARCHAR, type VARCHAR);
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro_parameters(macro_id BIGINT, impl_id BIGINT,column_id BIGINT, parameter_name VARCHAR, parameter_type VARCHAR, default_value VARCHAR, default_value_type VARCHAR);
+ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} initial_value_qualifier VARCHAR DEFAULT NULL;
+ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} initial_value_dialect VARCHAR DEFAULT NULL;
+ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} default_value_qualifier VARCHAR DEFAULT NULL;
```
Mytherin: Can this be `default_value_type`

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1503,21 +1517,55 @@ void DuckLakeMetadataManager::WriteNewSchemas(DuckLakeSnapshot commit_snapshot,
 	}
 }
 
+string GetExpressionType(ParsedExpression &expression) {
+	switch (expression.type) {
+	case ExpressionType::OPERATOR_CAST: {
+		auto &cast_expression = expression.Cast<CastExpression>();
+		if (cast_expression.child->type == ExpressionType::VALUE_CONSTANT) {
+			return "literal";
```
Mytherin, line 1523: Can we add a test where we verify that if we have a literal that contains e.g. function call-ish syntax (`DEFAULT VALUE 'random()'`) this still works

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -69,10 +70,38 @@ struct DuckLakeColumnInfo {
 	string name;
 	string type;
 	Value initial_default;
-	Value default_value;
+	unique_ptr<ParsedExpression> default_value;
```
Mytherin: This shouldn't be an expression.

The idea behind the structs in `ducklake_metadata_info` is that they represent what actually goes into and comes out of the metadata store - i.e. they should represent the specification. We cannot store a `unique_ptr<ParsedExpression>` in a metadata store. This should just stay as a `Value` - and we should add new fields (e.g. `string default_value_type`, `string default_value_dialect`) to correspond to the new fields we've added to the specification.

The actual transformation between string -> Expression should happen at a different layer - not in the metadata manager.

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -73,6 +74,30 @@ struct DuckLakeColumnInfo {
 	bool nulls_allowed {};
 	vector<DuckLakeColumnInfo> children;
 	vector<DuckLakeTag> tags;
+	DuckLakeColumnInfo() {
+	}
+
+	DuckLakeColumnInfo(const DuckLakeColumnInfo &other)
```
Mytherin: I think these explicit copy constructors can all go now

src/storage/ducklake_catalog.cpp
```diff
@@ -196,7 +196,13 @@ static unique_ptr<DuckLakeFieldId> TransformColumnType(DuckLakeColumnInfo &col)
 	if (col.children.empty()) {
 		auto col_type = DuckLakeTypes::FromString(col.type);
 		col_data.initial_default = col.initial_default.DefaultCastAs(col_type);
-		col_data.default_value = col.default_value.DefaultCastAs(col_type);
+
+		auto sql_expr = Parser::ParseExpressionList(col.default_value.GetValue<string>());
```
Mytherin: We should be checking the `default_value_type` before assuming it is an expression. What if the default value is the string literal `'random()'`? 

I think this works now because, in this version of the code, we always write expressions, so a default value of `random()` is written as `'random()'` which we can parse as an expression. But if we were to read an old version of the database that would not have expressions and that value would be stored as `random()` - incorrectly leading us to parse it as a function call.

src/storage/ducklake_field_data.cpp
```diff
@@ -40,14 +40,31 @@ DuckLakeFieldId::DuckLakeFieldId(DuckLakeColumnData column_data_p, string name_p
...
+                                                             const LogicalType &type) {
 	if (!default_expr) {
+		return make_uniq<ConstantExpression>(Value(type));
+	}
+	switch (default_expr->type) {
+	case ExpressionType::VALUE_CONSTANT:
+	case ExpressionType::FUNCTION:
+	case ExpressionType::COLUMN_REF:
+	case ExpressionType::OPERATOR_CAST:
+		return default_expr->Copy();
```
Mytherin: Maybe we can instead check for what types of operators are not allowed instead of checking for these types explicitly, e.g. we can check if the expression has a subquery or window function and throw an error in that case

src/storage/ducklake_field_data.cpp
```diff
@@ -40,14 +40,31 @@ DuckLakeFieldId::DuckLakeFieldId(DuckLakeColumnData column_data_p, string name_p
...
+	default:
+		throw NotImplementedException(
+		    "Only literals (e.g. 42 or 'hello world') and expressions (e.g., 'NOW()') are supported as default values");
```
Mytherin: Can we update this error message?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -161,6 +163,8 @@ void DuckLakeMetadataManager::MigrateV03(bool allow_failures) {
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro(schema_id BIGINT, macro_id BIGINT, macro_name VARCHAR, begin_snapshot BIGINT, end_snapshot BIGINT);
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro_impl(macro_id BIGINT, impl_id BIGINT, dialect VARCHAR, sql VARCHAR, type VARCHAR);
 CREATE TABLE {IF_NOT_EXISTS} {METADATA_CATALOG}.ducklake_macro_parameters(macro_id BIGINT, impl_id BIGINT,column_id BIGINT, parameter_name VARCHAR, parameter_type VARCHAR, default_value VARCHAR, default_value_type VARCHAR);
+ALTER TABLE {METADATA_CATALOG}.ducklake_column ADD COLUMN {IF_NOT_EXISTS} default_value_type VARCHAR DEFAULT NULL;
```
Mytherin: Maybe this should be `DEFAULT 'literal'`?

src/storage/ducklake_table_entry.cpp
```diff
@@ -1100,7 +1106,12 @@ DuckLakeColumnInfo DuckLakeTableEntry::GetColumnInfo(FieldIndex field_index) con
 	result.name = col.Name();
 	result.type = DuckLakeTypes::ToString(col.Type());
 	result.initial_default = col_data.initial_default;
-	result.default_value = col_data.default_value;
+
+	if (col_data.default_value) {
+		result.default_value = col_data.default_value->ToString();
```
Mytherin: I think if we have a string literal `my_default_value` - this might now get written to the system as `'my_default_value'` instead of as `my_default_value`. We should check if the root element is a `VALUE_CONSTANT` and then extract that constant directly to avoid this.

test/sql/default/default_expressions.test
```diff
@@ -0,0 +1,97 @@
...
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/default_expressions', METADATA_CATALOG 'xx')
+
+statement ok
+use ducklake
+
+statement error
+create table t (id integer, created_at timestamp default now());
+----
+Only literals (e.g. 42 or 'hello world') are supported as initial values
```
Mytherin: This should work. When we're creating an empty table, the initial default value is useless anyway since there are no existing files. We can just always set the initial default value to `NULL` in this scenario.

test/sql/default/default_expressions.test
```diff
@@ -0,0 +1,97 @@
...
+2	2	1	literal	duckdb
+2	2	round(pi())	expression	duckdb
+3	1	NULL	literal	duckdb
+3	2	NULL	literal	duckdb
+3	2	'random()'	literal	duckdb
+
+statement error
+ALTER TABLE t_2 ADD COLUMN j INTEGER DEFAULT RANDOM()
+----
+Only literals (e.g. 42 or 'hello world') are supported as initial values
```
Mytherin: Maybe this error message can be more descriptive for the work-around - e.g. "we cannot add a column with a non-literal default value - add the column and then explicitly set the default for new values using `ALTER ... SET DEFAULT`"

Mytherin (commented):
Thanks for the changes! Looks good - some more comments

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1503,21 +1515,66 @@ void DuckLakeMetadataManager::WriteNewSchemas(DuckLakeSnapshot commit_snapshot,
...
+		auto &cast_expression = expression.Cast<CastExpression>();
+		if (cast_expression.child->type == ExpressionType::VALUE_CONSTANT) {
+			return "literal";
+		}
+		return "expression";
+	}
+	case ExpressionType::FUNCTION:
+	case ExpressionType::COLUMN_REF:
+		return "expression";
+	case ExpressionType::VALUE_CONSTANT:
```
Mytherin, line 1527: Can this be `default: return "expression"`? This list seems unnecessarily restrictive - e.g. currently we can't have a default `CASE` expression or comparison in the root

Mytherin (commented):
Looks good - one minor comment then this is good to go

PR #581, opened 2025-11-24
add support for type promoting timestamp to timestamptz

guillesd:
As @pdet mentioned we should have a more permanent fix that uses `DUCKDB_API static LogicalType MaxLogicalType` instead of having our own mapping in DuckLake.

### DIFF
--- a/test/sql/alter/alter_timestamptz_promotion.test
+++ b/test/sql/alter/alter_timestamptz_promotion.test
@@ -19,3 +19,6 @@ CREATE TABLE ducklake.test(col1 TIMESTAMP);
 
 statement ok
 ALTER TABLE ducklake.test ALTER COLUMN col1 TYPE TIMESTAMPTZ;
+
+statement ok
+INSERT INTO ducklake.test VALUES ('2025-01-15 12:30:45'::TIMESTAMPTZ)


test/sql/alter/alter_timestamptz_promotion.test
```diff
@@ -0,0 +1,24 @@
...
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_add_col_files')
```
pdet, line 15: Can you change the data_path to '${DATA_PATH}/alter_timestamptz_promotion'

test/sql/alter/alter_timestamptz_promotion.test
```diff
@@ -0,0 +1,24 @@
...
+
+statement ok
+CREATE TABLE ducklake.test(col1 TIMESTAMP);
+
```
pdet, line 19: Can you add a timestamp insert here?

test/sql/alter/alter_timestamptz_promotion.test
```diff
@@ -0,0 +1,24 @@
...
+statement ok
+ALTER TABLE ducklake.test ALTER COLUMN col1 TYPE TIMESTAMPTZ;
+
+statement ok
+INSERT INTO ducklake.test VALUES ('2025-01-15 12:30:45'::TIMESTAMPTZ)
```
pdet, line 24: Can you verify the values after this insertion?

PR #582, opened 2025-11-24
Timestamptz type promotion to andium

guillesd:
As mentioned we should have a more permanent fix that uses DUCKDB_API static LogicalType MaxLogicalType instead of having our own mapping in DuckLake.

### DIFF
--- a/test/sql/alter/alter_timestamptz_promotion.test
+++ b/test/sql/alter/alter_timestamptz_promotion.test
@@ -12,13 +12,22 @@ test-env DATA_PATH __TEST_DIR__
 
 
 statement ok
-ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_add_col_files')
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/alter_timestamptz_promotion')
 
 statement ok
 CREATE TABLE ducklake.test(col1 TIMESTAMP);
+
+statement ok
+INSERT INTO ducklake.test VALUES ('2025-01-15 12:30:45'::TIMESTAMP)
 
 statement ok
 ALTER TABLE ducklake.test ALTER COLUMN col1 TYPE TIMESTAMPTZ;
 
 statement ok
 INSERT INTO ducklake.test VALUES ('2025-01-15 12:30:45'::TIMESTAMPTZ)
+
+query I
+FROM ducklake.test
+----
+2025-01-15 12:30:45+00
+2025-01-15 12:30:45+00


pdet:
Hi @guillesd could you adjust this PR with the review I left in the previous PR? https://github.com/duckdb/ducklake/pull/581

guillesd:
@pdet  Sorry I forgot to push the changes!

pdet:
Thank Gui, Looks great!

onnimonni:
Thanks for quick fix 😊!

Since this extension is core extension does it mean that when next release `1.4.3` is done these changes will be live?

Is there a way for me to test this already now locally?

guillesd:
Yes, we will push an updated DuckLake release with 1.4.3. 

You can always test this by either building the 1.4-andium branch or grabing one of the release binaries from the artifacts of the github actions build.

utay:
Hi @guillesd @pdet, it looks like this promotion might not be working as expected. The `ALTER TABLE` only updates the schema metadata but doesn't rewrite the Parquet files. When they’re read in a session with a non-UTC timezone, the values appear to be interpreted as local time, which shifts the actual instant and leads to incorrect results.

Here's a test that demonstrates it:
```
# name: test/sql/alter/alter_timestamptz_promotion_timezone.test
# description: Demonstrates that TIMESTAMP->TIMESTAMPTZ promotion doesn't behave the same as native TIMESTAMPTZ
# group: [alter]

require ducklake

require parquet

require icu

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/alter_timestamptz_promotion_tz')

# Create a table with TIMESTAMPTZ from the start (the "correct" reference)
statement ok
CREATE TABLE ducklake.native_tstz(t TIMESTAMPTZ);

# Create a table with TIMESTAMP that we will migrate to TIMESTAMPTZ
statement ok
CREATE TABLE ducklake.migrated_tstz(t TIMESTAMP);

# Insert timestamps in UTC
statement ok
SET timezone='UTC';

statement ok
INSERT INTO ducklake.native_tstz VALUES ('2026-01-25 12:00:00');

statement ok
INSERT INTO ducklake.migrated_tstz VALUES ('2026-01-25 12:00:00');

# Migrate the TIMESTAMP table to TIMESTAMPTZ
statement ok
ALTER TABLE ducklake.migrated_tstz ALTER COLUMN t TYPE TIMESTAMPTZ;

# ============================================================================
# In UTC: Both tables look identical
# ============================================================================
statement ok
SET timezone='UTC';

query I
SELECT t FROM ducklake.native_tstz;
----
2026-01-25 12:00:00+00

query I
SELECT t FROM ducklake.migrated_tstz;
----
2026-01-25 12:00:00+00

# Same epoch in UTC
query I
SELECT epoch(t) FROM ducklake.native_tstz;
----
1769342400

query I
SELECT epoch(t) FROM ducklake.migrated_tstz;
----
1769342400

# ============================================================================
# In America/Denver: The bug manifests
# 
# For correct TIMESTAMPTZ behavior:
# - The epoch (underlying instant) should NEVER change based on timezone
# - Only the display should change (12:00 UTC = 05:00 Denver)
# ============================================================================
statement ok
SET timezone='America/Denver';

# Native TIMESTAMPTZ: displays correctly as 05:00 Denver, epoch unchanged
query I
SELECT t FROM ducklake.native_tstz;
----
2026-01-25 05:00:00-07

query I
SELECT epoch(t) FROM ducklake.native_tstz;
----
1769342400

# BUG: Migrated table displays as 12:00 Denver (wrong!) with epoch shifted +7 hours
query I
SELECT t FROM ducklake.migrated_tstz;
----
2026-01-25 12:00:00-07

# BUG: The epoch changed! It's now 7 hours (25200 seconds) later.
# This means the same data represents a DIFFERENT instant depending on timezone.
query I
SELECT epoch(t) FROM ducklake.migrated_tstz;
----
1769367600
```

PR #592, opened 2025-12-03
Make merge_adjacent_files incremental by adding max_compacted_files option

pdet:
e.g.,
```sql
CALL ducklake_merge_adjacent_files('ducklake', 'example', max_compacted_files=>10);
``` 
Would perform a `merge_adjacent_files` compaction in producing a max of 10 files on table `example`.

Note that the number of actual compacted files is highly dependent on the target size of the final file.

### DIFF
--- a/.github/workflows/Catalogs.yml
+++ b/.github/workflows/Catalogs.yml
@@ -26,6 +26,7 @@ jobs:
       VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
       ENABLE_SQLITE_SCANNER: ON
       ENABLE_POSTGRES_SCANNER: ON
+      BUILD_EXTENSION_TEST_DEPS: full
 
     steps:
     - name: Install required ubuntu packages
--- a/src/functions/ducklake_cleanup_files.cpp
+++ b/src/functions/ducklake_cleanup_files.cpp
@@ -65,9 +65,10 @@ static unique_ptr<FunctionData> CleanupBind(ClientContext &context, TableFunctio
 	bool cleanup_all = false;
 	for (auto &entry : input.named_parameters) {
 		if (StringUtil::CIEquals(entry.first, "dry_run")) {
-			result->dry_run = true;
+			result->dry_run = entry.second.GetValue<bool>();
+			;
 		} else if (StringUtil::CIEquals(entry.first, "cleanup_all")) {
-			cleanup_all = true;
+			cleanup_all = entry.second.GetValue<bool>();
 		} else if (StringUtil::CIEquals(entry.first, "older_than")) {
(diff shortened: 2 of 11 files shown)


src/functions/ducklake_compaction_functions.cpp
```diff
@@ -281,6 +283,10 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 				}
 				// this file can be compacted along with the neighbors
 				current_file_size += file_size;
+				compacted_files++;
+				if (compacted_files > max_files) {
+					break;
```
Mytherin: This will abort the gathering of files for an individual compaction, leading to potentially smaller than desired files. For example, let's say we have a ton of files of 10KB each, calling `merge_adjacent_files()` might result in the following compaction:


```
10KB
10KB
10KB
10KB
10KB
10KB
10KB
10KB
10KB
...
10KB

-> 512MB
```

Now calling `merge_adjacent_files` with `max_count = 10` will instead do this:

```
10KB
10KB
10KB
10KB
...
10KB

-> 100KB
```

If we call that repeatedly, we get a ton of 100KB files. Those will then get compacted to 1MB files if we keep calling the function. Those will then get compacted to 10MB files, etc. 

That seems undesirable. I think the file count we want to count here is the *target* file count, not the *input* file count.

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -154,14 +154,15 @@ class DuckLakeCompactor {
 	Binder &binder;
 	TableIndex table_id;
 	double delete_threshold = 0.95;
+	uint64_t max_files = 0;
```
Mytherin: This should be an `optional_idx`

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -591,6 +614,7 @@ TableFunctionSet DuckLakeMergeAdjacentFilesFunction::GetFunctions() {
 		function.bind_operator = MergeAdjacentFilesBind;
 		if (type.size() == 2) {
 			function.named_parameters["schema"] = LogicalType::VARCHAR;
+			function.named_parameters["max_compacted_files"] = LogicalType::UBIGINT;
```
Mytherin, line 617: We should probably add a "table_name" qualifier here as well

src/functions/ducklake_compaction_functions.cpp
pdet: that's fair!

src/functions/ducklake_compaction_functions.cpp
pdet: I'm confused, isn't the table_name part of the non-named parameters already?

Mytherin: Ah you're right - just didn't see it here and misremembered it being there.

PR #593, opened 2025-12-03
Ordered compaction and inline flush (local_order_by option)

Alex-Monahan:
Hi folks! 

I am super open to feedback - this is the most complex C++ I have ever attempted. It was not vibe coded - I am trying to learn things myself!

This adds a new option  `local_order_by` that will sort data when compacting files or when flushing inlined data. It is a string that contains an order by expression. A local_order_by parameter is also added to `ducklake_merge_adjacent_files` and `ducklake_flush_inlined_data`. This option does not affect data as it is inserted - a user can specify their own order by clause manually. 

The benefits of this option are to allow for more efficient data skipping (including better Bloom filter utilization) when reading data for faster reads, at the expense of slower compaction/inline flushes. It was discussed here: https://github.com/duckdb/ducklake/discussions/329 . This is commonly described as the 2nd most important optimization to consider when using Iceberg (aside from defining effective partitions). It is most useful when a column has a high enough cardinality that explicit partitions are not feasible/optimal. A micro-benchmark is described below.

Today, this is restricted to only explicit column refs (Ex: `column_a asc, column_b desc`). This is because I couldn't figure out how to bind arbitrary expressions... I could use some help there! However, there are some benefits / other alternatives to this restriction:
* This simplifies cross engine compatibility - if expressions were allowed, they may not be compatible across engines
* An alternative could be a column with a default expression that is then sorted on. This is not identical, but could be an alternative.

The local_order_by is not a global ordering - it only applies to files already selected to be compacted together into a single file. This avoids very resource intensive full-table sorts (and aligns with Iceberg). This could be improved in the future by approximately ordering the file names that are selected for compaction according to the min/max statistics of the columns in the `local_order_by` option. (However, this would be more difficult if full expressions were allowed. I am open to feedback on which would be preferred!)

### Microbenchmark

For a microbenchmark, I used the dataset described in my blog here: https://duckdb.org/2025/06/06/advanced-sorting-for-fast-selective-queries#the-on-time-flights-dataset 

I created an explicit column with the result of the `zipped_varchar` macro as described in [this section of the blog](https://duckdb.org/2025/06/06/advanced-sorting-for-fast-selective-queries#experimental-design) and then ran 4 tests. All tests queried filtered data on both the origin and dest column. All had data inserted and then compacted.
* Worst Case: Single insertion into DuckLake in random order
* Optimal: Single insertion into DuckLake in zipped_varchar(origin, dest) order
* Regresses to worst case: Weekly insertion into DuckLake (ordered by zipped_varchar(origin, dest), but effectively ordered by week)
* Near Optimal: Weekly insertion into DuckLake with new local_order_by compaction enabled

Results showed that the compacted data performed just as well as the optimally ordered single-insert data (compared with much slower performance when data is inserted frequently and compacted in that order):
|           type            |         table          |      min_time      |     time_50th      |      max_time      |
|---------------------------|------------------------|-------------------:|-------------------:|-------------------:|
| select_origin_destination | flights_ordered        | 1.6 | 1.8 | 2.1  |
| select_origin_destination | flights_random         | 16.5 | 17.1  | 21.1  |
| select_origin_destination | flights_weekly         | 3.8 | 15.2  | 16.1  |
| select_origin_destination | THIS PR: flights_weekly_ordered | 1.6 | 1.8 | 2.0 |

### Other Notes

To reuse across inlining and compaction, I moved some compaction classes into a .hpp file. Let me know if that is ok!

I'm also more than happy to do the docs PR for this as well. 

Tagging @philippmd since he has looked at this with me a bit!

### DIFF
--- a/.github/workflows/Catalogs.yml
+++ b/.github/workflows/Catalogs.yml
@@ -26,6 +26,7 @@ jobs:
       VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
       ENABLE_SQLITE_SCANNER: ON
       ENABLE_POSTGRES_SCANNER: ON
+      BUILD_EXTENSION_TEST_DEPS: full
 
     steps:
     - name: Install required ubuntu packages
@@ -73,11 +74,11 @@ jobs:
 
     - name: Test SQLite
       run: |
-        build/release/test/unittest --test-config test/configs/sqlite.json
+        build/release/test/unittest --test-config test/configs/sqlite.json --test-dir ./ "test/sql/*"
 
     - name: Test Postgres
       env:
         PGSERVICE: postgres
       run: |
         createdb ducklakedb
-        build/release/test/unittest --test-config test/configs/postgres.json
+        build/release/test/unittest --test-config test/configs/postgres.json --test-dir ./ "test/sql/*"
--- a/.github/workflows/ConfigTests.yml
+++ b/.github/workflows/ConfigTests.yml
@@ -57,10 +57,9 @@ jobs:
     - name: Build extension
       env:
         GEN: ninja
-        STATIC_LIBCPP: 1
       run: |
-        make reldebug
+        make release
 
     - name: Test extension
       run: |
-        build/reldebug/test/unittest --test-config test/configs/attach_ducklake.json --test-dir duckdb
+        build/release/test/unittest --test-config test/configs/attach_ducklake.json --test-dir duckdb
--- a/.github/workflows/Debug.yml
+++ b/.github/workflows/Debug.yml
@@ -1,0 +1,73 @@
+name: Debug Mode Tests
+on: [push, pull_request,repository_dispatch]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref || '' }}-${{ github.base_ref || '' }}-${{ github.ref != 'refs/heads/main' || github.sha }}
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ducklake-config-tests:
+    name: DuckLake (Debug)
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Add commits/tags to build against other DuckDB versions
+        duckdb_version: [ '<submodule_version>' ]
+        arch: ['linux_amd64']
+        vcpkg_version: [ '2023.04.15' ]
+        include:
+          - arch: 'linux_amd64'
+            vcpkg_triplet: 'x64-linux'
+
+    env:
+      VCPKG_TARGET_TRIPLET: ${{ matrix.vcpkg_triplet }}
+      GEN: Ninja
+      VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
+      DISABLE_EXTENSIONS_FOR_TEST: ON
+
+    steps:
+    - name: "node-cleanup"
+      run: |
+        sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
+        sudo docker image prune --all --force
+        sudo docker builder prune -a
+
+    - name: Install required ubuntu packages
+      run: |
+        sudo apt-get update -y -qq
+        sudo apt-get install -y -qq build-essential cmake ninja-build ccache python3
+
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: 'true'
+
+    - name: Checkout DuckDB to version
+      if: ${{ matrix.duckdb_version != '<submodule_version>'}}
+      run: |
+        cd duckdb
+        git checkout ${{ matrix.duckdb_version }}
+
+    - name: Setup Ccache
+      uses: hendrikmuhs/ccache-action@main
+      with:
+        key: ${{ github.job }}
+        save: ${{ github.ref == 'refs/heads/main' || github.repository != 'duckdb/duckdb-postgres' }}
+
+    - name: Setup vcpkg
+      uses: lukka/run-vcpkg@v11.1
+      with:
+        vcpkgGitCommitId: a42af01b72c28a8e1d7b48107b33e4f286a55ef6
+
+    - name: Build extension
+      env:
+        GEN: ninja
+        CMAKE_BUILD_PARALLEL_LEVEL: 2
+      run: |
+        make debug
+
+    - name: Test extension
+      run: |
+        build/debug/test/unittest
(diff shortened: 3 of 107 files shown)


Alex-Monahan:
Let me know if I should dig into the failing tests! I see that other PRs are failing some tests as well.

pdet:
Hi @Alex-Monahan first of all, many thanks for the PR! -> "It was not vibe coded" = Me happy! :-)

The CI failure is for sure unrelated to your PR, it seems to be related to `apk update` not taking confirmation flags, although it clearly exists in `extension-ci-tools/docker/linux_amd64_musl/Dockerfile`, maybe @carlopi has an answer to it?

 I'll dig in tomorrow, and will drop a review at some point this week!

guillesd:
Hi and thanks @Alex-Monahan! I'm tagging @Tmonster since he's taking a look at Iceberg sorts, to understand whether this is similar/compatible to that!

carlopi:
apk issue should be fixed via https://github.com/duckdb/extension-ci-tools/pull/288 (but maybe needs to be propagated to the other branch first, @samansmink ?)

pdet:
Hi @Alex-Monahan, 

Expressions as strings in configurations can be a bit brittle. We think the best approach is to use `SET SORTED BY` (https://github.com/duckdb/duckdb/pull/16714), which was added to the grammar a few months back.

The way to proceed is to properly support this at the file level. It might be more challenging to make it work for partitioned tables, but we could start with an initial PR that supports it only for non-partitioned tables plus compaction, while throwing a not-implemented exception for the partitioned case.

It should work similarly to supporting partition keys, with the difference that should be able to specify `DESC/ASC` and `NULLS FIRST/NULLS LAST`.

Alex-Monahan:
I will have a look! This would still be a local sort, correct? Would it apply to inserts, or still just to compactions and inline flushes?

Alex-Monahan:
I'm continuing to try and think through things! If Discord or a zoom call would be better, just let me know!

Where would you want the order by configuration to live, and where would you like the file-level status indicator to live?

#### Table level configuration
1. In the catalog as a varchar column on the table table
2. In its own table, order_by, one row per expression
3. In its own table, order_by, one row per column and a FK to the column table (This would prevent multi-column expressions, like Geospatial space filling curve lat/long expressions)
4. As a column (or columns) in the column table
5. As an option (like in the PR), but set with improved syntax

#### File level status
1. In the data file table as a varchar column
2. In data file column stats extra stats

Or other things I didn't think of! Open to your thoughts!

Also, is there a plan for using that constraint in the optimizer? I am not sure how things could be guaranteed.

pdet:
I think it can apply to inserts as well, compactions (`delete_rewrite` + `merge_adjacent`), and the flush inlined. So basically at any point files are being written.

What I meant about being at the file level it that a global sort would be prohibitively expensive, at the file level should be enough for us to skip row groups at least.

I guess we would add a new table to the spec, as follows:

```
sort_key
----------------------------------------------------
sort_id BIGINT
table_id  BIGINT
sort_key_index  BIGINT    -- Order to be sorted
expression   VARCHAR  
dialect VARCHAR  -- e.g., DuckDB
sort_direction VARCHAR   -- `ASC` or `DESC`
null_order VARCHAR   -- `NULLS_FIRST` or `NULLS_LAST`
```

I'm just not sure if they should be versioned, because the way I see it the sorting is not a guarantee, but rather something we can use to skip partitions later on. e.g., if you add an existing file to DuckLake, we shouldn't rewrite. On the other hand it seems that iceberg does [order versioning](https://iceberg.apache.org/docs/1.4.2/evolution/?utm_source=chatgpt.com#sort-order-evolution)?
```
begin_snapshot  BIGINT
end_snapshot  BIGINT
```

pdet:
I think it can apply to inserts as well, compactions (`delete_rewrite` + `merge_adjacent`), and the flush inlined. So basically at any point files are being written.

What I meant about being at the file level it that a global sort would be prohibitively expensive, at the file level should be enough for us to skip row groups at least.

I guess we would add a new table to the spec, as follows:

```
sort_key
----------------------------------------------------
sort_id BIGINT
table_id  BIGINT
sort_key_index  BIGINT    -- Order to be sorted
expression   VARCHAR  
dialect VARCHAR  -- e.g., DuckDB
sort_direction VARCHAR   -- `ASC` or `DESC`
null_order VARCHAR   -- `NULLS_FIRST` or `NULLS_LAST`
```

I'm just not sure if they should be versioned, because the way I see it the sorting is not a guarantee, but rather something we can use to skip partitions later on. e.g., if you add an existing file to DuckLake, we shouldn't rewrite. On the other hand it seems that iceberg does [order versioning](https://iceberg.apache.org/docs/1.4.2/evolution/?utm_source=chatgpt.com#sort-order-evolution)?
```
begin_snapshot  BIGINT
end_snapshot  BIGINT
```

Alex-Monahan:
Ah, those details help a ton!

I agree that a global sort would not be feasible - makes sense to me!

My only hesitation is having inserts also be sorted. I can envision a use case for "ingest data as fast as possible, then make it fast to query in the background without slowing down my ingest". My instinct would be to not have inserts be sorted (since someone can add their own order by in their insert query already), but maybe it would be good to have some kind of option to enable/disable "sort_on_insert" or "sort_in_background_only"? 

One example of that fast ingest use case is when someone sets `per_thread_output=true`. In that case, I would still want to be able to sort while I compact, but not sort when I insert.

pdet:
> Ah, those details help a ton!
> 
> I agree that a global sort would not be feasible - makes sense to me!
> 
> My only hesitation is having inserts also be sorted. I can envision a use case for "ingest data as fast as possible, then make it fast to query in the background without slowing down my ingest". My instinct would be to not have inserts be sorted (since someone can add their own order by in their insert query already), but maybe it would be good to have some kind of option to enable/disable "sort_on_insert" or "sort_in_background_only"?
> 
> One example of that fast ingest use case is when someone sets `per_thread_output=true`. In that case, I would still want to be able to sort while I compact, but not sort when I insert.

I think adding a global/schema/table option is probably a good idea.

Tmonster:
Iceberg has a schema option for [sorting tables](https://iceberg.apache.org/spec/#sorting) where users can declare how inserts should be sorted, potentially using a transform if needed. I would also prefer something like this since it more closely aligns with Iceberg. So during table creation you could have something like
```
CREATE TABLE ducklake.test(i INTEGER, b VARCHAR, c BLOB) SORTED BY i DESC NULLS LAST;
-- using a transform
CREATE TABLE ducklake.test(i INTEGER, b VARCHAR, c BLOB) SORTED BY truncate(5, b) DESC NULLS LAST;
```

In Iceberg, the default sort order does not **have** to be followed if the order is prohibitively expensive. This is for streaming writes. This is something we could think about as well since we have inlining, potentially in ducklake sorting streaming data will be easier to handle.

Alex-Monahan:
Very cool that Iceberg also does not always sort on insert! The 2 cases I've thought about so far where I wouldn't want to sort on insert are when inlining or when `per_thread_output` is true. Maybe we don't sort on insert in those cases, as well as have an option for `sort_on_insert` that could override?

I think I can take a crack at this approach, with the exception of the expression side of things (I am stumped there!). Could I put up a PR with the temporary limitation of only ordering by columns? Then a later PR could relax that restriction?

PR #601, opened 2025-12-05
[BUG] Add partition information when doing ducklake_add_data_files

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/6792

### DIFF
--- a/src/functions/ducklake_cleanup_files.cpp
+++ b/src/functions/ducklake_cleanup_files.cpp
@@ -11,16 +11,6 @@
 
 namespace duckdb {
 
-namespace {
-// Conform timestamp to ISO-8601 extended format with optional fractional seconds and timezone offset, e.g.:
-// "2025-12-26T06:13:30.673176+00:00" (UTC) or "2025-12-26T01:13:30.673176-05:00" (EST)
-string FormatTimestampISO8601(timestamp_t timestamp) {
-	auto ts_string = Timestamp::ToString(timestamp);
-	std::replace(ts_string.begin(), ts_string.end(), ' ', 'T');
-	return ts_string + "+00";
-}
-} // namespace
-
 struct CleanupBindData : public TableFunctionData {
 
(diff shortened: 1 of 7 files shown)


src/functions/ducklake_add_data_files.cpp
```diff
@@ -907,6 +907,20 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 
 	// we successfully mapped this file - register the name map and refer to it in the file
 	result.mapping_id = transaction.AddNameMap(std::move(name_map));
+
+	const auto partition_data = table.GetPartitionData().get();
+	if (partition_data) {
+		D_ASSERT(file.hive_partition_values.size() == partition_data->fields.size());
```
Mytherin: I don't think an assertion here is enough - users can add any files with any partitioning scheme. This should probably be checked and tested more thoroughly.

Mytherin (commented):
Thanks for the PR! Looks good, one comment

test/sql/add_files/add_file_partitioned.test
```diff
@@ -0,0 +1,116 @@
...
+statement ok
+COPY (SELECT 4 AS id, 2 as partition_column, 2 as other_partition_column) TO '${DATA_PATH}/add_file_partitioned/main/other_partitioned_t/other_partition_column=2/file.parquet' (FORMAT parquet);
+
+statement error
+CALL ducklake_add_data_files('ducklake', 'partitioned_t', '${DATA_PATH}/add_file_partitioned/main/other_partitioned_t/other_partition_column=2/file.parquet')
+----
+contains an invalid partition value for the table configuration
+
+statement ok
+ALTER TABLE partitioned_t SET PARTITIONED BY (other_partition_column);
```
Mytherin, line 107: Can we add a test with a partitioning over multiple columns (e.g. `year, month, day`)?

And then add files that:

* Match the partition
* Match a subset of the partition (`year, month`)
* Match the partition, but in a different order (`day, year, month`)

Issue #597, opened 2025-12-04
DELETE inside a transaction removes rows that should not be deleted when the table was populated via multi-row INSERT


hand-burger:
### What happens?

A `DELETE` statement removes rows that do not match the predicate when the table was populated using a multi-row INSERT inside a single transaction.
This bug does not occur in base DuckDB and does not occur if the inserts are split into separate statements.

### To Reproduce

```SQL
LOAD ducklake;
ATTACH 'ducklake:my_ducklake.ducklake' AS my_ducklake;
USE my_ducklake;

CREATE TABLE t1(c1 int);

BEGIN;

INSERT INTO t1 VALUES (1), (2);

SELECT * FROM t1;

UPDATE t1 SET c1 = 3 WHERE c1 = 1;

SELECT * FROM t1;

DELETE FROM t1 WHERE c1 = 2;

SELECT * FROM t1;
```

### Output
```SQL
┌────┐
│ c1 │
╞════╡
│  1 │
│  2 │
└────┘

┌────┐
│ c1 │
╞════╡
│  2 │
│  3 │
└────┘

┌────┐
│ c1 │
╞════╡
│0 rows│
└────┘
```

### Expected Output
```SQL
┌────┐
│ c1 │
╞════╡
│  1 │
│  2 │
└────┘

┌────┐
│ c1 │
╞════╡
│  2 │
│  3 │
└────┘

┌────┐
│ c1 │
╞════╡
│  3 │
└────┘
```

If instead the `INSERT` is split into two separate inserts, it gives correct output.
This issue is also reproducible with `VARCHAR` column instead of `INT`.

### OS:

x86_64

### DuckDB Version:

1.4.2

### DuckLake Version:

77f2512

### DuckDB Client:

CLI

### Full Name:

Jack McAsh

### Affiliation:

Huawei

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

pdet:
Thanks for the report! I managed to reproduce it, I'll have a look.

pdet:
Should be fixed by https://github.com/duckdb/ducklake/pull/603

PR #603, opened 2025-12-05
[BUG] Delete old deletion files if they are no longer relevant in a transaction

pdet:
Fix: https://github.com/duckdb/ducklake/issues/597

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1613,27 +1613,6 @@ shared_ptr<DuckLakeInlinedData> DuckLakeTransaction::GetTransactionLocalInlinedD
 	return result;
 }
 
-void DuckLakeTransaction::DropTransactionLocalDeleteFile(TableIndex table_id, const string &path) {
-	auto entry = table_data_changes.find(table_id);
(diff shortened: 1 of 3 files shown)


src/storage/ducklake_transaction.cpp
```diff
@@ -1613,6 +1613,27 @@ shared_ptr<DuckLakeInlinedData> DuckLakeTransaction::GetTransactionLocalInlinedD
...
 
+void DuckLakeTransaction::DropTransactionLocalDeleteFile(TableIndex table_id, const string &path) {
+	auto entry = table_data_changes.find(table_id);
+	if (entry == table_data_changes.end()) {
+		// No file to remove delete from, was removed before.
+		return;
+	}
+	auto &table_changes = entry->second.new_data_files;
+	for (auto &table_file : table_changes) {
+		if (table_file.delete_file) {
```
Mytherin: Is this second function necessary? Can't we just always delete the corresponding delete file in `DropTransactionLocalFile`?

i.e. add this line in that function:

```cpp
if (table_file.delete_file) {
    fs.RemoveFile(table_file.delete_file->file_name);
}
```

Issue #531, opened 2025-10-26
TransactionContext Error on ducklake_add_data_files()


jwass:
### What happens?

I was following some of the steps on the recent frozen ducklake blog post https://ducklake.select/2025/10/24/frozen-ducklake/ using data from Overture Maps Foundation (https://overturemaps.org).

When calling `ducklake_add_data_files` I'm getting a `TransactionContext Error` "failed to write column stats" error about casting an INT128 to INT64. I'm unsure where the int128 is coming from as no column in the parquet file has that type.

I simplified to the repro below using a single file for creating the table and adding the data file.

### To Reproduce

Contents of `script.sql`

```
load ducklake;
attach 'ducklake:lake.ducklake' as lake (DATA_PATH 'tmp_always_empty');
use lake;
create table buildings as select * from read_parquet('https://s3-us-west-2.amazonaws.com/overturemaps-us-west-2/release/2025-09-24.0/theme=buildings/type=building/part-00000-93d280c5-9152-4ecc-b500-d9681495941f-c000.zstd.parquet') with no data;
call ducklake_add_data_files('lake', 'buildings', 'https://s3-us-west-2.amazonaws.com/overturemaps-us-west-2/release/2025-09-24.0/theme=buildings/type=building/part-00000-93d280c5-9152-4ecc-b500-d9681495941f-c000.zstd.parquet');

```

Running on the command line with `duckdb -f script.sql`

```
TransactionContext Error:
Failed to commit: Failed to commit DuckLake transaction.
Failed to write column stats information to DuckLake: Type INT128 with value 18446744073709551434 can't be cast because the value is out of range for the destination type INT64

LINE 1: ...4014, 'alternate', 'short', NULL, NULL),(1, 1, 31, 119658, 18446744073709551434, 10815532, NULL, NULL, NULL, NULL)...
                                                                      ^
```

### OS:

MacOS Sequoia 15.6.1  x86_64

### DuckDB Version:

v1.4.1 (Andium) b390a7c376

### DuckLake Version:

f134ad8

### DuckDB Client:

CLI

### Full Name:

Jacob Wasserman

### Affiliation:

Meta

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

[...]

PR #611, opened 2025-12-10
Use num_values from stats if possible

pdet:
For nested types, lists more specifically, we can't calculate the number of `value_count` as:
```
column_stats.value_count = to_string(file.row_count - stats.null_count);
```
Since a list can have more elements than a row group. 

Hence, we have to use `num_values` from the parquet metadata.

Fix: https://github.com/duckdb/ducklake/issues/531

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1174,10 +1174,17 @@ DuckLakeFileInfo DuckLakeTransaction::GetNewDataFile(DuckLakeDataFile &file, Duc
 		if (stats.has_null_count) {
 			if (stats.has_num_values) {
 				column_stats.value_count = to_string(stats.num_values);
+				column_stats.null_count = to_string(stats.null_count);
 			} else {
-				column_stats.value_count = to_string(file.row_count - stats.null_count);
+				if (stats.null_count > file.row_count) {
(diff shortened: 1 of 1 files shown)


Mytherin (approved):
Thanks! LGTM

Mytherin:
Actually - do we write `num_values` in the Parquet writer for nested types? If so could we add a test that triggers this?

src/storage/ducklake_transaction.cpp
```diff
@@ -1172,7 +1172,11 @@ DuckLakeFileInfo DuckLakeTransaction::GetNewDataFile(DuckLakeDataFile &file, Duc
 		column_stats.max_val = stats.has_max ? DuckLakeUtil::StatsToString(stats.max) : "NULL";
 		column_stats.column_size_bytes = to_string(stats.column_size_bytes);
 		if (stats.has_null_count) {
-			column_stats.value_count = to_string(file.row_count - stats.null_count);
+			if (stats.has_num_values) {
+				column_stats.value_count = to_string(stats.num_values);
+			} else {
+				column_stats.value_count = to_string(file.row_count - stats.null_count);
```
Mytherin: We should probably still deal with `stats.null_count > file.row_count` - e.g. by ignoring the stats

PR #622, opened 2025-12-16
Batch SQL Statements to avoid unnecessary round-tripping

pdet:
This PR batches all queries that were previously executed separately during a commit operation.
When we commit, we run all modifications (e.g., create snapshots, add new files, alter schemas, add inlined data) in a single batch, thereby drastically reducing the number of DBMS round-trips required.

### DIFF
--- a/src/metadata_manager/postgres_metadata_manager.cpp
+++ b/src/metadata_manager/postgres_metadata_manager.cpp
@@ -50,10 +50,7 @@ unique_ptr<QueryResult> PostgresMetadataManager::Execute(DuckLakeSnapshot snapsh
 	query = StringUtil::Replace(query, "{METADATA_SCHEMA_ESCAPED}", schema_identifier_escaped);
 	query = StringUtil::Replace(query, "{METADATA_PATH}", metadata_path);
 	query = StringUtil::Replace(query, "{DATA_PATH}", data_path);
-
-	string pg_query = "CALL postgres_execute(" + catalog_literal + ", ";
-	pg_query += "$$" + query + "$$);";
-	return connection.Query(pg_query);
+	return connection.Query(StringUtil::Format("CALL postgres_execute(%s, %s)", catalog_literal, SQLString(query)));
 }
 
 string PostgresMetadataManager::GetLatestSnapshotQuery() const {


Alex-Monahan:
Awwwww yeah! I was hoping for this as soon as I saw the TPS tests flying around on Twitter... Psyched to benchmark    this!!

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2027,42 +2089,33 @@ void DuckLakeMetadataManager::WriteNewDataFiles(DuckLakeSnapshot commit_snapshot
...
-void DuckLakeMetadataManager::DropDataFiles(DuckLakeSnapshot commit_snapshot, const set<DataFileIndex> &dropped_files) {
-	FlushDrop(commit_snapshot, "ducklake_data_file", "data_file_id", dropped_files);
+void DuckLakeMetadataManager::DropDataFiles(string &batch_query, const set<DataFileIndex> &dropped_files) {
+	FlushDrop(batch_query, "ducklake_data_file", "data_file_id", dropped_files);
 }
 
-void DuckLakeMetadataManager::DropDeleteFiles(DuckLakeSnapshot commit_snapshot,
-                                              const set<DataFileIndex> &dropped_files) {
-	FlushDrop(commit_snapshot, "ducklake_delete_file", "data_file_id", dropped_files);
+void DuckLakeMetadataManager::DropDeleteFiles(string &batch_query, const set<DataFileIndex> &dropped_files) {
```
Mytherin: Instead of passing this `string` around everywhere, can we turn this into `string GetDropDeleteFilesQuery(...)` that returns the to-be-executed query?

I think that's a lot cleaner / more clear from an API perspective. This could then also be overwritten by various metadata managers.

src/storage/ducklake_transaction.cpp
```diff
@@ -1410,15 +1420,21 @@ void DuckLakeTransaction::FlushChanges() {
...
 
-			WriteSnapshotChanges(commit_state, transaction_changes);
+			WriteSnapshotChanges(batch_queries, commit_state, transaction_changes);
 			if (SchemaChangesMade()) {
 				// Insert our new schema in our table that tracks schema changes
-				metadata_manager->InsertNewSchema(commit_snapshot);
+				metadata_manager->InsertNewSchema(commit_snapshot, batch_queries);
+			}
+
+			auto res = Query(commit_snapshot, batch_queries);
```
Mytherin: This batching is a good improvement - but it's not actually avoiding round-trips with Postgres currently.

In order to avoid round-trips to Postgres, we need to batch everything into a call to `postgres_execute`, e.g.:

```sql
CALL postgres_execute('metadata_db', $$
<query 1>
<query 2>
...
$$);
```

That way all queries get executed as a single block within Postgres, instead of having them be executed as a single block within DuckDB, which will then send multiple queries back and forth to Postgres.

The way I would implement this is to have this be an `ExecuteQuery(...)` method in the `MetadataManager`, e.g. this becomes:

```sql
auto res = metadata_manager->Execute(commit_snapshot, batch_queries);
// ...
```

The `PostgresMetadataManager` can then overwrite this and wrap these queries in `postgres_execute`.

Note that that also means the queries no longer execute within DuckDB, but they execute within Postgres. This means that:
1. We can no longer use DuckDB-specific functions or extensions in the queries - it all has to be compatible with Postgres
2. The way we refer to tables changes (i.e. it is no longer `{metadata_store}.{metadata_schema}.{table}` but just `{metadata_schema}`

Point 2 is relatively easy to fix as we have `{METADATA_CATALOG}` we can replace with the appropriate prefix. Point 1 might be harder - that depends on the queries we're using here. I think for most queries this is relatively straightforward, but there might be some more difficult ones in there.

I think the way to achieve this would be to rewrite the functions in the metadata manager as mentioned (i.e. `GetDropDataFilesQuery(...)`) - and then try to rewrite queries to work in both DuckDB and Postgres as much as possible. For any where this does not work out - provide an override for that function in the `PostgresMetadataManager` to implement the query for Postgres specifically.

src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -18,6 +22,40 @@ bool PostgresMetadataManager::TypeIsNativelySupported(const LogicalType &type) {
...
+	query = StringUtil::Replace(query, "{METADATA_CATALOG_NAME_LITERAL}", catalog_literal);
+	query = StringUtil::Replace(query, "{METADATA_CATALOG_NAME_IDENTIFIER}", catalog_identifier);
+	query = StringUtil::Replace(query, "{METADATA_SCHEMA_NAME_LITERAL}", schema_literal);
+	query = StringUtil::Replace(query, "{METADATA_CATALOG}", schema_identifier);
+	query = StringUtil::Replace(query, "{METADATA_SCHEMA_ESCAPED}", schema_identifier_escaped);
+	query = StringUtil::Replace(query, "{METADATA_PATH}", metadata_path);
+	query = StringUtil::Replace(query, "{DATA_PATH}", data_path);
+
+	string pg_query = "CALL postgres_execute(" + catalog_literal + ", ";
+	pg_query += "$$" + query + "$$);";
```
Mytherin: I would use single-quotes here. Otherwise all the code above needs to make sure to escape `$$` correctly which is likely somewhat painful. Escaping single quotes is much easier:

`StringUtil::Format('CALL postgres_execute(%s, %s)', SQLString(catalog_literal), SQLString(query));`

We should try to use that everywhere instead of doing string concatenation.

pdet:
> Awwwww yeah! I was hoping for this as soon as I saw the TPS tests flying around on Twitter... Psyched to benchmark this!!

This should yield a ~4x better performance.

djouallah:
Force install from core nightly don't seem to work

pdet:
> Force install from core nightly don't seem to work

Thanks, should be fixed :)

PR #624, opened 2025-12-16
Implement compaction of files with partial file info

### LINKED ISSUE #551: TransactionContext error when executing ducklake_rewrite_data_files
### What happens?

I am getting the following error when trying to call ducklake_rewrite_data_files:

```
Only the first compacted file can have existing partial file info
```

### To Reproduce

Not sure how to reproduce this at this point. Would appreciate any pointers on how to debug this issue, maybe I could get a reproducible example.

### OS:

ubuntu-24.04-arm64

### DuckDB Version:

1.4.1

### DuckLake Version:

0.3

### DuckDB Client:

duckdb-go 2.5.0

### Hardware:

_No response_

### Full Name:

Maxim Zudilov

### Affiliation:

Groovy Media

- [ ] Yes, I have

- [x] Yes, I have

redox:
Following https://github.com/duckdb/ducklake/issues/551 we gave a try at implementing compaction of multiple files with existing partial file info with the team @altertable-ai. So far, it would fail with the error "Only the first compacted file can have existing partial file info".

This issue occurred when using `REWRITE_DELETES` compaction followed by `merge_adjacent_files` on files that already had partial file information. Shout out to @adileo for providing a reproduction case.

This PR modifies the compaction logic to properly handle multiple files with partial file info by:
 - Removing the restriction that only the first file can have partial file info
 - Shifting row counts in partial file info by the accumulated row limit from previously processed files
 - Properly merging partial file info from all source files into the new compacted file

We added a `compact_partial_file_info.test` that reproduces the original
issue and verifies the fix and updated `merge_adjacent_max_files.test` as the resulting snapshots (2 of them having partial file info) can be further merged.

While our initial tests are showing promising results, we would love to have the experts' point of views!

Fix #551

### DIFF
--- a/test/sql/compaction/merge_rewrite_partial_file_info.test
+++ b/test/sql/compaction/merge_rewrite_partial_file_info.test
@@ -1,4 +1,4 @@
-# name: test/sql/compaction/merge_partial_file_info.test
+# name: test/sql/compaction/merge_rewrite_partial_file_info.test
 # description: Reproduce issue ##551 with REWRITE_DELETES and partial file info
 # group: [compaction]
 


test/sql/compaction/merge_rewrite_partial_file_info.test
```diff
@@ -0,0 +1,118 @@
...
+statement ok
+CALL ducklake_rewrite_data_files('ducklake', 't', delete_threshold => 0);
+
+# Merge all of them into a single file
+statement ok
+CALL ducklake_merge_adjacent_files('ducklake', 't');
+
+# Verify that we ended up with a single file
+query I
+SELECT COUNT(*) FROM ducklake_list_files('ducklake', 't');
```
pdet, line 115: Could you check that all snapshots produce correct results?

pdet (commented):
Hi @redox thanks for the PR, I just have one comment to add a few more tests, basically checking that time travel is preserved over all snapshots.

Could you also retarget your PR to v1.4?

test/sql/compaction/merge_rewrite_partial_file_info.test
redox: Yes, good catch; and some are incorrect - will give this a try

redox:
> Hi @redox thanks for the PR, I just have one comment to add a few more tests, basically checking that time travel is preserved over all snapshots.
> 
> Could you also retarget your PR to v1.4?

@pdet done, let us know how this looks - we're still navigating the internals of the delete files and partial_file_info so willing to make sure we don't miss something 🙌

pdet:
Thanks for adding the test! I think it makes sense to not write partial information to newly written files from deletion rewrites.

Since you are removing the merging constraint in the partial file info, can you add one more test where you do a merge adjacent of two files with partial info, and verify that it is correct with time-travelling through the snapshots?
I think you can achieve that by playing with the `target_file_size`.

redox:
> Since you are removing the merging constraint in the partial file info, can you add one more test where you do a merge adjacent of two files with partial info, and verify that it is correct with time-travelling through the snapshots?
> I think you can achieve that by playing with the `target_file_size`.

Done!

pdet:
Hi @redox thanks for all the changes! I think it looks great.

Maybe @Mytherin also wants to have a quick look.

Mytherin:
Thanks! LGTM

redox:
Great to hear @Mytherin @pdet - was fun to dig a bit more into the compact internals :)

Coming from the search industry, I started thinking about it a bit more and opened https://github.com/duckdb/ducklake/discussions/627 - would love to hear your thoughts when you have time.

PR #638, opened 2025-12-22
Move all queries on metadata to DuckLakeMetadataManager

qsliu2017:
I'm working on integrating ducklake with embed duckdb, e.g. pg_duckdb. This PR might be helpful to implement custom metadata manager.

Specifically, this PR:
1. move all queries on metadata to `DuckLakeMetadataManager::Query`, so that custom implementation can override it to handle metadata queries.
2. `DuckLakeTransaction::Query` only process queries on duckdb.
3. add `DuckLakeMetadataManager::Register(const string &name, DuckLakeMetadataManager::create_t create)` to register custom MetadataManager.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -30,11 +30,18 @@ optional_ptr<AttachedDatabase> GetDatabase(ClientContext &context, const string 
 unordered_map<string /* name */, DuckLakeMetadataManager::create_t> DuckLakeMetadataManager::metadata_managers = {
     {"postgres", PostgresMetadataManager::Create}, {"postgres_scanner", PostgresMetadataManager::Create}};
 
-bool DuckLakeMetadataManager::Register(const string &name, DuckLakeMetadataManager::create_t create) {
-	return metadata_managers.emplace(name, create).second;
+mutex DuckLakeMetadataManager::metadata_managers_lock;
+
+void DuckLakeMetadataManager::Register(const string &name, DuckLakeMetadataManager::create_t create) {
+	lock_guard<mutex> lock(metadata_managers_lock);
+	if (metadata_managers.find(name) != metadata_managers.end()) {
+		throw InternalException("Metadata manager with name \"%s\" already exists!", name);
+	}
+	metadata_managers[name] = create;
 }
 
 unique_ptr<DuckLakeMetadataManager> DuckLakeMetadataManager::Create(DuckLakeTransaction &transaction) {
+	lock_guard<mutex> lock(metadata_managers_lock);
 	auto &catalog = transaction.GetCatalog();
 	auto catalog_type = catalog.MetadataType();
 	auto create = metadata_managers[catalog_type];
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -99,7 +99,7 @@ public:
 	virtual ~DuckLakeMetadataManager();
 
 	typedef unique_ptr<DuckLakeMetadataManager> (*create_t)(DuckLakeTransaction &transaction);
-	static bool Register(const string &name, create_t);
+	static void Register(const string &name, create_t);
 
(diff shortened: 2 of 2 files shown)


src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -1,7 +1,10 @@
 #include "metadata_manager/postgres_metadata_manager.hpp"
+
+#include <utility>
```
pdet: Is this include necessary?

pdet (commented):
Hi @qsliu2017, thanks for the PR!
I’m a bit confused about why you wouldn’t need to replace these query variables if you were going to integrate with a different catalog. Can you expand on that?

qsliu2017:
> why you wouldn’t need to replace these query variables if you were going to integrate with a different catalog.

Do you mean `StringUtil::Replace(query, "{METADATA_CATALOG_NAME_LITERAL}", catalog_literal);`? They've been moved to `DuckLakeMetadataManager`.

In my implementation, queries on metadata will not go through `DuckLakeTransaction::Query`, so it's better that we replace catalog related variables here.

src/metadata_manager/postgres_metadata_manager.cpp
qsliu2017: might be auto import 😅

qsliu2017:
just added metadata manager registration. @pdet would you please take a look?

src/storage/ducklake_transaction.cpp
```diff
@@ -1470,7 +1471,9 @@ void DuckLakeTransaction::FlushChanges() {
 			if (res->HasError()) {
 				res->GetErrorObject().Throw("Failed to flush changes into DuckLake: ");
 			}
-			connection->Commit();
+			if (connection) {
+				connection->Commit();
+			}
```
pdet, line 1476: Is this a possible scenario?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -26,11 +27,19 @@ DuckLakeMetadataManager::~DuckLakeMetadataManager() {
 }
 optional_ptr<AttachedDatabase> GetDatabase(ClientContext &context, const string &name);
 
+unordered_map<string /* name */, DuckLakeMetadataManager::create_t> DuckLakeMetadataManager::metadata_managers = {
+    {"postgres", PostgresMetadataManager::Create}, {"postgres_scanner", PostgresMetadataManager::Create}};
+
+bool DuckLakeMetadataManager::Register(const string &name, DuckLakeMetadataManager::create_t create) {
+	return metadata_managers.emplace(name, create).second;
```
pdet: What happens if we try to register an already existing manager?

qsliu2017: do nothing and returns false, according to https://en.cppreference.com/w/cpp/container/unordered_map/emplace.html

pdet: Sorry, I was not clear with my message.

It is best to throw an error if an entry already exists.
For example, registering encodings in DuckDB would:

```cpp
void DBConfig::RegisterEncodeFunction(const EncodingFunction &function) const {
	lock_guard<mutex> l(encoding_functions->lock);
	const auto name = function.GetName();
	if (encoding_functions->functions.find(name) != encoding_functions->functions.end()) {
		throw InvalidInputException("Decoding function with name %s already registered", name);
	}
	encoding_functions->functions[name] = function;
}

```

src/storage/ducklake_transaction.cpp
qsliu2017: that's for our implement. we don't use `connection` to execute queries, so it's never initialized.

src/storage/ducklake_metadata_manager.cpp
qsliu2017: done.

qsliu2017:
comments addressed. please take another look. @pdet

pdet:
Thanks, LGTM!

PR #642, opened 2025-12-22
Ordered compaction and inlining with full catalog integration

Alex-Monahan:
Hi folks!

This is a new PR meant to solve the same use case as https://github.com/duckdb/ducklake/pull/593, but addressing the PR feedback! Thank you for the guidance - it was super helpful. I am still open to any changes you recommend!

## PR Overview

The purpose of this PR is to sort data while it is written to speed up selective read queries in the future.

This uses the pre-existing DuckDB `SET SORTED BY` syntax (from here https://github.com/duckdb/duckdb/pull/16714) to sort data when it is compacted or inlined data is flushed. For example, 
```sql
ALTER TABLE ducklake.my_table SET SORTED BY (sort_key_1 ASC, sort_key_2 DESC);
```

Then when either `ducklake_merge_adjacent_files` or `ducklake_flush_inlined_data` are called, those operations will sort the data prior to writing it out as parquet.

## New Tables in DuckLake Spec

This adds 2 new tables to the DuckLake spec: `ducklake_sort_info` and `ducklake_sort_expression`. 

`ducklake_sort_info` keeps a version history of the sort settings for tables over time. It has 1 row per time that a table has a new sort setting applied. (The prior PR used an option for this, but that has been removed based on the feedback).
```sql
CREATE TABLE {METADATA_CATALOG}.ducklake_sort_info(
  sort_id BIGINT,
  table_id BIGINT,
  begin_snapshot BIGINT,
  end_snapshot BIGINT
);
```

`ducklake_sort_expression` tracks the details of that sort. For each time a new sort setting is applied, this table includes one row for each expression in the order by. (If I order by `column3 asc, column42 desc, column64 asc`, then there will be 3 rows.) 
```sql
CREATE TABLE {METADATA_CATALOG}.ducklake_sort_expression(
  sort_id BIGINT,
  table_id BIGINT,
  sort_key_index BIGINT,     -- The sequence the SORTED BY expressions are evaluated in
  expression VARCHAR,
  dialect VARCHAR,
  sort_direction VARCHAR,    -- ASC or DESC
  null_order VARCHAR         -- NULLS_LAST or NULLS_FIRST
);
```

## Future Work / Limitations

There are still a few limitations with this PR:
1. This does not order during insert (only during compaction and inline flush). 
    * I would love to try to do a follow up PR to add this!
    * A user can still specify their own `ORDER BY` during an insert, so there is a workaround for the moment
1. Only explicit column names can be used in the sorting, not expressions.
    * I have plans to add this in a follow up PR
    * There is a friendly error message (and tests) to document this limitation.
    * The spec has a column `expression`, so the intention was to make the spec itself forwards-compatible with expression-oriented sorting.
1. Files are still selected for compaction based on insertion order. It could be better to sort the list of files by min/max metadata before selecting files for compaction.
    * Let me know if this is desirable and I can work on it after the "order during insert" in number 1!

I believe that I made this fully compatible with the batching code, but I was testing locally on a DuckDB catalog and not on Postgres. Any extra eyes on that side would be great!

If this looks good, I can also do any docs PRs that you recommend - happy to help there.

Thanks folks! CC @philippmd as well as an FYI

### DIFF
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -206,7 +206,6 @@ struct DuckLakePartitionInfo {
 
 struct DuckLakeSortFieldInfo {
 	idx_t sort_key_index = 0;
-	// TODO: Validate that expression is case insensitive when stored
 	string expression;
 	string dialect;
 	OrderType sort_direction;
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -983,7 +983,7 @@ void DuckLakeTransaction::GetNewTableInfo(DuckLakeCommitState &commit_state, Duc
 			// note that in case of nested types we might be dropping multiple columns here
 			HandleChangedFields(commit_state.GetTableId(table), table.GetChangedFields(), result,
 			                    columns_handled_by_later_ops);
-      transaction_changes.altered_tables_with_schema_version_changes.insert(table_id);
+			transaction_changes.altered_tables_with_schema_version_changes.insert(table_id);
 			column_schema_change = true;
 			break;
 		}
--- a/test/sql/sorted_table/merge_adjacent_sorted_case_insensitivity.test
+++ b/test/sql/sorted_table/merge_adjacent_sorted_case_insensitivity.test
@@ -1,0 +1,132 @@
+# name: test/sql/sorted_table/merge_adjacent_sorted_case_insensitivity.test
+# description: test ducklake merge adjacent files with SET SORTED BY when column case does not match the table
+# group: [sorted_table]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '__TEST_DIR__/merge_adjacent_options_sorted_case_insensitivity/')
+
+statement ok
+USE ducklake;
+
+# Create one Table with columns that have upper and lower case
+statement ok
+CREATE TABLE sort_on_compaction (unique_id BIGINT, Sort_Key_1 BIGINT, Sort_Key_2 VARCHAR);
+
+statement ok
+INSERT INTO sort_on_compaction (unique_id, sort_key_1, sort_key_2)
+FROM range(4) t(i)
+SELECT 
+    i AS unique_id,
+    i % 2 AS sort_key_1,
+    'woot' || i AS sort_key_2
+ORDER BY 
+    i DESC
+;
+
+statement ok
+INSERT INTO sort_on_compaction (unique_id, sort_key_1, sort_key_2)
+FROM range(4) t(i)
+SELECT 
+    i + 4 AS unique_id,
+    (i + 4) % 2 AS sort_key_1,
+    'woot' || (i + 4) AS sort_key_2
+ORDER BY 
+    i DESC
+;
+
+query III
+FROM sort_on_compaction
+----
+3	1	woot3
+2	0	woot2
+1	1	woot1
+0	0	woot0
+7	1	woot7
+6	0	woot6
+5	1	woot5
+4	0	woot4
+
+# SET SORTED BY should be a case insensitive operation (when original columns are case sensitive)
+statement ok
+ALTER TABLE ducklake.sort_on_compaction SET SORTED BY (SORT_KEY_1 ASC NULLS LAST, SoRt_KeY_2 ASC NULLS LAST);
+
+statement ok
+CALL ducklake_merge_adjacent_files('ducklake', 'sort_on_compaction');
+
+query III
+FROM sort_on_compaction
+----
+0	0	woot0
+2	0	woot2
+4	0	woot4
+6	0	woot6
+1	1	woot1
+3	1	woot3
+5	1	woot5
+7	1	woot7
+
+
+# Create one Table with columns that are all lower case
+statement ok
+CREATE TABLE sort_on_compaction_2 (unique_id BIGINT, sort_key_1 BIGINT, sort_key_2 VARCHAR);
+
+statement ok
+INSERT INTO sort_on_compaction_2 (unique_id, sort_key_1, sort_key_2)
+FROM range(4) t(i)
+SELECT 
+    i AS unique_id,
+    i % 2 AS sort_key_1,
+    'woot' || i AS sort_key_2
+ORDER BY 
+    i DESC
+;
+
+statement ok
+INSERT INTO sort_on_compaction_2 (unique_id, sort_key_1, sort_key_2)
+FROM range(4) t(i)
+SELECT 
+    i + 4 AS unique_id,
+    (i + 4) % 2 AS sort_key_1,
+    'woot' || (i + 4) AS sort_key_2
+ORDER BY 
+    i DESC
+;
+
+query III
+FROM sort_on_compaction_2
+----
+3	1	woot3
+2	0	woot2
+1	1	woot1
+0	0	woot0
+7	1	woot7
+6	0	woot6
+5	1	woot5
+4	0	woot4
+
+# SET SORTED BY should be a case insensitive operation (when original columns are lower case)
+statement ok
+ALTER TABLE ducklake.sort_on_compaction_2 SET SORTED BY (SORT_KEY_1 ASC NULLS LAST, SoRt_KeY_2 ASC NULLS LAST);
+
+statement ok
+CALL ducklake_merge_adjacent_files('ducklake', 'sort_on_compaction_2');
+
+query III
+FROM sort_on_compaction_2
+----
+0	0	woot0
+2	0	woot2
+4	0	woot4
+6	0	woot6
+1	1	woot1
+3	1	woot3
+5	1	woot5
+7	1	woot7
+


Alex-Monahan:
The failing test did not reproduce for me locally, so I made a commit with just a comment change to force a CI re-run in the hopes that it was a flaky test. The test is not an area of the code that I think I should have edited, but, there are reasons that we have tests... 

Failing test was `test/sql/data_inlining/data_inlining_flush_schema.test`.

Alex-Monahan:
Hmm, this is a different failing test and does not appear related. Please let me know if you would like for me to investigate, otherwise it seems random...

Alex-Monahan:
Thinking ahead to adding the insert capability, is the right spot to add that in ducklake_insert.cpp in `DuckLakeInsert::PlanCopyForInsert`, right before creating the `auto &physical_copy` ?

pdet:
> Hmm, this is a different failing test and does not appear related. Please let me know if you would like for me to investigate, otherwise it seems random...

There are a couple of spurious tests failing that I still have to track down, the ones I have seen here are unrelated to your PR.

pdet:
> Thinking ahead to adding the insert capability, is the right spot to add that in ducklake_insert.cpp in `DuckLakeInsert::PlanCopyForInsert`, right before creating the `auto &physical_copy` ?

That sounds about right!

Alex-Monahan:
I will resolve the conflicts once I am back from vacation and have access to a laptop once again (in the next few days)!

pdet:
No worries @Alex-Monahan! Merging it with main should also resolve any failures unrelated to this PR.

Alex-Monahan:
Merged with main! The tests I added for the PR are passing locally, so it would be great to give CI another shot!

Alex-Monahan:
Are you suuuure I can't rename the table to `ducklake_sort_info` ? :-)

pdet:
Oh, I'm totally fine with `ducklake_sort_info`, the description of the tables I wrote in the other PR was more of a guide not really a strict name ruling.

Before we move forward, let's wait for @Mytherin to also have a quick look if he is happy with the the spec changes.

Alex-Monahan:
Welcome back @Mytherin , hope you had an excellent time away! I'd love your feedback - happy to address changes big or small. Sorry to ping, I just know that DuckLake 1.0 is not far away, and it takes me a bit of time to make C++ edits! :-) 

Many thanks!!

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -88,6 +88,7 @@ CREATE TABLE {METADATA_CATALOG}.ducklake_schema_versions(begin_snapshot BIGINT,
 CREATE TABLE {METADATA_CATALOG}.ducklake_macro(schema_id BIGINT, macro_id BIGINT, macro_name VARCHAR, begin_snapshot BIGINT, end_snapshot BIGINT);
 CREATE TABLE {METADATA_CATALOG}.ducklake_macro_impl(macro_id BIGINT, impl_id BIGINT, dialect VARCHAR, sql VARCHAR, type VARCHAR);
 CREATE TABLE {METADATA_CATALOG}.ducklake_macro_parameters(macro_id BIGINT, impl_id BIGINT,column_id BIGINT, parameter_name VARCHAR, parameter_type VARCHAR, default_value VARCHAR, default_value_type VARCHAR);
+CREATE TABLE {METADATA_CATALOG}.ducklake_sort_info(sort_id BIGINT, table_id BIGINT, begin_snapshot BIGINT, end_snapshot BIGINT, sort_key_index BIGINT, expression VARCHAR, dialect VARCHAR, sort_direction VARCHAR, null_order VARCHAR);
```
Mytherin: I think we need two tables (`ducklake_sort_info` and `ducklake_sort_column`) to correctly allow for multi-column sorts.

I'm also not sure if we should have arbitrary expressions here - maybe this should use the same `transform` spec as the partition keys?

Mytherin (commented):
Thanks! I've left a comment on the spec side of things - haven't looked at the code

src/storage/ducklake_metadata_manager.cpp
Alex-Monahan: Thanks @Mytherin!

The current version does support multi-column sorting, but maybe the table structure is less ideal. Multiple `sort_id`'s can be valid for a table at the same time, one for each column in the sort. Maybe there could be a `sort_id` and a `sort_expression_id` ? Today `sort_key_index` is what tracks the order to apply a multi column sort (0, 1, 2, etc.). 

Iceberg does use that transform approach. I think this expression approach feels like it is taking some inspiration from DuckLake macros. One goal I have is to easily allow z-order and Hilbert sorts, but I haven't thought through how to easily add those as transforms without `spatial`. 

I'll research the other formats to see how they handle z-ordering.

Alex-Monahan: As far as I can tell, Z-Ordering is not supported by the [Iceberg sorting spec](https://iceberg.apache.org/spec/#sorting). They only support sorting by columns or the same transforms as partitioning. However, the Spark Iceberg client has a [rewrite_data_files procedure](https://iceberg.apache.org/docs/1.6.0/spark-procedures/?h=zorder#rewrite_data_files), which has a specific zorder parameter. I think z-ordering is the best practice (?) so it looks quite tied to Spark.

Delta uses [Liquid clustering](https://docs.databricks.com/aws/en/delta/clustering#key-selection-guidelines) and has deprecated z-ordering. Liquid clustering replaces both partitioning and sorting. They only allow column names (no transforms even). 

My primary end goal is easy zorder or Hilbert sorting across multiple columns. I think it would be great to avoid storing that expression as a column in the data if possible, since I'm not sure it would compress super well. 

Is there any chance we could keep the arbitrary expression concept? I can split into 2 tables like partitioning if you prefer!

Mytherin: I think splitting into two tables is better. We can keep expressions instead of transforms - that's fine by me

Alex-Monahan: Edit: sorry, posted this without refreshing the page and seeing your comment above!


Another benefit of an expression is that a zorder/Hilbert sort order could be added to a table later on without adding a column and forcing any `select *` queries needing to be updated.

Another complexity is that a zorder transform would apply to more than 1 column at a time, vs the single column transforms already in the spec for partitioning. 

If Discord or a call is easier, let me know! I appreciate your feedback!

Alex-Monahan: I will make the change into 2 tables as quick as I can! Thanks!

Alex-Monahan:
I've added in that separate table! Existing tests are passing with the join added - let me know if there are other things worth testing now that there are 2 tables involved!

Mind kicking off CI for me? I appreciate it!

src/storage/ducklake_metadata_manager.cpp
Alex-Monahan: I have split it out into 2 tables: `ducklake_sort_info` and `ducklake_sort_expression`! (I didn't use `_column` because there won't always be a way to always join back to the column table.)

Alex-Monahan:
I realized that I was allowing the `SET SORTED BY` command to complete even when the columns were not valid (a legacy from the old option approach that was hard to validate up front). I've now restricted it so that it will fail if the column is incorrect and added some tests!

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -303,6 +245,91 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
...
+	auto bindings = plan->GetColumnBindings();
+
+	vector<BoundOrderByNode> orders;
+
+	vector<OrderByNode> pre_bound_orders;
+
+	for (auto &pre_bound_order : sort_data->fields) {
+		if (pre_bound_order.dialect != "duckdb") {
+			continue;
+		}
```
pdet, line 269: I'm wondering if we should error if it is an unsupported dialect?

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -303,6 +245,91 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
...
+	for (idx_t col_idx = 0; col_idx < current_columns.size(); col_idx++) {
+		alias_map[current_columns[col_idx]] = col_idx;
+	}
+
+	root_get->ResolveOperatorTypes();
+	auto &root_types = root_get->types;
+
+	vector<std::string> unmatching_names;
+	for (auto &pre_bound_order : pre_bound_orders) {
+		std::string name = pre_bound_order.expression->GetName();
```
pdet: nit: auto

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -560,6 +564,45 @@ ORDER BY part.table_id, partition_id, partition_key_index
...
+
+		DuckLakeSortFieldInfo sort_field;
+		sort_field.sort_key_index = row.GetValue<uint64_t>(2);
+		sort_field.expression = row.GetValue<string>(3);
+		sort_field.dialect = row.GetValue<string>(4);
+
+		std::string sort_direction_str = StringUtil::Upper(row.GetValue<string>(5));
+		sort_field.sort_direction = (sort_direction_str == "DESC" ? OrderType::DESCENDING : OrderType::ASCENDING);
+
+		std::string null_order_str = StringUtil::Upper(row.GetValue<string>(6));
```
pdet: nit auto

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -560,6 +564,45 @@ ORDER BY part.table_id, partition_id, partition_key_index
...
+		}
+		auto &sort_entry = sorts.back();
+
+		DuckLakeSortFieldInfo sort_field;
+		sort_field.sort_key_index = row.GetValue<uint64_t>(2);
+		sort_field.expression = row.GetValue<string>(3);
+		sort_field.dialect = row.GetValue<string>(4);
+
+		std::string sort_direction_str = StringUtil::Upper(row.GetValue<string>(5));
+		sort_field.sort_direction = (sort_direction_str == "DESC" ? OrderType::DESCENDING : OrderType::ASCENDING);
```
pdet: Should these be case-insensitive?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2724,6 +2768,107 @@ WHERE table_id IN (%s) AND end_snapshot IS NULL
...
+	unordered_set<idx_t> old_sort_set;
+	for (auto &sort : old_sorts) {
+		old_sort_set.insert(sort.table_id.index);
+		if (new_sort_map.find(sort.table_id.index) != new_sort_map.end()) {
+			if (new_sort_map[sort.table_id.index] == sort) {
+				// If a new sort already exists in an old sort, it's a nop, we can remove it
+				new_sort_map.erase(sort.table_id.index);
+			}
+		}
+	}
```
pdet: Can this be a function? e.g., CheckTableSortEqual

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2724,6 +2768,107 @@ WHERE table_id IN (%s) AND end_snapshot IS NULL
...
+
+	vector<idx_t> sort_ids_to_erase;
+	for (auto &sort : new_sorts) {
+		if (old_sort_set.find(sort.table_id.index) == old_sort_set.end() && sort.fields.empty()) {
+			// If a map does not exist on the old sort and the sort has no fields, this is an reset over
+			// an empty sort definition, hence also a nop
+			sort_ids_to_erase.push_back(sort.table_id.index);
+		}
+	}
```
pdet, line 2747: Can this alsso be a function? e.g., ResetTableSort

src/storage/ducklake_table_entry.cpp
```diff
@@ -343,6 +357,25 @@ string GetPartitionColumnName(ColumnRefExpression &colref) {
 	return colref.GetColumnName();
 }
 
+string GetSortColumnName(DuckLakeTableEntry &table, ParsedExpression &expr) {
+	// Only allow column references, reject expressions
+	if (expr.type != ExpressionType::COLUMN_REF) {
+		throw BinderException("SET SORTED BY only supports column references, not expressions: %s", expr.ToString());
```
pdet: Should this be a NotImplementedException?

pdet (commented):
Hey Alex, the PR is looking super solid, thanks again for all the effort you put in!

I'm wondering if you could also add the following tests:
1. If you drop a table with an existing order, and then create a table with the same name but different columns?
2. If you get the `changes` information on `snapshot()`
3. If the alter statement won't intefeer with compaction (see, https://github.com/duckdb/ducklake/pull/671)
4. If we can also verify the new parquet files are correctly written directly (without acessing them from the table), and also if different files have the correct orders, when having different orders over multiple snapshots.
5. More transaction tests, for example, what happens if we begin a transaction and in the same transaction we add a new sorting order and alter a column name?
6. What happens if you do a batch insertion on a table with a set column order (not inlined)

The test files you are pretty detailed, but my bio llm context window is small, do you mind also moving them to a `sorted_table` folder and modularizing them a bit more?
 
I also guess in general we can do `StringUtil::Lower()` for catalog access when the result can be case insensitive.

src/functions/ducklake_compaction_functions.cpp
Alex-Monahan: Ah, I will look at the macros as an example. Would you prefer that I filter them out in SQL when I pull from the catalog?

src/storage/ducklake_metadata_manager.cpp
Alex-Monahan: I used UPPER as a crude way to do that, but I will learn the cleaner way!

src/storage/ducklake_table_entry.cpp
Alex-Monahan: Love that! Sure

src/functions/ducklake_compaction_functions.cpp
pdet: This is more of a question, I'm open to suggestions. On one hand,  we can be more permissive with anything dialect related, but I fear this might cause confusion, while an error removes any kind of ambiguity of the system's state.

Alex-Monahan: Looking back at this, the reason I implemented it this way was in case we wanted to deliberately have 2 sort orders active at once: one for DuckDB, and one for another engine like Spark for example. So if there were 2 separate sets of sorts, skipping the non-DuckDB ones (but not throwing an error) should allow them to co-exist. 

Unfortunately, this case is very difficult to test in this repo! I think as catalog editing gets added on the Spark side, we can do some testing in that repo.

So, do you mind if we leave it as-is?

src/storage/ducklake_metadata_manager.cpp
Alex-Monahan: I shamelessly stole this code from GetNewPartitions, so that could probably use the same love!

Alex-Monahan:
> Hey Alex, the PR is looking super solid, thanks again for all the effort you put in!
> 
> I'm wondering if you could also add the following tests:
> 
> 1. If you drop a table with an existing order, and then create a table with the same name but different columns?
> 2. If you get the `changes` information on `snapshot()`
> 3. If the alter statement won't intefeer with compaction (see, #671)
> 4. If we can also verify the new parquet files are correctly written directly (without acessing them from the table), and also if different files have the correct orders, when having different orders over multiple snapshots.
> 5. More transaction tests, for example, what happens if we begin a transaction and in the same transaction we add a new sorting order and alter a column name?
> 6. What happens if you do a batch insertion on a table with a set column order (not inlined)
> 
> The test files you are pretty detailed, but my bio llm context window is small, do you mind also moving them to a `sorted_table` folder and modularizing them a bit more?
> 
> I also guess in general we can do `StringUtil::Lower()` for catalog access when the result can be case insensitive.

Thanks as always for the feedback! These tests make sense and I'm working on them! I had a few follow up questions.

On 3, would you like for me to validate that an alter table statement on a different table does not interfere? (Ahead of time or in the same transaction?) Based on https://github.com/duckdb/ducklake/discussions/684, I was assuming that any alter table on the same table would be a hard line that prevents compaction - please let me know if I am wrong! 

On 4, let me try to repeat back to you what I think you are looking for! I think you want to test the situation where I insert into the same table 3 times, each time with a different sort order setting, and then I run 1 compaction on that table. My goal was to only consider the most recent ducklake_sort_expression setting, regardless of the value of the setting at insert time, so this test would ensure that! I could also test multiple compactions and make sure that the final compaction resorts to the latest sort setting.

On 6, I'm not sure exactly how to proceed - I could use your help! This PR does not include the capability to do the sorting during an insert, so I'm not sure how the features would intersect just yet. Does adjusting the column order during insert result in the parquet file having the columns stored in a different order? I'm not sure what behavior I should be exercising with the test. Just let me know!

pdet:
> > Hey Alex, the PR is looking super solid, thanks again for all the effort you put in!
> > I'm wondering if you could also add the following tests:
> > 
> > 1. If you drop a table with an existing order, and then create a table with the same name but different columns?
> > 2. If you get the `changes` information on `snapshot()`
> > 3. If the alter statement won't intefeer with compaction (see, [Change schema tracking from global to table. #671](https://github.com/duckdb/ducklake/pull/671))
> > 4. If we can also verify the new parquet files are correctly written directly (without acessing them from the table), and also if different files have the correct orders, when having different orders over multiple snapshots.
> > 5. More transaction tests, for example, what happens if we begin a transaction and in the same transaction we add a new sorting order and alter a column name?
> > 6. What happens if you do a batch insertion on a table with a set column order (not inlined)
> > 
> > The test files you are pretty detailed, but my bio llm context window is small, do you mind also moving them to a `sorted_table` folder and modularizing them a bit more?
> > I also guess in general we can do `StringUtil::Lower()` for catalog access when the result can be case insensitive.
> 
> Thanks as always for the feedback! These tests make sense and I'm working on them! I had a few follow up questions.
> 
> On 3, would you like for me to validate that an alter table statement on a different table does not interfere? (Ahead of time or in the same transaction?) Based on #684, I was assuming that any alter table on the same table would be a hard line that prevents compaction - please let me know if I am wrong!
> 
Basically, we should check that the alter statement does not increase the schema counter of a table, since the alter statement here does not change the actual schema, and the sorting is somewhat "optional" I believe that files should still be compactable before/after the alter.

> On 4, let me try to repeat back to you what I think you are looking for! I think you want to test the situation where I insert into the same table 3 times, each time with a different sort order setting, and then I run 1 compaction on that table. My goal was to only consider the most recent ducklake_sort_expression setting, regardless of the value of the setting at insert time, so this test would ensure that! I could also test multiple compactions and make sure that the final compaction resorts to the latest sort setting.

I wanted to see if files would always be generated with the latest sorting (e.g., by flushing) , and also compacted with the latest sorting, but now that I think about it again, maybe is an overkill because you already have similar tests to that, let's skip it.

> On 6, I'm not sure exactly how to proceed - I could use your help! This PR does not include the capability to do the sorting during an insert, so I'm not sure how the features would intersect just yet. Does adjusting the column order during insert result in the parquet file having the columns stored in a different order? I'm not sure what behavior I should be exercising with the test. Just let me know!

Let's skip this.

Alex-Monahan:
When investigating 3, I found that you are correct that I am currently incrementing the schema_version when altering the sort order. Dang! I think there are few other cases where we probably want to have the same "alter table, but don't change the schema_version". I'm thinking:
* SET_SORT_KEY 
* SET_COMMENT
* SET_COLUMN_COMMENT

I'll start trying to implement that behavior, but let me know if I need to course correct! From my first prototype, it requires some additional complexity in a few spots (I think we will need a different construct than `new_tables` for storing those kinds of alters), so I'm open to feedback and ideas!

pdet:
> When investigating 3, I found that you are correct that I am currently incrementing the schema_version when altering the sort order. Dang! I think there are few other cases where we probably want to have the same "alter table, but don't change the schema_version". I'm thinking:
> 
> * SET_SORT_KEY
> * SET_COMMENT
> * SET_COLUMN_COMMENT
> 
> I'll start trying to implement that behavior, but let me know if I need to course correct! From my first prototype, it requires some additional complexity in a few spots (I think we will need a different construct than `new_tables` for storing those kinds of alters), so I'm open to feedback and ideas!

Yes, I think none of those should impact the schema ranges for compaction. I think the main thing you have to do is to ensure that these statements won't write a new entry into `ducklake_schema_versions`, as this table keeps track of schema changes per ducklake table.

Alex-Monahan:
Yup, I will prevent that! Unfortunately, I'm also finding that I need to update the local schema cache (since usually a new schema version would be handled differently). I'm making progress!

Alex-Monahan:
I now have the schema_version not incrementing on `SET_SORT_KEY, SET_COMMENT, and SET_COLUMN_COMMENT`! 

I split the tests apart and also added tests for 1, 2, 3, and 5! I added quite a lot of tests to ensure transactionality for 5. Since `alter table` can't happen in the same transaction as a compaction (Getting the deliberate `Transactions can either make changes OR perform compaction - not both` error), that meant fewer cases in the compaction case than the inlining. 

I also added tests for the schema_version non-incrementing for the comment cases.

Let me know what you think!

Alex-Monahan:
So, to fix the assertion issues on my fork's CI, I had to relax an assertion. Please let me know if I am off base, but I think that the assertion was too tight. 

In src/storage/ducklake_catalog.cpp lines 439-461, the `table_entry_map` can have views added to it. However, there was an assertion in `DuckLakeCatalogSet::GetEntryById(TableIndex index)` that required a table (and not a view). Allowing a view there appears to solve things.

Alex-Monahan:
As I am thinking more about this, I'm wondering if the updates I made to the catalog cache would be safe across processes. Is it safe to not update the schema_version? Could other processes use a stale sort order, or reset it back to being unsorted?

pdet:
> So, to fix the assertion issues on my fork's CI, I had to relax an assertion. Please let me know if I am off base, but I think that the assertion was too tight.
> 
> In src/storage/ducklake_catalog.cpp lines 439-461, the `table_entry_map` can have views added to it. However, there was an assertion in `DuckLakeCatalogSet::GetEntryById(TableIndex index)` that required a table (and not a view). Allowing a view there appears to solve things.

Could you point me out to which test broke this requirement?

From what I can tell, there are other parts of the code that require this to return a table, as we deference it to a `DuckLakeTableEntry`

 e.g.,
```
unique_ptr<DuckLakeStats> DuckLakeCatalog::ConstructStatsMap(vector<DuckLakeGlobalStatsInfo> &global_stats,
                                                             DuckLakeCatalogSet &schema) {
	auto lake_stats = make_uniq<DuckLakeStats>();
	for (auto &stats : global_stats) {
		// find the referenced table entry
		auto table_entry = schema.GetEntryById(stats.table_id);
```

pdet:
> As I am thinking more about this, I'm wondering if the updates I made to the catalog cache would be safe across processes. Is it safe to not update the schema_version? Could other processes use a stale sort order, or reset it back to being unsorted?

Maybe this is something we can test with concurrentloops? (see `test/sql/snapshot_info/ducklake_last_commit.test`)

Alex-Monahan:
> > So, to fix the assertion issues on my fork's CI, I had to relax an assertion. Please let me know if I am off base, but I think that the assertion was too tight.
> > In src/storage/ducklake_catalog.cpp lines 439-461, the `table_entry_map` can have views added to it. However, there was an assertion in `DuckLakeCatalogSet::GetEntryById(TableIndex index)` that required a table (and not a view). Allowing a view there appears to solve things.
> 
> Could you point me out to which test broke this requirement?
> 
> From what I can tell, there are other parts of the code that require this to return a table, as we deference it to a `DuckLakeTableEntry`
> 
> e.g.,
> 
> ```
> unique_ptr<DuckLakeStats> DuckLakeCatalog::ConstructStatsMap(vector<DuckLakeGlobalStatsInfo> &global_stats,
>                                                              DuckLakeCatalogSet &schema) {
> 	auto lake_stats = make_uniq<DuckLakeStats>();
> 	for (auto &stats : global_stats) {
> 		// find the referenced table entry
> 		auto table_entry = schema.GetEntryById(stats.table_id);
> ```

Sure! The tests that broke are [here in this CI run on my fork](https://github.com/Alex-Monahan/ducklake/actions/runs/21047728480/job/60526337029). They were all running a query like
```sql
COMMENT ON VIEW ducklake.comment_view IS 'con1';
```

I can create a view-specific GetEntryById function if that would be better!

Alex-Monahan:
> > As I am thinking more about this, I'm wondering if the updates I made to the catalog cache would be safe across processes. Is it safe to not update the schema_version? Could other processes use a stale sort order, or reset it back to being unsorted?
> 
> Maybe this is something we can test with concurrentloops? (see `test/sql/snapshot_info/ducklake_last_commit.test`)

Unfortunately, I believe that `concurrentloop` uses the same DuckDB instance with multiple threads, and the DuckLake catalog is only created once per `ATTACH`, so it is shared across all threads. I think the issue that might exist from no longer incrementing schema_version would be when two totally separate DuckLakeCatalog instances have different sort information in their cache (with the same schema_version). Is there a multi-process version of concurrentloop? Or maybe a C++ or Python test? Do you want me to remove the schema_version modifications and save them for a later PR?

pdet:
> > > As I am thinking more about this, I'm wondering if the updates I made to the catalog cache would be safe across processes. Is it safe to not update the schema_version? Could other processes use a stale sort order, or reset it back to being unsorted?
> > 
> > 
> > Maybe this is something we can test with concurrentloops? (see `test/sql/snapshot_info/ducklake_last_commit.test`)
> 
> Unfortunately, I believe that `concurrentloop` uses the same DuckDB instance with multiple threads, and the DuckLake catalog is only created once per `ATTACH`, so it is shared across all threads. I think the issue that might exist from no longer incrementing schema_version would be when two totally separate DuckLakeCatalog instances have different sort information in their cache (with the same schema_version). Is there a multi-process version of concurrentloop? Or maybe a C++ or Python test? Do you want me to remove the schema_version modifications and save them for a later PR?

Could we achieve this with multiple connections then? Because that's also possible within sqltests

Alex-Monahan:
> > > > As I am thinking more about this, I'm wondering if the updates I made to the catalog cache would be safe across processes. Is it safe to not update the schema_version? Could other processes use a stale sort order, or reset it back to being unsorted?
> > > 
> > > 
> > > Maybe this is something we can test with concurrentloops? (see `test/sql/snapshot_info/ducklake_last_commit.test`)
> > 
> > 
> > Unfortunately, I believe that `concurrentloop` uses the same DuckDB instance with multiple threads, and the DuckLake catalog is only created once per `ATTACH`, so it is shared across all threads. I think the issue that might exist from no longer incrementing schema_version would be when two totally separate DuckLakeCatalog instances have different sort information in their cache (with the same schema_version). Is there a multi-process version of concurrentloop? Or maybe a C++ or Python test? Do you want me to remove the schema_version modifications and save them for a later PR?
> 
> Could we achieve this with multiple connections then? Because that's also possible within sqltests

I am not sure! If you have a spot where I can find an example, I can give it a shot.

To understand the behavior, I made a Python script that kicks off 2 separate CLI processes. I found that the sort is ignored by the other process if the schema was already cached ahead of time. The good news is that the catalog DB itself continues to have the right values, but the cache does not get invalidated correctly. 

The flow is:
- Process 1: connects, creates the table and inserts into it
- Process 2: connects and runs an `ALTER TABLE ADD COLUMN`, which caches the catalog
- Process 1: `ALTER TABLE SET SORTED BY`
- Process 1: Completes / exits
- Process 2: Compacts (using the cached catalog)
- Process 2: Pulls updated table (which does not show the right order, since the cached catalog was used)
- Process 2: Completes / exits

If I omit the `ALTER TABLE ADD COLUMN` step in process 2, then there is no issue and the sort occurs correctly. 

What do you recommend I do? I've thought about 3 options, but 3 would need some help!
1. Keep the schema_version from incrementing, but accept this concurrency behavior.
2. Allow the schema_version to increment but accept that a compaction barrier gets put in when sort is changed
3. Keep the schema_version from incrementing but find some other way to correctly invalidate the cache or use a different key for the cache

[ducklake_set_sorted_multiprocess_add_column.py](https://github.com/user-attachments/files/24726428/ducklake_set_sorted_multiprocess_add_column.py)

The logs get printed out for process 1 before printing out process 2, but I logged out some timestamps to show the true order:

```sql
uv run ./test/ducklake_set_sorted_multiprocess_add_column.py
┌─────────┬──────────────────────┐
│ process │     sum("range")     │
│ varchar │        int128        │
├─────────┼──────────────────────┤
│ sql_1   │ 12499999997500000000 │
└─────────┴──────────────────────┘
┌─────────┬─────────────────────────────────────────────────────────────┐
│ process │ (CAST(now() AS VARCHAR) || ' sql_1 finished SET SORTED BY') │
│ varchar │                           varchar                           │
├─────────┼─────────────────────────────────────────────────────────────┤
│ sql_1   │ 2026-01-19 19:16:39.22008-07 sql_1 finished SET SORTED BY   │
└─────────┴─────────────────────────────────────────────────────────────┘


┌─────────┬─────────────────────┐
│ process │    sum("range")     │
│ varchar │       int128        │
├─────────┼─────────────────────┤
│ sql_2   │ 1999999999000000000 │
└─────────┴─────────────────────┘
┌─────────┬────────────────────────────────────────────────────────────────────┐
│ process │ (CAST(now() AS VARCHAR) || ' sql_2 finished adding column') │
│ varchar │                              varchar                               │
├─────────┼────────────────────────────────────────────────────────────────────┤
│ sql_2   │ 2026-01-19 19:16:35.416129-07 sql_2 finished adding column  │
└─────────┴────────────────────────────────────────────────────────────────────┘
┌─────────┬──────────────────────┐
│ process │     sum("range")     │
│ varchar │        int128        │
├─────────┼──────────────────────┤
│ sql_2   │ 40499999995500000000 │
└─────────┴──────────────────────┘
┌─────────┬───────────────────────────────────────────────────────┐
│ process │ (CAST(now() AS VARCHAR) || ' sql_2 about to compact') │
│ varchar │                        varchar                        │
├─────────┼───────────────────────────────────────────────────────┤
│ sql_2   │ 2026-01-19 19:16:46.373997-07 sql_2 about to compact  │
└─────────┴───────────────────────────────────────────────────────┘
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│ 0 rows  │
└─────────┘
┌─────────┬───────────┬────────────┬────────────┬─────────┐
│ process │ unique_id │ sort_key_1 │ sort_key_2 │  bonus  │
│ varchar │   int64   │   int64    │  varchar   │ varchar │
├─────────┼───────────┼────────────┼────────────┼─────────┤
│ sql_2   │         3 │          1 │ woot3      │ NULL    │
│ sql_2   │         2 │          0 │ woot2      │ NULL    │
│ sql_2   │         1 │          1 │ woot1      │ NULL    │
│ sql_2   │         0 │          0 │ woot0      │ NULL    │
│ sql_2   │         7 │          1 │ woot7      │ NULL    │
│ sql_2   │         6 │          0 │ woot6      │ NULL    │
│ sql_2   │         5 │          1 │ woot5      │ NULL    │
│ sql_2   │         4 │          0 │ woot4      │ NULL    │
└─────────┴───────────┴────────────┴────────────┴─────────┘
┌─────────┬─────────────┬────────────────┬────────────────────────────────────────────┐
│ process │ snapshot_id │ schema_version │                  changes                   │
│ varchar │    int64    │     int64      │          map(varchar, varchar[])           │
├─────────┼─────────────┼────────────────┼────────────────────────────────────────────┤
│ sql_2   │           0 │              0 │ {schemas_created=[main]}                   │
│ sql_2   │           1 │              1 │ {tables_created=[main.sort_on_compaction]} │
│ sql_2   │           2 │              1 │ {tables_inserted_into=[1]}                 │
│ sql_2   │           3 │              1 │ {tables_inserted_into=[1]}                 │
│ sql_2   │           4 │              2 │ {tables_altered=[1]}                       │
│ sql_2   │           5 │              2 │ {tables_altered=[1]}                       │
│ sql_2   │           6 │              2 │ {}                                         │
└─────────┴─────────────┴────────────────┴────────────────────────────────────────────┘
┌─────────┬──────────┬────────────────┬──────────────┬────────────────┬────────────┬────────────────┬────────────┐
│ process │ table_id │ begin_snapshot │ end_snapshot │ sort_key_index │ expression │ sort_direction │ null_order │
│ varchar │  int64   │     int64      │    int64     │     int64      │  varchar   │    varchar     │  varchar   │
├─────────┼──────────┼────────────────┼──────────────┼────────────────┼────────────┼────────────────┼────────────┤
│ sql_2   │        1 │              5 │         NULL │              0 │ sort_key_1 │ DESC           │ NULLS_LAST │
│ sql_2   │        1 │              5 │         NULL │              1 │ sort_key_2 │ DESC           │ NULLS_LAST │
└─────────┴──────────┴────────────────┴──────────────┴────────────────┴────────────┴────────────────┴────────────┘


```

src/include/storage/ducklake_sort_data.hpp
```diff
@@ -0,0 +1,31 @@
...
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/common.hpp"
+#include "common/index.hpp"
+#include "duckdb/common/enums/order_type.hpp"
+
+namespace duckdb {
+class BaseStatistics;
```
pdet: Is this necessary?

pdet:
Hi @Alex-Monahan I had another pass on your PR, and it is looking great! I think there was a slight miscommunication issue wrt the snapshot changes. What I meant is that Sort/Comment, etc should not impact the `ducklake_schema_versions` table because that is used to ensure compaction of tables that have the same data-schema as in columns being the same in amount and type.

Alex-Monahan:
> Hi @Alex-Monahan I had another pass on your PR, and it is looking great! I think there was a slight miscommunication issue wrt the snapshot changes. What I meant is that Sort/Comment, etc should not impact the `ducklake_schema_versions` table because that is used to ensure compaction of tables that have the same data-schema as in columns being the same in amount and type.

Hmm, so it is ok if they increment the schema_version inside of `ducklake_snapshot`, just not `ducklake_schema_versions`? I'm having trouble detangling how to only prevent updating the schema_version selectively.

pdet:
> > Hi @Alex-Monahan I had another pass on your PR, and it is looking great! I think there was a slight miscommunication issue wrt the snapshot changes. What I meant is that Sort/Comment, etc should not impact the `ducklake_schema_versions` table because that is used to ensure compaction of tables that have the same data-schema as in columns being the same in amount and type.
> 
> Hmm, so it is ok if they increment the schema_version inside of `ducklake_snapshot`, just not `ducklake_schema_versions`? I'm having trouble detangling how to only prevent updating the schema_version selectively.

I think it's fine that they increase the global counter of the schema_version, I'm really sorry for the confusion and added work!

Alex-Monahan:
@pdet, I believe I finally understood your advice! I am now incrementing the global schema_version, but not incrementing the ducklake_schema_versions table if the only change was a comment or a set sorted. 

My Python test script works correctly now, and I updated the tests to focus on the ducklake_schema_versions table.

I believe this is ready for another review (but CI is still running on my fork).  Thank you!!

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -203,6 +204,39 @@ struct DuckLakePartitionInfo {
 	}
 };
 
+struct DuckLakeSortFieldInfo {
+	idx_t sort_key_index = 0;
+	// TODO: Validate that expression is case insensitive when stored
+	string expression;
```
pdet, line 209: We should handle the case insensitiveness for the column names in this PR already I think, and a test

```
CREATE TABLE t (MyColumn INT, AnotherCol VARCHAR);

ALTER TABLE t SET SORTED BY (mycolumn ASC);

ALTER TABLE t SET SORTED BY (MyColumn ASC);
```

Alex-Monahan:
Ok, the previously failing tests on my CI are now green! All I had to change were the tests: added some `ORDER BY`'s for SQLite and used unique folder names so that I get accurate file counts. 

Sorry for being optimistic about CI being a straight shot...

Alex-Monahan:
> Hey Alex, thanks again for all the great work! I think this is pretty much ready, just had a couple of last comments.
> 
> One thing i'm also wondering is, what happens if we set the exact same order twice in a row?
> 
> E.g.,
> 
> ```sql
> CREATE TABLE ducklake.renamed_columns_test (unique_id INTEGER, sort_key_1 INTEGER, sort_key_2 VARCHAR);
> 
> 
> ALTER TABLE ducklake.renamed_columns_test SET SORTED BY (sort_key_1 ASC NULLS LAST, sort_key_2 ASC NULLS LAST)
> 
> -- should this error? do nothing? add a new entry?
> ALTER TABLE ducklake.renamed_columns_test SET SORTED BY (sort_key_1 ASC NULLS LAST, sort_key_2 ASC NULLS LAST).
> 
> Can you also add in the description the schema and a brief comment of the tables you added? That will make @guillesd work easier for the docs!
> ```

Thanks for the review! I [have a test for that over in `test/sql/sorted_table/merge_adjacent_sorted_repeated.test`](https://github.com/duckdb/ducklake/pull/642/files#diff-e6a2940db4fa03b9a12cee206a17c6cf6160b6baa6790470d32632ab8d923c3dR19)! I chose the "do nothing" path. I do a [deduplication check here](https://github.com/duckdb/ducklake/pull/642/files#diff-47ab182e78054c014fa65d03724bf2581c037ef5812e0ce59d19f431696058b9R2726) so that we don't add redundant entries into the catalog. Let me know if that is ok!

I'm happy to update the description, and @guillesd, let me know if I can be helpful on the docs!

guillesd:
Hey @Alex-Monahan if you can provide an **EDIT** in the PR description just detailing a bit your final implementation details, and adding the syntax (and options if applicable). That would be great. I tagged the PR so an issue should be now linked in the ducklake docs page.

src/include/storage/ducklake_sort_data.hpp
Alex-Monahan: It was not necessary! Removed it!

src/include/storage/ducklake_metadata_info.hpp
Alex-Monahan: Yes, that is handled already! I added some tests and removed that outdated ToDo.

Alex-Monahan:
> Hey @Alex-Monahan if you can provide an **EDIT** in the PR description just detailing a bit your final implementation details, and adding the syntax (and options if applicable). That would be great. I tagged the PR so an issue should be now linked in the ducklake docs page.

Thank you! I've updated the description! No new options, but 2 new DuckLake spec tables.

Alex-Monahan:
It looks like CI failed on the Docker Build step. Mind giving it a re-run? [CI runs smoothly on my fork now!](https://github.com/Alex-Monahan/ducklake/actions)

pdet:
Thanks!

redox:
Great job @Alex-Monahan @pdet - an impressive amount of work we've been following from the sidelines!

Issue #653, opened 2025-12-27
MinIO tests might run on local disk


utay:
### What happens?

I noticed that `./build/release/test/unittest --test-config test/configs/minio.json` doesn't seem to apply the `test_env` variables to the tests. 

After some digging, it might be a regression from https://github.com/duckdb/duckdb/commit/e7b1a6d55b8571ab4846c895f033f23f1e6b114d where `UpdateEnvironment` populates `test_env` before the config file is read.

Here's a possible fix:
```patch
diff --git a/test/helpers/test_config.cpp b/test/helpers/test_config.cpp
index f1d8f8142c..ed674a6a02 100644
--- a/test/helpers/test_config.cpp
+++ b/test/helpers/test_config.cpp
@@ -467,7 +467,8 @@ vector<ConfigSetting> TestConfiguration::GetConfigSettings() {
 }

 string TestConfiguration::GetTestEnv(const string &key, const string &default_value) {
-       if (test_env.empty() && options.find("test_env") != options.end()) {
+       if (!test_env_from_config_loaded && options.find("test_env") != options.end()) {
+               test_env_from_config_loaded = true;
                auto entry = options["test_env"];
                auto list_children = ListValue::GetChildren(entry);
                for (const auto &value : list_children) {
diff --git a/test/include/test_config.hpp b/test/include/test_config.hpp
index aa4e8246f5..ade00efbaa 100644
--- a/test/include/test_config.hpp
+++ b/test/include/test_config.hpp
@@ -103,6 +103,7 @@ private:
        string working_dir;
        string test_uuid;
        unordered_map<string, string> test_env;
+       bool test_env_from_config_loaded = false;

        vector<unordered_set<string>> select_tag_sets;
        vector<unordered_set<string>> skip_tag_sets;
``` 

While this might be a bug in the duckdb codebase, I opened an issue in this repo because duckdb doesn't use `test_env` in any of its test config files.

Apologies in advance if this isn’t actually a bug and I’m just missing something in my local test setup!

### To Reproduce

I could reproduce with the following test file:
```
# name: test/sql/repro.test
# description: XXX
# group: [sql]

require ducklake

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

test-env DATA_PATH __TEST_DIR__

query I
SELECT '${DATA_PATH}';
----
s3://mybucket
```

Then run:
```
$ ./build/release/test/unittest --test-config test/configs/minio.json "test/sql/repro.test"
[...]

PR #655, opened 2025-12-28
Use RemoveFiles (bulk delete) to clean up old files

utay:
This PR is the final change to address https://github.com/duckdb/ducklake/discussions/384: take advantage of S3 bulk deletion instead of doing one delete request per file.

It requires these 2 PRs to be merged before:
- https://github.com/duckdb/duckdb/pull/20333
- https://github.com/duckdb/duckdb-httpfs/pull/181

The test `test/sql/cleanup/cleanup_old_files.test` already covers this change when run with `--test-config test/configs/minio.json` (btw I think I spotted a bug with the test configs, see https://github.com/duckdb/ducklake/issues/653).

### DIFF
--- a/src/functions/ducklake_cleanup_files.cpp
+++ b/src/functions/ducklake_cleanup_files.cpp
@@ -131,9 +131,12 @@ void DuckLakeCleanupExecute(ClientContext &context, TableFunctionInput &data_p, 
 	if (!state.executed && !data.dry_run) {
 		// delete the files
 		auto &fs = FileSystem::GetFileSystem(context);
-		for (auto &file : data.files) {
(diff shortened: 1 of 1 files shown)


utay:
Hi @pdet, I think this should be mergeable since https://github.com/duckdb/duckdb/pull/20333 has landed on duckdb `main`. For S3, it should have the same behavior (as `RemoveFiles` uses `TryRemoveFile` in a loop by default) until https://github.com/duckdb/duckdb-httpfs/pull/181 gets merged.

pdet:
Sure, LGTM! Thanks for all the work across multiple repos, highly appreciated!

PR #660, opened 2025-12-31
Fix column renaming to not update end_snapshot on historical entries

utay:
We got this error - after renaming a column - when running compaction:
```
D call <catalog>.merge_adjacent_files();
Catalog Error:
Column with name user_id already exists!
```

It appears to be a bug, we wouldn’t expect schema evolution to affect compaction in this way.

This fix and test were implemented by Opus 4.5 after I provided some context. It appears that the `WriteDroppedColumns` function was incorrectly updating `end_snapshot` for all historical entries sharing the same `column_id`, rather than only the currently active one.

I also asked the LLM how to recover from the metadata corruption; it suggested the following query, which seems reasonable to me and unblocked us (⚠️ use at your own risk - I’m not 100% confident it’s correct):
<details>
<summary>Recovery Query</summary>

```sql
-- First, check the corrupted entries
WITH column_history AS (
    SELECT 
        table_id, column_id, column_name, begin_snapshot, end_snapshot,
        LEAD(begin_snapshot) OVER (PARTITION BY table_id, column_id ORDER BY begin_snapshot) AS next_begin
    FROM __ducklake_metadata_altertable.ducklake_column
    WHERE end_snapshot IS NOT NULL
)
SELECT * FROM column_history
WHERE end_snapshot != next_begin
ORDER BY table_id, column_id, begin_snapshot;

-- Then fix them
WITH corrections AS (
    SELECT 
        table_id, column_id, begin_snapshot,
        LEAD(begin_snapshot) OVER (PARTITION BY table_id, column_id ORDER BY begin_snapshot) AS correct_end
    FROM __ducklake_metadata_altertable.ducklake_column
)
UPDATE __ducklake_metadata_altertable.ducklake_column AS c
SET end_snapshot = corrections.correct_end
FROM corrections
WHERE c.table_id = corrections.table_id 
  AND c.column_id = corrections.column_id 
  AND c.begin_snapshot = corrections.begin_snapshot
  AND c.end_snapshot IS NOT NULL
  AND corrections.correct_end IS NOT NULL
  AND c.end_snapshot != corrections.correct_end;
```
</details>

### DIFF
--- a/test/sql/compaction/compaction_multiple_rename_column.test
+++ b/test/sql/compaction/compaction_multiple_rename_column.test
@@ -14,10 +14,12 @@ require ducklake
 
 require parquet
 
-load __TEST_DIR__/compaction_multi_rename.db
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/compaction_multiple_rename_column.db
+
+test-env DATA_PATH __TEST_DIR__
 
 statement ok
-ATTACH 'ducklake:__TEST_DIR__/compaction_multi_rename_meta.db' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_compaction_multi_rename_files')
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/compaction_multiple_rename_column')
 
 # Create table with user_id column (schema version 1)
 statement ok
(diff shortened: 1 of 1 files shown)


test/sql/compaction/compaction_multiple_rename_column.test
```diff
@@ -0,0 +1,69 @@
...
+# Bug scenario: When renaming a column multiple times (A -> B -> A -> C), the last rename
+# would corrupt all historical entries for that column_id, making multiple entries visible
+# at intermediate snapshots. This caused "Column with name X already exists!" errors when
+# loading old schema versions (e.g., during compaction or time travel).
+
+require ducklake
+
+require parquet
+
+load __TEST_DIR__/compaction_multi_rename.db
```
pdet: Do we really need a load here?

test/sql/compaction/compaction_multiple_rename_column.test
```diff
@@ -0,0 +1,69 @@
...
+
+statement ok
+ATTACH 'ducklake:__TEST_DIR__/compaction_multi_rename_meta.db' AS ducklake (DATA_PATH '__TEST_DIR__/ducklake_compaction_multi_rename_files')
```
pdet: Replace with, as this will allows us to run this test over multiple different configurations in our CI:

```
test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/compaction_multiple_rename_column/', METADATA_CATALOG 'metadata');

statement ok
USE ducklake;

```

test/sql/compaction/compaction_multiple_rename_column.test
utay: No, I removed these `load`

test/sql/compaction/compaction_multiple_rename_column.test
utay: Of course! Fixed.

I assigned `DUCKLAKE_CONNECTION` to `__TEST_DIR__/compaction_multiple_rename_column.db` instead of `__TEST_DIR__/{UUID}.db`. Otherwise, it seems like it was using different UUIDs for the `ATTACH`s and it wasn't reusing the same database file.

pdet:
Thanks! LGTM!

PR #668, opened 2026-01-05
Implement Top-N Dynamic Filter File Pruning from Min/Max Column Statistics

redox:
I happen to query our append-only dataset with `SELECT * FROM my_ducklake.main.events ORDER BY timestamp DESC LIMIT 10` and quickly experienced quite some performance issues because no file pruning was done: so it ends up sorting all rows from all files to get the top 10. If I'm not mistaken, in such case we should be able to skip quite some data files that don't contain relevant data based on their min/max values.

Here is what I've done:
 - Updated `DuckLakeMetadataManager::GetFilesForTable()` to detect Top-N dynamic filters and retrieve their associated `min_value` & `max_value` to later on sort the candidate files accordingly.
    - For `ORDER BY col DESC LIMIT N`: Files with highest max values are processed first
    - For `ORDER BY col ASC LIMIT N`: Files with lowest min values are processed first
 - Updated `DuckLakeMultiFileReader::InitializeReader()` to actually prune the files when their min/max ranges indicate they cannot contain qualifying rows
 - Added a test validating this with `EXPLAIN ANALYZE` diffs

On my own tests, this is a tremendous performance improvement; but I would love the expert's POV to make sure I don't forget anything 😅 

Let me know what y'all think!


redox:
Gosh, I need to handle the `NULLS FIRST/LAST` as well

redox:
> Gosh, I need to handle the `NULLS FIRST/LAST` as well

Oh ok, the TopN optimizer for `NULLS FIRST/LAST` is [disabled in 1.4](https://github.com/duckdb/duckdb/blob/v1.4-andium/src/optimizer/topn_optimizer.cpp#L64) 😅 but seems to be implemented in main with the new `ConjunctionOrFilter` (I'll check what it takes to support it)

redox:
> Oh ok, the TopN optimizer for `NULLS FIRST/LAST` is [disabled in 1.4](https://github.com/duckdb/duckdb/blob/v1.4-andium/src/optimizer/topn_optimizer.cpp#L64) 😅 but seems to be implemented in main with the new `ConjunctionOrFilter` (I'll check what it takes to support it)

I dig a bit and I'm struggling with an uninitialized `DynamicFilterData`. @Mytherin does this ring a bell? I read https://github.com/duckdb/duckdb/commit/f85397a208ab8b7abf769e47c907423d4c0f8ef6 but it's not clear to me why `setValue` has this specific handling of `NULL` - which I believe has an impact here because I end up dealing with a non-initialized filter.

redox:
@pdet I'm curious to hear your expert's thoughts on this one 🙏

pdet:
Hi @redox, I'm sorry is a bit of a busy week, I'll have a look in the next couple of days!

redox:
> Hi @redox, I'm sorry is a bit of a busy week, I'll have a look in the next couple of days!

Take your time, I know those weeks! Just wanted to make sure it was on your radar 👌

pdet:
Hi @redox I think this is looking good, one thing that I wanted to do before accepting new filtering pushdown PRs was too increase the filtering tests we have. As this would drastically reduce my anxiety of incorrectly skipping files, so maybe that's a nice thing to pick up in this PR.

One nice thing is that DuckDB already has a quite complete test load for filters that we could reuse, the difference of course is that for these tests to make a difference for ducklake, we would need to force them to write many parquet files.

My battle plan would be to add a CI run that stress tests the DuckLake Filter skipping (Somewhat similar to https://github.com/duckdb/ducklake/blob/main/.github/workflows/ConfigTests.yml). But enforcing that our parquet files are as small as possible, so we can stress-test the filter-skipping correctness of DuckLake. Hence a script to collect tests that are filter related would also be nice.

I think one way of  producing small parquet files is if we set `ROW_GROUP_SIZE 1`  ,and `ROW_GROUPS_PER_FILE 1`  in the parquet writer, we should limit the parquet files to a vector size. By default the vector size is 2048. The problem is that this might cause overshooting, but I think that setting `ROW_GROUPS_PER_FILE` 1 should resolve that, or worst case, `set threads = 1`.

Many tests are use very small datasets, having the vector_size limit to 2048 could still leave things untested, so I would probably try to do a DuckDB build with a vector size of 2. Then we could get parquet files with 2 rows. If this CI is too heavy, we can also limit the filters to one code/header file and only run it if these are changed.

Let me know if this makes sense.

redox:
> Let me know if this makes sense.

Yes this makes a lot of sense, I'll check how easily I could add more filtering tests; following your guidance.

redox:
@pdet without going the CI run yet (that maybe you should own - always a bit tricky for non-maintainer to iterate of those), I've added a [test file](https://github.com/duckdb/ducklake/pull/668/commits/a0895ac071cfb717a5ebdd816e4e3420f62ff036) that I believe is stressing quite a bit the filtering capabilities.

 - It builds a reference table `events_ref` in DuckDB with a mix of types
 - It then writes the reference data into many Parquet files `(ROW_GROUP_SIZE 2048, ROW_GROUPS_PER_FILE 1)` (I initially made `ROW_GROUP_SIZE` smaller but I think it's giving nicer results without)
 - It registers those files into a DuckLake `events` table via `ducklake_add_data_files(...)`
 - For a bunch of queries inspired by the DuckDB tests; it asserts DuckLake's `events` returns exactly the same rows as` events_ref` by using `EXCEPT ALL` in both directions (so it catches both missing rows and extra rows)

Let me know how you feel about it!

redox:
@pdet I've targeted `main` now that you restrict PR targeting 1.4 to only bug fixes - I've also documented that the optimization won't apply on NULL FIRSTS ordering (could be something we add in another PR).

pdet:
Hi @redox, I had time to take another look at your PR. Sorry it took so long.

We should integrate with the current filter pushdown system, and we should not use `stable_sort`. I think your comparator calls `Value::DefaultCastAs`, which can throw, and that can cause the container to end up in a state where the order is not well defined.

Can you integrate this with the CTE-based filter pushdown structure to handle the dynamic filters? Then we can use the same architecture for all filter pushdown and avoid the exception issues, by using a try_cast. We could also run this directly in SQL and let the optimizer handle the query.

redox:
> Hi @redox, I had time to take another look at your PR. Sorry it took so long.

No worries @pdet!

> We should integrate with the current filter pushdown system, and we should not use `stable_sort`. I think your comparator calls `Value::DefaultCastAs`, which can throw, and that can cause the container to end up in a state where the order is not well defined.
> 
> Can you integrate this with the CTE-based filter pushdown structure to handle the dynamic filters? Then we can use the same architecture for all filter pushdown and avoid the exception issues, by using a try_cast. We could also run this directly in SQL and let the optimizer handle the query.

Ah yes, I didn't realize it could be done there directly; I think I understood your guidance and applied your feedback, moved the sort to the SQL query along with being more gentle with the casts.

Let me know if that sounds legit now! At least, all tests are still passing 😅

pdet:
Jup, these changes LGTM! Thanks for all the work, I think we are just missing a merge? The branch is showing me conflicts right now, then I'll merge when CI passes :-)

redox:
> I think we are just missing a merge?

Yes, I will work on this now - I have a failing test; will report back 🫡

redox:
> I have a failing test; will report back 🫡

@pdet the test is failing on `main` as well actually (on my m4 pro); @rgernhardt does this ring a bell its'a new test you added in https://github.com/duckdb/ducklake/pull/734 ?

```
1. test/sql/compaction/merge_adjacent_max_files.test:91
Wrong result in query! (test/sql/compaction/merge_adjacent_max_files.test:91)!
SELECT table_name, files_processed, files_created FROM ducklake_merge_adjacent_files('ducklake', 'example', max_compacted_files=>2) ORDER BY ALL;
Mismatch on row 1, column files_processed(index 2)
6 <> 5
Expected result:
example 5       1
example 5       1

Actual result:
example 6       1
example 6       1
```

redox:
Thank you @pdet - excited about this!

rgernhardt:
> > I have a failing test; will report back 🫡
> 
> @pdet the test is failing on `main` as well actually (on my m4 pro); @rgernhardt does this ring a bell its'a new test you added in #734 ?
> 
> ```
> 1. test/sql/compaction/merge_adjacent_max_files.test:91
> ================================================================================
> Wrong result in query! (test/sql/compaction/merge_adjacent_max_files.test:91)!
> ================================================================================
> SELECT table_name, files_processed, files_created FROM ducklake_merge_adjacent_files('ducklake', 'example', max_compacted_files=>2) ORDER BY ALL;
> ================================================================================
> Mismatch on row 1, column files_processed(index 2)
> 6 <> 5
> ================================================================================
> Expected result:
> ================================================================================
> example 5       1
> example 5       1
> 
> ================================================================================
> Actual result:
> ================================================================================
> example 6       1
> example 6       1
> ```

@redox 
I apologize, I didn't have notifications enabled and missed this mention. I have a fix for that test out here: https://github.com/duckdb/ducklake/pull/802

PR #671, opened 2026-01-06
Change schema tracking from global to table.

pdet:
This PR changes we track snapshot ranges for schema changes, instead of tracking them globally, we do it per table.

The `ducklake_schema_versions` table gets a new `table_id` column, and queries are adjusted accordingly.

This change maximizes our compaction capabilites over scenarios with extreme schema changes.


pdet:
Tests are failing because of https://github.com/duckdb/duckdb/pull/20345

Will have a better look tomorrow.

PR #674, opened 2026-01-07
Add custom MultiFileReader for reading delete files

redox:
This follows https://github.com/duckdb/ducklake/pull/641 that we closed to work-around a GH issue.

We noticed that read performance can degrade when a table accumulates a large number of delete files, especially when those files live on remote object storage.

**TLDR;** We can actually speed things up by reusing metadata that DuckLake already has, avoiding a bunch of unnecessary storage `HEAD` requests when loading delete files. 

For data files, this metadata is passed down to DuckDB’s `MultiFileReader` and ultimately to the filesystem layer. This matters because the Parquet reader needs the file size to locate footer metadata efficiently. However, for delete files, this optimization wasn’t applied yet. When using an `HTTPFileSystem` (e.g. S3 or Azure), missing metadata means DuckDB issues a `HEAD` request per file to fetch it. Since DuckLake files are immutable, those extra requests are pure overhead—and they really add up when a table has many delete files.

This PR introduces a custom multi-file reading path for delete files that eagerly injects the metadata we already have, eliminating redundant storage calls.

### What Changed  

1. **New `DeleteFileFunctionInfo` struct**  
   Extends `TableFunctionInfo` to carry `DuckLakeFileData` metadata through the table function binding phase.

2. **Custom `DeleteFileMultiFileReader`**  
   A specialized `MultiFileReader` that:
   - Intercepts file list creation
   - Pre-populates `ExtendedOpenFileInfo` with known metadata:
     - File size (`file_size_bytes`)
     - ETag (empty placeholder)
     - Last modified timestamp (epoch)
     - Encryption key (when present)
   - Builds a `SimpleMultiFileList` upfront
   - Overrides `CreateFileList()` to return this pre-built list, skipping DuckDB’s default discovery logic entirely

3. **Updates to `ScanDeleteFile()`**  
   - Switched `parquet_scan` from a const reference to a mutable copy so it can be customized
   - Attaches `DeleteFileFunctionInfo` and the custom reader factory to the table function
   - Passes the real `parquet_scan` function into `TableFunctionBindInput` (instead of a dummy), ensuring the correct execution context

The win is most noticeable on remote storage backends (S3, Azure, etc.) and for tables with many delete files, where these requests were a significant source of latency.

Net result: **faster reads, fewer network round-trips, and better scalability** ✨

### DIFF
--- a/src/storage/ducklake_delete_filter.cpp
+++ b/src/storage/ducklake_delete_filter.cpp
@@ -110,12 +110,11 @@ idx_t DuckLakeDeleteFilter::Filter(row_t start_row_index, idx_t count, Selection
 DeleteFileScanResult DuckLakeDeleteFilter::ScanDeleteFile(ClientContext &context, const DuckLakeFileData &delete_file,
                                                           optional_idx snapshot_filter_min,
                                                           optional_idx snapshot_filter_max) {
-	ParquetFileScanner scanner(context, delete_file);
-
 	// Set up custom MultiFileReader to avoid HEAD requests
 	auto function_info = make_shared_ptr<DeleteFileFunctionInfo>();
 	function_info->file_data = delete_file;
-	scanner.SetMultiFileReaderCreator(DeleteFileMultiFileReader::CreateInstance, std::move(function_info));
+	ParquetFileScanner scanner(context, delete_file, DeleteFileMultiFileReader::CreateInstance,
+	                           std::move(function_info));
 
 	auto &return_types = scanner.GetTypes();
 	auto &return_names = scanner.GetNames();
--- a/src/common/parquet_file_scanner.cpp
+++ b/src/common/parquet_file_scanner.cpp
@@ -6,7 +6,13 @@
 namespace duckdb {
 
 ParquetFileScanner::ParquetFileScanner(ClientContext &context, const DuckLakeFileData &file)
-    : context(context), multi_file_reader_creator(nullptr) {
+    : ParquetFileScanner(context, file, nullptr, nullptr) {
+}
+
+ParquetFileScanner::ParquetFileScanner(ClientContext &context, const DuckLakeFileData &file,
+                                       table_function_get_multi_file_reader_t multi_file_reader_creator_p,
+                                       shared_ptr<TableFunctionInfo> function_info_p)
+    : context(context) {
 	auto &instance = DatabaseInstance::GetDatabase(context);
 	ExtensionLoader loader(instance, "ducklake");
 	auto &parquet_scan_entry = loader.GetTableFunction("parquet_scan");
@@ -28,6 +34,14 @@ ParquetFileScanner::ParquetFileScanner(ClientContext &context, const DuckLakeFil
 	TableFunctionRef empty;
 	TableFunction dummy_table_function;
 	dummy_table_function.name = "ParquetFileScanner";
+
+	if (multi_file_reader_creator_p) {
+		dummy_table_function.get_multi_file_reader = multi_file_reader_creator_p;
+		if (function_info_p) {
+			dummy_table_function.function_info = std::move(function_info_p);
+		}
+	}
+
 	TableFunctionBindInput bind_input(children, named_params, input_types, input_names, nullptr, nullptr,
 	                                  dummy_table_function, empty);
 
@@ -59,12 +73,6 @@ void ParquetFileScanner::SetColumnIds(vector<column_t> column_ids_p) {
 	column_ids = std::move(column_ids_p);
 }
 
-void ParquetFileScanner::SetMultiFileReaderCreator(table_function_get_multi_file_reader_t creator,
-                                                   shared_ptr<TableFunctionInfo> function_info_p) {
-	multi_file_reader_creator = creator;
-	function_info = std::move(function_info_p);
-}
-
 void ParquetFileScanner::InitializeScan() {
 	if (initialized) {
 		return;
@@ -74,14 +82,6 @@ void ParquetFileScanner::InitializeScan() {
 	if (column_ids.empty()) {
 		for (idx_t i = 0; i < return_types.size(); i++) {
 			column_ids.push_back(i);
-		}
-	}
-
-	// Set custom multi-file reader creator if provided
-	if (multi_file_reader_creator) {
-		parquet_scan.get_multi_file_reader = multi_file_reader_creator;
-		if (function_info) {
-			parquet_scan.function_info = function_info;
 		}
 	}
 
--- a/src/include/common/parquet_file_scanner.hpp
+++ b/src/include/common/parquet_file_scanner.hpp
@@ -21,6 +21,9 @@ namespace duckdb {
 class ParquetFileScanner {
 public:
 	ParquetFileScanner(ClientContext &context, const DuckLakeFileData &file);
+	ParquetFileScanner(ClientContext &context, const DuckLakeFileData &file,
+	                   table_function_get_multi_file_reader_t multi_file_reader_creator,
+	                   shared_ptr<TableFunctionInfo> function_info = nullptr);
 
 	const vector<LogicalType> &GetTypes() const;
 	const vector<string> &GetNames() const;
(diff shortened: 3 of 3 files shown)


test/sql/delete/delete_metadata.test
```diff
@@ -0,0 +1,48 @@
...
+require-env S3_TEST_SERVER_AVAILABLE 1
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+require ducklake
+
+require parquet
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH 's3://mybucket')
```
pdet, line 16: Other small nit 

```

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/delete_metadata}')
```

This test will still only be executed with minio due to the require-env but could eventually be executed in dfferent scenarios in the future

test/sql/delete/delete_metadata.test
```diff
@@ -0,0 +1,48 @@
...
+query II
+EXPLAIN ANALYZE SELECT COUNT(*) FILTER(WHERE id%2=0) FROM ducklake.test
+----
+analyzed_plan	<REGEX>:.*#HEAD: 0.*
+
+# we can time travel to see the state of the table before deletes
+query II
+EXPLAIN ANALYZE SELECT COUNT(*) FILTER(WHERE id%2=0) FROM ducklake.test AT (VERSION => 2)
+----
+analyzed_plan	<REGEX>:.*#HEAD: 0.*
```
pdet, line 43: I've actually got this branch locally and did a build without your changes in `src/storage/ducklake_delete_filter.cpp` and the tests seems to have the same effect? Maybe I'm missing something?

src/storage/ducklake_delete_filter.cpp
```diff
@@ -1,11 +1,53 @@
...
+
+	explicit DeleteFileMultiFileReader(const TableFunction &table_function) {
+		auto &info = table_function.function_info->Cast<DeleteFileFunctionInfo>();
+		auto &delete_file = info.file_data;
+
+		OpenFileInfo file_info(delete_file.path);
+		auto extended_info = make_shared_ptr<ExtendedOpenFileInfo>();
+		extended_info->options["file_size"] = Value::UBIGINT(delete_file.file_size_bytes);
+		extended_info->options["etag"] = Value("");
+		extended_info->options["last_modified"] = Value::TIMESTAMP(timestamp_t(0));
```
pdet, line 28: @samansmink since you are more aware than I am of the effects of S3 files in the multifile reader, can this have any ill effects?

test/sql/delete/delete_metadata.test
redox: Hmm 🤔 I'll check this again on behalf of Mathieu who initially opened this MR. Thank you for this @pdet, I should have double-checked as well 🫡

redox:
Closing this for now; the bug is somewhere between https://github.com/duckdb/ducklake and https://github.com/dentiny/duck-read-cache-fs - we need to figure out how to test this from here if we want to fix it here.

test/sql/delete/delete_metadata.test
leo-altertable: @pdet Because of our local development setups we had a lot of trouble reproducing failure / success with and without the fix. That being said, it looks like the base commit we were using might have caused the flakiness you observed.

I went and created a Dockerfile (based on the Minio.yml workflow config to be as close as possible to the CI):

```Dockerfile
# Based on the MinIO.yml workflow configuration
FROM ubuntu:latest

# Prevent interactive prompts during package installation
ENV DEBIAN_FRONTEND=noninteractive

# Install required Ubuntu packages
RUN apt-get update -y -qq && \
    apt-get install -y -qq \
        software-properties-common \
        curl \
        git \
        zip \
        unzip \
        tar \
        pkg-config && \
    add-apt-repository ppa:git-core/ppa && \
    apt-get update -y -qq && \
    apt-get install -y -qq \
        build-essential \
        cmake \
        ninja-build \
        ccache \
        python3 \
        clang \
        llvm && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

# Set working directory
WORKDIR /workspace

# Copy the project
COPY . .

# Setup vcpkg
RUN git clone https://github.com/microsoft/vcpkg.git /workspace/vcpkg && \
    cd /workspace/vcpkg && \
    git checkout ce613c41372b23b1f51333815feb3edd87ef8a8b && \
    ./bootstrap-vcpkg.sh

# Set build environment variables
ENV VCPKG_TARGET_TRIPLET=x64-linux-release
ENV GEN=ninja
ENV VCPKG_FEATURE_FLAGS=-binarycaching
ENV VCPKG_TOOLCHAIN_PATH=/workspace/vcpkg/scripts/buildsystems/vcpkg.cmake
ENV BUILD_EXTENSION_TEST_DEPS=full
ENV S3_TEST_SERVER_AVAILABLE=1
ENV CC=clang
ENV CXX=clang++

# Show versions for verification
RUN ninja --version && \
    cmake --version && \
    clang --version

# Build the project
RUN make release

# Default command
CMD ["/bin/bash"]

```

Building with `docker build -t ducklake-build .`, and running the test with `docker run --network host -it ducklake-build bash -c "S3_TEST_SERVER_AVAILABLE=1 ./build/release/test/unittest --test-config test/configs/minio.json test/sql/delete/delete_metadata.test"` (having a minio instance running on localhost:9000) shows that:
* Without the `ducklake_delete_filter.cpp` fix of this PR, the test fails
* With the fix, the test succeeds.

test/sql/delete/delete_metadata.test
redox: Yes actually we cannot do this because of https://github.com/duckdb/duckdb/pull/20396 - once the 1.4 branch updates its dependency to a more recent duckdb 1.4 tag - we'll be able to do this.

pdet:
Aha, we actually have a very similar PR to iceberg, code-wise looks very similar:
https://github.com/duckdb/duckdb-iceberg/pull/579/files

Maybe the test we have there would be a better fit?

leo-altertable:
Hi @pdet!
I looked at the test in the PR you linked and slightly changed the test so it's more alike. Do you see anything else I should work on?

pdet:
> Hi @pdet! I looked at the test in the PR you linked and slightly changed the test so it's more alike. Do you see anything else I should work on?

No, I think that looks fine. Can you retarget it to `main`? I'm only doing critical bug fixes to v1.4 from now on. Many thanks,

leo-altertable:
@pdet I rebased and tried to use the new `ParquetFileScanner` you introduced.

That being said, I had to add a new method allowing to set an optional multi file reader creator. Hopefully it doesn't break the abstraction you had in mind.

pdet:
Hm, not sure why the CI is not running on this PR anymore. @carlopi any ideas?

carlopi:
@pdet: I have to say I am not really sure.

What I would try is one of:
 * 1. adding a commit of some kind on top of the PR (I think as repo maintainers can push stuff), it might be that PR trigger are not considering re-targeting (or draft status)
  * 2. run CI manually, via web UI in Action tab, then selecting branch
  
I would probably go for 1, seems simpler to me.

That said, it would be handy to uniform CI rules across repos.

leo-altertable:
Looks like pushing an empty commit indeed did the trick.

leo-altertable:
Looks like the CI failure is linked to a recent changed made to `main` in the duckdb repo, doesn't seem linked to the current PR.

pdet:
> Looks like the CI failure is linked to a recent changed made to `main` in the duckdb repo, doesn't seem linked to the current PR.

Jup, I'm applying the patches in this [PR](https://github.com/duckdb/ducklake/pull/730) as soon as that gets merged things should be stable again

pdet:
Hi `main`'s ci should be passing now, could you merge with it?

leo-altertable:
@pdet Thanks a lot for all the back and forth in this PR! Looks like everything is finally green!

pdet:
Thanks!

PR #676, opened 2026-01-07
Implement Bucket Transform Table Partitioning

Costa-SM:
I noticed that DuckLake doesn't have support for bucket partitioning and that, apart from me, other people were also interested in this feature ([discussion](https://github.com/duckdb/ducklake/discussions/670)), so I thought I'd give it a go. 
Apart from the test I implemented, I've been experimenting with it for the most part of today with a built-from-source version of DuckLake, and it seems to be working properly.

Please let me know what you guys think. :)

## Syntax
```sql
-- Create a table and set bucket partitioning
CREATE TABLE events(user_id VARCHAR, event_type VARCHAR, ts TIMESTAMP, value INTEGER);
ALTER TABLE events SET PARTITIONED BY (bucket(8, user_id));

-- You can combine with other partition transforms
ALTER TABLE events SET PARTITIONED BY (bucket(8, user_id), month(ts));
```


pdet:
Can you merge with `main`?

test/sql/partitioning/bucket_partitioning.test
```diff
@@ -0,0 +1,50 @@
+# name: test/sql/partitioning/bucket_partitioning.test
```
pdet, line 1: Can you add tests to all failing condition? Also add tests that stress case sensitiviness?
Can you add column types other than integers? including nested types?
Can you add tests using bucket partitining together with other partitions? identity, date, and whatnot.

pdet (commented):
Hi @Costa-SM,
thanks again for the PR! I think it looks great,  I'm just missing a few more tests, I left a comment with a few ideas, also I think there is a conflict with main!

redox:
@Costa-SM can I support you with this PR? It's only something we're interested in, I can definitely either help or take take over

Costa-SM:
Hi folks ! Sorry about that 😞 
I've merged main into the branch and added the test cases. 
Thanks for the offer to help @redox :)

src/storage/ducklake_transaction.cpp
```diff
@@ -773,6 +773,9 @@ DuckLakePartitionInfo DuckLakeTransaction::GetNewPartitionKey(DuckLakeCommitStat
 		case DuckLakeTransformType::HOUR:
 			partition_field.transform = "hour";
 			break;
+		case DuckLakeTransformType::BUCKET:
+			partition_field.transform = StringUtil::Format("bucket(%d)", field.transform.bucket_count);;
```
pdet: nit: ;;

src/storage/ducklake_insert.cpp
```diff
@@ -391,6 +392,25 @@ static unique_ptr<Expression> GetFunction(ClientContext &context, DuckLakeCopyIn
 	return function;
 }
 
+static unique_ptr<Expression> GetBucketExpression(ClientContext &context, DuckLakeCopyInput &copy_input,
```
pdet: This function is very similar to the one in `src/storage/ducklake_update.cpp`, is it possible we unify them?

test/sql/partitioning/bucket_partitioning.test
```diff
@@ -0,0 +1,384 @@
...
+# description: Test bucket partitioning
```
pdet, line 2: Can you also add a test with updates? And a test where you detatch attach?
You can also do them on different test files.

src/storage/ducklake_update.cpp
```diff
@@ -318,6 +319,26 @@ static unique_ptr<Expression> GetFunction(ClientContext &context, unique_ptr<Bou
...
+	auto hash_expr = GetFunction(context, std::move(column_reference), "hash");
+
+	// hash_value % bucket_count
+	vector<unique_ptr<Expression>> children;
+	children.emplace_back(std::move(hash_expr));
+	children.emplace_back(make_uniq<BoundConstantExpression>(Value::BIGINT(field.transform.bucket_count)));
+
+	ErrorData error;
+	FunctionBinder binder(context);
+	auto mod_expr = binder.BindScalarFunction(DEFAULT_SCHEMA, "%", std::move(children), error, false);
```
pdet: Is it possible we get a negative number?

pdet (commented):
Thanks! I just have a couple more comments!

src/storage/ducklake_transaction.cpp
Costa-SM: Fixed.

src/storage/ducklake_insert.cpp
Costa-SM: Yup, there were a few others that were very similar as well. I grouped everything into the new `ducklake_partition_data.cpp` file.

test/sql/partitioning/bucket_partitioning.test
Costa-SM: Yup, added. Kept them in the same file since they're all tests for the bucket partitioning.

src/storage/ducklake_update.cpp
Costa-SM: It was, my bad. The hash was a `UBIGINT`, so we could have some problems with that. I fixed it by moving the `bucket_count` from a `BIGINT` cast to `UBIGINT` as well.

src/storage/ducklake_partition_data.cpp
```diff
@@ -0,0 +1,65 @@
...
+	return function;
+}
+
+unique_ptr<Expression> DuckLakePartitionUtils::ApplyBucketTransform(ClientContext &context,
+                                                                    unique_ptr<Expression> column_expr,
+                                                                    idx_t bucket_count) {
+	D_ASSERT(bucket_count > 0);
+
+	// Not compatible with Iceberg since we're not using murmur3 hash, but instead DuckDB's own hash.
+	auto hash_expr = ApplyScalarFunction(context, "hash", std::move(column_expr));
```
Costa-SM: @pdet, should we be concerned about the stability of DuckDB's hash function? Is that going to change in the future? 
If so, the entries will be shuffled around.

pdet:
This looks great, can you just retag it to v1.5?

src/storage/ducklake_partition_data.cpp
pdet: I've chatted with @lnkuiper it is indeed not safe. A recommendation would be to use another hash function - maybe `xxhash`, that's what Parquet uses for bloom filters. It's stable

Tmonster: There is a PR open in duckdb-iceberg now to implement the Murmur3 hash. Maybe we can figure out a way to share this between the two? 
https://github.com/duckdb/duckdb-iceberg/pull/737

Costa-SM: If the murmur3 hash will only used for the bucket partitioning in DuckLake, we could add an implementation of it straight into the `ducklake_partition_data.cpp` file. WDYT? Or should I just keep the standard hash function for now?
Otherwise, to share the functions across the extensions, we'd have to add it to one of their common dependencies (DuckDB itself? Not sure if it'd make sense to support multiple hash functions straight from there).
I also saw that apparently there's [an extension](https://duckdb.org/community_extensions/extensions/hashfuncs) that provides different hash functions.

Tmonster: Long term the fix will be to add this to duckdb core yes. For this PR, however, I'm fine if it is added to `ducklake_partition_data.cpp`. I would like the default hash function to be the murmur3 hash to keep compatibility with Iceberg (I haven't looked closely at the PR, so unsure if that is currently the case).
This way, if an iceberg table is moved to a ducklake table, the way all of the partition values are calculated&read is consistent.

guillesd:
@pdet I think this was a bit lost in the weeds but maybe nice to take a look at it again?

pdet:
Yes, I think the PR looks great, the conflicts just need to be resolved.
@Costa-SM could you have a look at that?

guillesd:
@pdet this should be good to go!
Thanks @Costa-SM!

pdet:
Hi @guillesd, after having a look at the code again, I think my main concern is that the hash function won't be compatible with the iceberg spec as is.

I think we should change this for the 32-bit Murmur3 hash, x86 variant, seeded with 0. Otherwise, we might have to add yet another bucketing function for iceberg parity (as tom commented).

Maybe it is not very dramatic as we will have to do it for our other time partition function already, so I'm not strongly against that. WDYT?

guillesd:
We could implement mumur 3 for this PR but I guess in the future we will want to move this to DuckDB so that it can be reused across extensions

guillesd:
Based on Iceberg test vectors @pdet 

```
  ┌───────────┬─────────┬────────────┬────────────┐
  │   Input   │  Type   │  Expected  │ Our Result │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 34        │ int     │ 2017239379 │ 2017239379 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 'iceberg' │ string  │ 1210000089 │ 1210000089 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 1.0       │ float   │ -142385009 │ -142385009 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 1.0       │ double  │ -142385009 │ -142385009 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ true      │ boolean │ 1392991556 │ 1392991556 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 0.0       │ float   │ 1669671676 │ 1669671676 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ 0.0       │ double  │ 1669671676 │ 1669671676 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ -0.0      │ double  │ 1669671676 │ 1669671676 │
  ├───────────┼─────────┼────────────┼────────────┤
  │ ''        │ string  │ 0          │ 0          │
  └───────────┴─────────┴────────────┴────────────┘
  ```

File: test/sql/partitioning/bucket_partitioning.test [hunk 7]
```diff
@@ -0,0 +1,714 @@
...
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_bucket_partition', METADATA_CATALOG 'ducklake_metadata', DATA_INLINING_ROW_LIMIT 0)
```
pdet: @guillesd can we remove the `DATA_INLINING_ROW_LIMIT 0`? I think a few flushes will need to be added  to ensure files are created at the right time, but it just occured to me that this PR is not testing the flushing into partitions.

guillesd: Done! I just had to make `GetPartitionSQLExpression` return the actual bucketing sql expression (`(murmur3_32(col) & 2147483647) % N`). With this flushing works!

pdet:
Thanks for the PR @Costa-SM and @guillesd, LGTM!

PR #681, opened 2026-01-09
Fix stack overflow issue when binding views that reefer to tables that do not exist anymore

pdet:
When creating a view that references a table, if we detach and attach, that view becomes unbound, if we then drop the table and try to access the information of that view, we might enter an infinite recursion on the `GetSimilarEntry` path.

This PR adds a test and fixes this by overriding the `GetSimilarEntry` function and avoiding the binding code path.

cc: @prmoore77


prmoore77:
Thank you, @pdet!

PR #689, opened 2026-01-13
implicit conversion unique_ptr in return

qsliu2017:
use `std::move` to convert `unique_ptr<Derive>` to `unique_ptr<Base>`. some strict compiler will complain implicit conversion. (clang19 -std=c++11)

see https://stackoverflow.com/questions/36752678/clang-returning-stdunique-ptr-with-type-conversion.


pdet:
Thanks, LGTM

PR #690, opened 2026-01-13
Fix table snapshot lookup for inlined data reader

rgernhardt:
We've encountered a bug where using data inlining in combination with schema changes would result in error messages of the form: Cannot open file "ducklake_inlined_data_X_Y".  This happens when users make schema changes to multiple tables in the same transactions and later try to read from those tables.

**Problem**: The purpose of the function GetSnapshotForSchema is given a schema version and table we want to read, return the snpashot ID we can use to load the catalog. This should be the begin snapshot of the table (when the table was created). GetSnapshotForSchema uses GetCatalogIdForSchema. They query in that function is missing a filter on table id, so it implicitly assumes that there is only a single table entry in the ducklake_inlined_data_tables table for each schema version. This is not true, if users add columns to multiple tables in the same transaction. If this happens, GetSnapshotForSchema can return a wrong snapshot id.

**How the issue can manifests**: When the ducklake file reader tries to load an inlined table in TryCreateInlinedDataReader and GetSnapshotForSchema returns a wrong snapshot id at which the table did not exist, then then TryCreateInlinedDataReader returns nullptr and the reader falls back to trying to read the inlined table as a file (which does not exist). The user will see an error like: Cannot open file "ducklake_inlined_data_X_Y"

**Fix**: Since we already know the table id, we don't need to take the schema version into consideration. Just directly look up the begin snapshot of the table.

**Caching and test issue**: This is can be hidden by the schema cache. TryCreateInlinedDataReader uses GetEntryById to load the table schema. This uses the schema cache in ducklake_catalog, which will work even if we got the wrong snapshot id earlier. Thus, the test needs to detach and reattach to get a fresh cache.

**Other impacted codepaths**: GenerateFlushCommand is also using GetSnapshotForSchema. Here the issue could manifest during a call to ducklake_flush_inlined_data() as InternalException: "DuckLakeCompactor: failed to find table entry for given snapshot id".

### DIFF
--- a/src/include/metadata_manager/postgres_metadata_manager.hpp
+++ b/src/include/metadata_manager/postgres_metadata_manager.hpp
@@ -16,19 +16,24 @@ class PostgresMetadataManager : public DuckLakeMetadataManager {
 public:
 	explicit PostgresMetadataManager(DuckLakeTransaction &transaction);
 
+	static unique_ptr<DuckLakeMetadataManager> Create(DuckLakeTransaction &transaction) {
+		return make_uniq<PostgresMetadataManager>(transaction);
+	}
+
 	bool TypeIsNativelySupported(const LogicalType &type) override;
 
 	string GetColumnTypeInternal(const LogicalType &type) override;
 
 	unique_ptr<QueryResult> Execute(DuckLakeSnapshot snapshot, string &query) override;
 
-	unique_ptr<QueryResult> Query(DuckLakeSnapshot snapshot, string &query) override;
+	unique_ptr<QueryResult> Query(string query) override;
+	unique_ptr<QueryResult> Query(DuckLakeSnapshot snapshot, string query) override;
 
 protected:
 	string GetLatestSnapshotQuery() const override;
 
 private:
-	unique_ptr<QueryResult> ExecuteQuery(DuckLakeSnapshot snapshot, string &query, string command);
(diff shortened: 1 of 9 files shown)


rgernhardt:
**Open question for ducklake team**: 

Why do we try to take the schema_version into account when finding the snapshot id to load?
We already know the table id, couldn't we directly go to the ducklake_table and get the begin_snapshot from that table? That would be clearer and avoid the join from ducklake_inlined_data_tables on ducklake_table. 

I am probably missing some context

pdet:
> **Open question for ducklake team**:
> 
> Why do we try to take the schema_version into account when finding the snapshot id to load? We already know the table id, couldn't we directly go to the ducklake_table and get the begin_snapshot from that table? That would be clearer and avoid the join from ducklake_inlined_data_tables on ducklake_table.
> 
> I am probably missing some context

I think you are right, let's do that.

rgernhardt:
> > **Open question for ducklake team**:
> > Why do we try to take the schema_version into account when finding the snapshot id to load? We already know the table id, couldn't we directly go to the ducklake_table and get the begin_snapshot from that table? That would be clearer and avoid the join from ducklake_inlined_data_tables on ducklake_table.
> > I am probably missing some context
> 
> I think you are right, let's do that.

Done, simplified to directly lookup the beginSnapshot by table id

pdet:
Thanks for the PR, LGTM!

PR #693, opened 2026-01-13
Flush Deletions from Inlined Insertions

pdet:
Up to now, we were ignoring deletions in inlined insertions, which in practice would make us lose time-travel information by not flushing the deletes.

This PR, fixes that, when flushing data, we also group the deletes per `snapshot_end`, and produce the deletion files. They are included in `ducklake_delete_file` accordingly.

I've also verified we are able to deal with flushed that can produce multiple files, with multiple deletion files being generated for each data file. In addition, I fixed a bug where inlined data would be incorrectly scanned if we had a filter that produced an empty chunk, the multifile reader would interpret that as an EOF.


pdet:
Closing that for now, will reopen with partial deletion file implemented, I dont like writing all these files.

PR #694, opened 2026-01-14
Don't build the whole matrix

leo-altertable:
Build only for linux x86_64 and macos arm 64. Remove deploy step.


leo-altertable:
Sorry for the noise, did not noticed the PR opened to your repo by default.

PR #697, opened 2026-01-15
Add AUTOMATIC_MIGRATION option, and make automatic migration opt_in

pdet:
Before this PR when connecting a new version of the DuckLake extension, an automatic migration of the catalog would be triggered. Here we add a new option to the connection, `AUTOMATIC_MIGRATION` which by default is set to false. If false it triggers an error message saying there is a catalog mismatch, and that the user needs to set `AUTOMATIC_MIGRATION True` to use the extension.

This should prevent users from accidentally migrating their catalog.

cc @guillesd

### DIFF
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -142,12 +142,12 @@ void DuckLakeInitializer::LoadExistingDuckLake(DuckLakeTransaction &transaction)
 	for (auto &tag : metadata.tags) {
 		if (tag.key == "version") {
 			string version = tag.value;
-			if (version != "0.4-dev1" && !options.migrate_if_required) {
-				// Throw when Loading the Ducklake if a Migration is required and migrate_if_required option is false
-				throw InvalidInputException("DuckLake Extension requires a DuckLake Catalog version of 0.4-dev1 or "
-				                            "higher, current version is %s "
-				                            "and migrate_if_required is set to false",
-				                            version);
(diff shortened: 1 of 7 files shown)


src/storage/ducklake_initializer.cpp
```diff
@@ -142,12 +142,12 @@ void DuckLakeInitializer::LoadExistingDuckLake(DuckLakeTransaction &transaction)
 	for (auto &tag : metadata.tags) {
 		if (tag.key == "version") {
 			string version = tag.value;
-			if (version != "0.4-dev1" && !options.migrate_if_required) {
-				// Throw when Loading the Ducklake if a Migration is required and migrate_if_required option is false
-				throw InvalidInputException("DuckLake Extension requires a DuckLake Catalog version of 0.4-dev1 or "
-				                            "higher, current version is %s "
-				                            "and migrate_if_required is set to false",
-				                            version);
+			if (version != "0.4-dev1" && !options.automatic_migration) {
```
guillesd, line 145: I'm guessing this has to be changed when releasing 0.4? I mean `version != "0.4"`

pdet: Yeah this always needs to be the latest version, which currently is 0.4-dev1 but will be 0.4

vimota: Does that mean that we shouldn't expect Ducklake extension to be backwards compatible. So upgrading duckdb version means it won't work with prior ducklake catalog versions unless automatic migration is triggered?

PR #707, opened 2026-01-20
Remove partial_file_info from ducklake_data_file, replace it with partial_max and created_by_rewrite

pdet:
This PR removes the `partial_file_info` column from `ducklake_data_file`.

This column used to store the last snapshot present in a compacted file and the rows that belonged to each snapshot. We already store information regarding row to snapshot ownership in the `_ducklake_internal_snapshot_id` column in the file. We can safely remove this information from the catalog, especially because `partial_file_info` would grow in size with the number of snapshots compacted into a file.

We still need to store the last snapshot. For this, we create a new bigint column called `partial_max` in `ducklake_data_file`.

Additionally, we create a new boolean column named `created_by_rewrite`, which indicates whether a file was created by the `ducklake_rewrite_data_files` function. This is necessary because we cannot compact files created by that function together with other files that have a lower snapshot ID.

To illustrate this case, imagine the following data files:

| file_id | start_snapshot | end_snapshot |
|--------:|---------------:|-------------:|
| 1       | 1              | NULL         |
| 2       | 2              | NULL         |
| 3       | 3              | NULL         |

And the following delete files:

| delete_file_id | file_id | start_snapshot | end_snapshot |
|---------------:|--------:|---------------:|-------------:|
| 1              | 1       | 4              | NULL         |
| 2              | 2       | 5              | NULL         |

When running `ducklake_rewrite_data_files`, we would end up with:

| file_id | start_snapshot | end_snapshot |
|--------:|---------------:|-------------:|
| 1       | 1              | 5            |
| 2       | 2              | 5            |
| 3       | 3              | NULL         |
| 4       | 5              | NULL         |

| delete_file_id | file_id | start_snapshot | end_snapshot |
|---------------:|--------:|---------------:|-------------:|
| 1              | 1       | 4              | 5            |

Here, we create a new file that contains the merged data from files 1 and 2 with the applied deletions, and we adjust the end snapshots of the original files to match the start snapshot of the new file.

The problem arises if we later try to run a compaction. If we do not know that file 4 was written by `ducklake_rewrite_data_files`, the new compacted file could have its start snapshot set to that of file 3. This would cause the data from files 1 and 2 to be duplicated.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -278,6 +278,11 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 					// don't consider merging if the file is larger than the target size
 					break;
 				}
+				// Files with created_by_rewrite = true can only be merged if they have the lowest snapshot
+				// (i.e., they are the first file in the merge). If encountered later, we can't merge them in.
+				if (candidate.created_by_rewrite && compaction_idx > start_idx) {
+					break;
+				}
 				// this file can be compacted along with the neighbors
 				current_file_size += file_size;
 			}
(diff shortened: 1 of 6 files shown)


pdet:
Closing this in favour of: https://github.com/duckdb/ducklake/pull/708

PR #708, opened 2026-01-20
Remove partial_file_info from ducklake_data_file replace it with partial_max

pdet:
This PR removes the `partial_file_info` column from `ducklake_data_file`.

This column used to store the last snapshot present in a compacted file and the rows that belonged to each snapshot. We already store information regarding row to snapshot ownership in the `_ducklake_internal_snapshot_id` column in the file. We can safely remove this information from the catalog, especially because `partial_file_info` would grow in size with the number of snapshots compacted into a file.

We still need to store the last snapshot. For this, we create a new bigint column called `partial_max` in `ducklake_data_file`.


djouallah:
@pdet do you mind clarify this as I could not find any definitive answer, is there a way to compact, force etc a table to expire all snapshots if they refer to a partial data file,  otherwise, this will create a breaking point with iceberg data files,  assuming one don't care about  table history.

PR #712, opened 2026-01-21
Cleanup old files

pdet:


src/storage/ducklake_transaction.cpp
```diff
@@ -2129,6 +2129,14 @@ void DuckLakeTransaction::TransactionLocalDelete(TableIndex table_id, const stri
 	auto &table_changes = entry->second;
 	for (auto &file : table_changes.new_data_files) {
 		if (file.file_name == data_file_path) {
+			if (!file.delete_files.empty()) {
+				auto context_ref = context.lock();
+				auto &fs = FileSystem::GetFileSystem(*context_ref);
+				for (auto &old_file : file.delete_files) {
+					fs.RemoveFile(old_file.file_name);
+				}
```
utay, line 2139: Hi @pdet, I just stumbled on this one. Should it use `fs.RemoveFiles` instead? To benefit from future bulk delete improvements

pdet: Absolutely!

PR #721, opened 2026-01-26
Generate Partial Deletion files over consecutive deletes

pdet:
We introduced partial deletion files in https://github.com/duckdb/ducklake/pull/696

This PR continues that work by also creating partial deletion files by merging pre-existing deletion files into new ones.

In practice, when a file has it's first deletion, the deletion file will follow the old format. Any subsequent delete will schedule the previous file to be deleted and will generate a new deletion file, which is now a partial deletion file with information of which deletes were created in which snapshots.

This PR also extends the tests to the `ducklake_table_deletions` function. Fixes many issues with them, and upgrade them to work with partial deletion files.

I've also added aliases for `ducklake_table_deletions` to `table_deletions` and `ducklake_table_insertions` to `table_insertions`.

@guillesd I think we are missing information on the `ducklake_table_deletions` and `ducklake_table_insertions` functions, maybe we should add to the data_change_feed page? https://ducklake.select/docs/stable/duckdb/advanced_features/data_change_feed

### DIFF
--- a/test/configs/attach_ducklake.json
+++ b/test/configs/attach_ducklake.json
@@ -463,7 +463,8 @@
     {
       "reason": "FIXME: Wrong result",
       "paths": [
-        "test/sql/join/positional/test_positional_join.test"
+        "test/sql/join/positional/test_positional_join.test",
+        "test/sql/update/test_null_update.test"
       ]
     },
     {


guillesd:
Hey @pdet, if I understand correctly we are rewriting delete files on subsequent deletes to partial delete files. If we do this by default, I guess there is a performance impact on deletions and an improvement in rewriting files and reads? Maybe it is nice to track this things? Like adding some more .benchmark files around? Not a strong opinion though, just an idea.

On another note, what is the size limit of these partial delete files? Is it the same as the data files?

pdet:
> Hey @pdet, if I understand correctly we are rewriting delete files on subsequent deletes to partial delete files. If we do this by default, I guess there is a performance impact on deletions and an improvement in rewriting files and reads? Maybe it is nice to track this things? 

I think the performance impact on writing should not be significant, since we are just adding an integer column to the file. When new deletes are applied, we always have to rewrite the entire deletion file anyway. It could have an impact during time travel, as we now have to filter out snapshots from the Parquet file, whereas before we knew we could always just read the whole thing. On the positive side, this simplifies things significantly, since we only need to keep one deletion file around, and it also substantially reduces disk usage.

> Like adding some more .benchmark files around? Not a strong opinion though, just an idea.

I fully agree with that. My plan is to add proper regression infrastructure with benchmarks in the future, but I think that is a bit orthogonal to this PR, I will tackle this after finishing the missing blocks for DuckLake 1.0.

> On another note, what is the size limit of these partial delete files? Is it the same as the data files?

We do not really impose a size limit on deletion files as we do on data files. The reason for that is that they will likely be much smaller than the data files in all cases, since they always store at most three columns, which are all highly compressible, and will have at most the same number of rows as the data files. Of course if your table is just a boolean column, and you delete 99.9% of it, there won't be true, but that is quite a corner case, we could have a look at that if it ever becomes a problem.

PR #728, opened 2026-01-29
Optimize COUNT(*) queries

rgernhardt:
Add DuckLakeGetPartitionStats to enable COUNT(*) queries to return results from metadata without scanning parquet files.

The optimization is disabled for:
- COUNT(*) in time travel queries
- COUNT(*) on Transaction-local tables
- COUNT(*) in transactions on tables with uncomitted local changes

After this change, runtime for COUNT(*) is nearly constant regardless of number of files

Benchmarked on S3 backed tables (us-east-1) using my Macbook M4 (no inlined data or deletes): 

| Files | COUNT(*) | COUNT(*) optimized | Speedup |
|-------|----------|--------------------|---------|
| 10    | 361ms    | 42ms               | 8.6x    |
| 100   | 2,716ms  | 42ms               | 64x     |
| 500   | 10,901ms | 42ms               | 258x    |

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -129,6 +129,9 @@ public:
 	                                                                  DuckLakeSnapshot snapshot,
 	                                                                  DuckLakeFileSizeOptions options);
 	virtual idx_t GetBeginSnapshotForTable(TableIndex table_id);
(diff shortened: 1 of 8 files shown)


rgernhardt:
Btw. had to run make pull and update the duckdb and extension submodules, the build was breaking without that. Think there was some API changes in MultiFileList

PR #732, opened 2026-01-29
Count star optimization

rgernhardt:
Add DuckLakeGetPartitionStats to enable COUNT(*) queries to return results from metadata without scanning parquet files.

The optimization is disabled for:
- COUNT(*) in time travel queries
- COUNT(*) on Transaction-local tables
- COUNT(*) in transactions on tables with uncomitted local changes

After this change, runtime for COUNT(*) is nearly constant regardless of number of files

Benchmarked on S3 backed tables (us-east-1) using my Macbook M4 (no inlined data or deletes): 

| Files | COUNT(*) | COUNT(*) optimized | Speedup |
|-------|----------|--------------------|---------|
| 10    | 361ms    | 42ms               | 8.6x    |
| 100   | 2,716ms  | 42ms               | 64x     |
| 500   | 10,901ms | 42ms               | 258x    |

### DIFF
--- a/.github/workflows/MinIO.yml
+++ b/.github/workflows/MinIO.yml
@@ -21,6 +21,7 @@ jobs:
       VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
       PIP_BREAK_SYSTEM_PACKAGES: 1
       BUILD_EXTENSION_TEST_DEPS: full
+      S3_TEST_SERVER_AVAILABLE: 1
 
     steps:
       - name: Install required ubuntu packages
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -77,6 +77,9 @@ static void LoadInternal(ExtensionLoader &loader) {
 	DuckLakeLastCommittedSnapshotFunction last_committed;
 	loader.RegisterFunction(last_committed);
 
+	DuckLakeSettingsFunction settings;
+	loader.RegisterFunction(settings);
+
 	// secrets
 	auto secret_type = DuckLakeSecret::GetSecretType();
 	loader.RegisterSecretType(secret_type);
--- a/src/functions/CMakeLists.txt
+++ b/src/functions/CMakeLists.txt
@@ -13,6 +13,7 @@ add_library(
   ducklake_set_option.cpp
   ducklake_snapshots.cpp
   ducklake_options.cpp
+  ducklake_settings.cpp
   ducklake_table_changes.cpp
   ducklake_table_info.cpp
   ducklake_table_insertions.cpp)
--- a/src/functions/ducklake_settings.cpp
+++ b/src/functions/ducklake_settings.cpp
@@ -1,0 +1,51 @@
+#include "functions/ducklake_table_functions.hpp"
+#include "storage/ducklake_transaction.hpp"
+#include "storage/ducklake_catalog.hpp"
+
+namespace duckdb {
+
+static unique_ptr<FunctionData> DuckLakeSettingsBind(ClientContext &context, TableFunctionBindInput &input,
+                                                     vector<LogicalType> &return_types, vector<string> &names) {
+	auto &catalog = BaseMetadataFunction::GetCatalog(context, input.inputs[0]);
+	auto &ducklake_catalog = catalog.Cast<DuckLakeCatalog>();
+
+	names.emplace_back("catalog_type");
+	return_types.emplace_back(LogicalType::VARCHAR);
+
+	names.emplace_back("extension_version");
+	return_types.emplace_back(LogicalType::VARCHAR);
+
+	names.emplace_back("data_path");
+	return_types.emplace_back(LogicalType::VARCHAR);
+
+	auto result = make_uniq<MetadataBindData>();
+	vector<Value> row_values;
+
+	// catalog_type: default to "duckdb" if not explicitly set (DuckDB is the default metadata storage)
+	auto catalog_type = ducklake_catalog.MetadataType();
+	if (catalog_type.empty()) {
+		catalog_type = "duckdb";
+	} else if (catalog_type == "postgres_scanner") {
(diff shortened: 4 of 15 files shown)


rgernhardt:
I had to update the submodules to get the build to pass, there seem to be some interface changes in the main branch at the commit [@ 18d7116](https://github.com/duckdb/duckdb/tree/18d7116c829cb70498caf18b2639d5e63abcfb72) that are not compatible. Locally all tests are passing

Saw some test failures that seem unrelated to my changes:
- sorted_table tests (9 failures) - Missing features in ducklake
- inlining failures with SQLite backend - Seems preexisting? Failing for me when I run locally with SqlLite backend as well without any of my changes

Now I am actually seeing build errors in CI: 
Seems like the version specified in .github/duckdb-version is used for CI and not in sync with the submodule commit. 
Will wait for that to be fixed

test/sql/stats/count_star_optimization.test
```diff
@@ -0,0 +1,244 @@
+# name: test/sql/stats/count_star_optimization.test
```
pdet: Can you also test with `ducklake_add_data_files`  and `ducklake_rewrite_data_files`

pdet: If possible in other test files

src/storage/ducklake_transaction.cpp
```diff
@@ -2187,6 +2187,14 @@ bool DuckLakeTransaction::HasLocalDeletes(TableIndex table_id) {
 	return !entry->second.new_delete_files.empty();
 }
 
+bool DuckLakeTransaction::HasAnyLocalChanges(TableIndex table_id) {
```
pdet, line 2190: Is this the same as `HasTransactionLocalChanges`?

test/sql/stats/count_star_optimization.test
pdet: Can you also modularize this test a bit? e.g., having ht einline data test in a different test, makes it easier to review.

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -331,6 +331,50 @@ WHERE table_id = {TABLE_ID})";
 	throw InternalException("Table %llu does not exist", table_id.index);
 }
 
+idx_t DuckLakeMetadataManager::GetTotalDeleteCount(TableIndex table_id, DuckLakeSnapshot snapshot) {
+	string query = R"(
+SELECT COALESCE(SUM(delete_count), 0)
+FROM {METADATA_CATALOG}.ducklake_delete_file
+WHERE table_id = {TABLE_ID}
+  AND ({SNAPSHOT_ID} >= begin_snapshot OR begin_snapshot IS NULL)
```
pdet: Is there a situation where `begin_snapshot IS NULL`?

test/sql/stats/count_star_optimization.test
pdet: can you also test with inlined data over multiple table schemas (e.g., add column, remove column and so on)

pdet:
Hi `main`'s ci should be passing now, could you merge with it?

src/storage/ducklake_transaction.cpp
rgernhardt: It looks like `HasTransactionLocalChanges` only checks for newly created data files or new inline rows.
`HasAnyLocalChanges` in addition also detects deletes, inlined_deletes, compactions or file deletes.

I renamed `HasTransactionLocalChanges` -> `HasTransactionLocalInserts` to clarify

src/storage/ducklake_metadata_manager.cpp
rgernhardt: You're right, removing

test/sql/stats/count_star_optimization.test
rgernhardt: I tried a more systematic approach for the tests now, let me know what you think

rgernhardt:
@pdet I tried to cover all the scenarios that should cause fallback to scan now across regular rows, inlined rows and the different maintenance functions.

Just need your guidance on `HasAnyLocalChanges` vs `HasTransactionLocalChanges` now:

I am pretty sure the existing `HasTransactionLocalChanges` is not working as intended. It's used to signal not to use the table stats when there are *any* transaction local changes, but it only checks for inserts not deletes (or truncates).

This confirms the bug to me

```
  -- Setup                                                                                                                                 
  ATTACH 'ducklake:test.db' AS ducklake (DATA_PATH 'data');                                                                                
  CREATE TABLE ducklake.test(i INTEGER);                                                                                                   
  INSERT INTO ducklake.test FROM range(1000);                                                                                              
                                                                                                                                           
  -- Begin transaction and delete 900 rows                                                                                                 
  BEGIN;                                                                                                                                   
  DELETE FROM ducklake.test WHERE i >= 100;                                                                                                
                                                                                                                                           
  -- COUNT(*) correctly shows 100 (uses HasAnyLocalChanges)                                                                                
  SELECT COUNT(*) FROM ducklake.test;                                                                                                      
  -- Returns: 100                                                                                                                          
                                                                                                                                           
  -- EXPLAIN incorrectly shows ~1,000 (uses HasTransactionLocalChanges which misses deletes)                                               
  EXPLAIN SELECT * FROM ducklake.test;                                                                                                     
  -- Returns: ~1,000 rows  <-- BUG: should not use stale stats                                                                             
                                                                                                                                           
  ROLLBACK;                                                                                                                                
```

*But*, as far as I can tell the table level stats are already fundamentally not accounting for with deletes. So it's not like this will fix much.
Given that, what is you prefernce?

A. Change `HasTransactionLocalChanges` to account for all local changes and then use it for the count(*) optimization as well and delete the new `HasAnyLocalChanges`?
B. Leave things as is any maybe just rename `HasTransactionLocalChanges` -> `HasTransactionLocalInserts`?

pdet:
Ha, got it, we can go with B! Thanks for looking into this in-depth!

pdet:
Thanks, all looks great, the failure is unrelated!

PR #734, opened 2026-01-29
Maintenance functions for merge/rewrite/flush returns information about what was processed

rgernhardt:
Add return values for merge/rewrite/flush operations so users can reason about the impact of their runs.

**Flushing data**:

ducklake_flush_inlined_data now returns 3 columns:                                                                                          
- schema_name (VARCHAR) - the name of the schema                                                                                               
- table_name (VARCHAR) - the name of the table that was flushed                                                                              
- rows_flushed (BIGINT) - number of rows flushed from inlined data

There is one row for each table that had any flushes.

**Merging data:**

ducklake_merge_adjacent_files and ducklake_rewrite_data_files return 4 columns:                       
- schema_name (VARCHAR) - the name of the schema                                                                                                                                     
- table_name (VARCHAR) - the name of the table                                                                                               
- files_processed (BIGINT) - number of input files that were merged/rewritten                                                                
- files_created (BIGINT) - number of output files created.                                                                      
                                                                                                                                             

There is one row **per output file created**. So the value for files_created in each row is always 1. This show exactly how many files were compacted into each output file and can give insights into compaction efficiency.

To see the total number of output files produced by table, users could run 
 ```sql
  select schema_name, table_name, sum(files_created) from ducklake_merge_adjacent_files('db') group by schema_name, table_name
```

Also, ducklake_merge_adjacent_files now supports the option max_compacted_files when passing in only a database name. The limit is *per table*
```
SELECT table_name, files_processed, files_created FROM ducklake_merge_adjacent_files('ducklake', max_compacted_files=>1) ORDER BY table_name

# Output will show at most 1 row per table
table1	N	1
table2	N	1  
```

### DIFF
--- a/test/sql/data_inlining/data_inlining_encryption.test
+++ b/test/sql/data_inlining/data_inlining_encryption.test
@@ -33,7 +33,6 @@ SELECT COUNT(*) FROM GLOB('${DATA_PATH}/ducklake_inlining_encryption_files/main/
 query III
 SELECT schema_name, table_name, rows_flushed FROM ducklake_flush_inlined_data('ducklake') ORDER BY ALL
 ----
-main	test	0
 main	test	1000
 
 # now we have a file

### CODE REVIEW

**Comment by rgernhardt** [comment]
> Thanks for the PR, looks great! We have a few tests that call these functions with `statement ok` instead of verifying their result. Could you add result verification? I think this would not only be useful for more thorough testing of this PR, but also serve as tests for future modifications (e.g., if we break something during inlining or unexpectedly change the behavior of a compaction method).

Done. After this I was able to remove the two tests I added earlier, because all the existing tests cover all scenarios. I also realized we need the schema_name to be able to tell apart tables with the same name in different schemas.

**Comment by rgernhardt** [comment]
@pdet I noticed something about ducklake_flush_inlined_data I didn't realize before.

We create one flush_operation for each inlined table (and even if it has 0 rows to flush). So if a single ducklake table has multiple schema changes and thus multiple inlined tables, we generate a union of multiple flush operators. Then the output shows multiple rows for the single table.

We could make the function just return one row per table by wrapping the union operator with an aggregation operation when we construct the logical tree. That would effectively do the `select schema_name, table_name, sum(files_created) from ducklake_merge_adjacent_files('db') group by schema_name, table_name` for the user. Then we'd really get at most one row per table. (we could also add a `having rows > 0` to filter inlined table with 0 rows)

I actually got this to work. But it adds ~70 more lines of code to the `ducklake_flush_inlined_data.cpp` to construct the new part of the query tree. So not sure if that's something you'd want.
We could probably do something similar for the compaction operator and show at most one row per table, but would add a similar amount of code.

For now, I've left the structure of the union operator as is. And you could just document the recommended query to aggregate and filter to the user.

**Comment by pdet** [comment]
Is your proposed solution to do a post-union aggregation?
e.g.,
```sql
SELECT
  table_id,
  SUM(rows_flushed) AS rows_flushed,
  MAX(last_version) AS last_version
FROM (
  flush_1
  UNION ALL
  flush_2
  UNION ALL
  flush_3
) t
GROUP BY table_id;
```

If yes, I think we can do that.

rgernhardt:
> Is your proposed solution to do a post-union aggregation? e.g.,
> 
> ```sql
> SELECT
>   table_id,
>   SUM(rows_flushed) AS rows_flushed,
>   MAX(last_version) AS last_version
> FROM (
>   flush_1
>   UNION ALL
>   flush_2
>   UNION ALL
>   flush_3
> ) t
> GROUP BY table_id;
> ```
> 
> If yes, I think we can do that.

Yes exactly.
Today, the bind operation returns the union op 
```
(flush_1 
UNION ALL
flush_2
UNION ALL
flush_3)
```

We could construct the operator tree for the query you describe in code and return that instead. 
If that sounds good to you, I can do that. It adds a bit more complexity to the code, but makes the return value of the function more intuitive.

pdet:
> > Is your proposed solution to do a post-union aggregation? e.g.,
> > ```sql
> > SELECT
> >   table_id,
> >   SUM(rows_flushed) AS rows_flushed,
> >   MAX(last_version) AS last_version
> > FROM (
> >   flush_1
> >   UNION ALL
> >   flush_2
> >   UNION ALL
> >   flush_3
> > ) t
> > GROUP BY table_id;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If yes, I think we can do that.
> 
> Yes exactly. Today, the bind operation returns the union op
> 
> ```
> (flush_1 
> UNION ALL
> flush_2
> UNION ALL
> flush_3)
> ```
> 
> We could construct the operator tree for the query you describe in code and return that instead. If that sounds good to you, I can do that. It adds a bit more complexity to the code, but makes the return value of the function more intuitive.

Sounds good to me!

rgernhardt:
@pdet I added the output process for the flush_inlined_data function as discussed. Now we get exactly one row per table that had any flushes.
I left the output for merge_adjacent_files as is, because I think it can give useful information about the exact distribution between input and output files for merges.

Also added the small fix to allow max_compacted_files parameter on merge_adjacent_files when running on a database

pdet:
Looks great! Thanks for all the hard work, the failure is unrelated!

PR #735, opened 2026-01-30
Fix ducklake_add_data_files with multiple hive-partitioned columns

gijshendriksen:
I ran into an issue with the `ducklake_add_data_files` function when processing files that had multiple hive-partitioned columns.

The issue turned out to be that `DuckLakeFileProcessor::AddFileToTable` assumes that `table.GetPartitionData()` and `file.hive_partition_values` were both sorted on `field_id`. However, `file.hive_partition_values` was created by iterating over an `unordered_map` (in `DuckLakeFileProcessor::MapColumns`) and so had a different ordering. This difference was handled fine by the actual column mapping, but not in the check whether all fields were valid. I changed this check to be order-independent, and now the files get added properly.

Let me know if you agree with this change and whether you think the check is still valid now. I couldn't figure out in what scenario the fields would actually be populated to be invalid, so I did not add a test case to verify the check. If you have any suggestions there I'd be happy to add test cases as well.

## To reproduce

```sql
CREATE SECRET (
    TYPE s3,
    ENDPOINT 'a3s.fi',
    URL_STYLE 'path',
    SCOPE 's3://2006391-owi-remote-index'
);

ATTACH 'ducklake:test.ducklake' AS dl;
USE dl;

CREATE TABLE remote_index AS
FROM 's3://2006391-owi-remote-index/year=2025/month=12/day=25/collection=26eb6c02-e237-11f0-b4b1-8ebf6bb2cab9/index/representation=title/language=eng/postings_0.parquet'
WITH NO DATA;

ALTER TABLE remote_index SET PARTITIONED BY (
    year, month, day, collection, representation, language
);

CALL ducklake_add_data_files('dl', 'remote_index', 's3://2006391-owi-remote-index/year=2025/month=12/day=25/collection=26eb6c02-e237-11f0-b4b1-8ebf6bb2cab9/index/representation=title/language=eng/postings_0.parquet');
```

Before my changes, I would get the following error:
```
Invalid Input Error:
File "s3://2006391-owi-remote-index/year=2025/month=12/day=25/collection=26eb6c02-e237-11f0-b4b1-8ebf6bb2cab9/index/representation=title/language=eng/postings_0.parquet" contains an invalid partition value for the table configuration.
```

After adding the fix, the file is added and  `ducklake_file_partition_value` is populated correctly.

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -1172,8 +1172,15 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 		if (file.hive_partition_values.size() != partition_data->fields.size()) {
 			invalid_partition = true;
 		} else {
-			for (idx_t i = 0; i < partition_data->fields.size(); i++) {
-				if (file.hive_partition_values[i].field_index.index != partition_data->fields[i].field_id.index) {
+			for (auto hive_partition_value : file.hive_partition_values) {
+				bool found_field = false;
+				for (auto field : partition_data->fields) {
+					if (field.field_id.index == hive_partition_value.field_index.index) {
+						found_field = true;
+						break;
+					}
+				}
+				if (!found_field) {
 					invalid_partition = true;
 					break;
 				}
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -1172,9 +1172,9 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 		if (file.hive_partition_values.size() != partition_data->fields.size()) {
 			invalid_partition = true;
 		} else {
-			for (auto hive_partition_value : file.hive_partition_values) {
+			for (const auto& hive_partition_value : file.hive_partition_values) {
 				bool found_field = false;
-				for (auto field : partition_data->fields) {
+				for (const auto& field : partition_data->fields) {
 					if (field.field_id.index == hive_partition_value.field_index.index) {
 						found_field = true;
 						break;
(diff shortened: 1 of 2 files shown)


pdet:
Hi @gijshendriksen, could you add a test? It's easier for me to visualize the problem that way.

gijshendriksen:
Hi @pdet, thanks for the quick response. I've added a test case that fails on `main` but succeeds on my branch.

gijshendriksen:
I've slightly extended the test case, to check that `ducklake_file_partition_value` is populated correctly, and to verify that an error is still raised when the Hive partition keys do not match. I'm not quite sure if there are other cases for which the error should be raised that we should try to include in the test case. If you have suggestions, let me know!

src/functions/ducklake_add_data_files.cpp
```diff
@@ -1172,8 +1172,15 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 		if (file.hive_partition_values.size() != partition_data->fields.size()) {
 			invalid_partition = true;
 		} else {
-			for (idx_t i = 0; i < partition_data->fields.size(); i++) {
-				if (file.hive_partition_values[i].field_index.index != partition_data->fields[i].field_id.index) {
+			for (auto hive_partition_value : file.hive_partition_values) {
```
pdet: const auto&

src/functions/ducklake_add_data_files.cpp
```diff
@@ -1172,8 +1172,15 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
...
+				bool found_field = false;
+				for (auto field : partition_data->fields) {
```
pdet: const auto&

pdet (commented):
Thanks for the PR, I think it looks great and the change is clear, just have a small comment, and is good to go!

gijshendriksen:
Thanks for the review @pdet! I've applied the suggested changes.

pdet:
Thanks again for the great work!

PR #737, opened 2026-02-02
Deletion Inlining

pdet:
This PR implements deletion inlining for DuckLake. Similar to insertion inlining, deletions that fall under our inlining threshold (i.e., the `DATA_INLINING_ROW_LIMIT` setting) are directly stored in the metadata database instead of creating separate Parquet delete files. This approach reduces the overhead of writing and maintaining many small files.

A new metadata table is created per DuckLake table to track inlined deletions. Each table that has inlined deletions will have its own table containing information about the deletions:

```sql
CREATE TABLE ducklake_inlined_delete_{table_id}(
    file_id BIGINT,        
    row_id BIGINT,         
    begin_snapshot BIGINT 
);
```

Inlined deletions can be flushed to Parquet delete files using `ducklake_flush_inlined_data()`:

```sql
-- Flush all inlined data (inserts and deletions) to Parquet files
```

When flushed, inlined deletions are stored as partial deletion files, hence they have a `_ducklake_internal_snapshot_id` column for time-travel support. 
A file will be created per existing `file_id` in the deletion inlining table . 
If the table already has a deletion file for a given file_id, the files are merged, in the same way as partial deletion files happen.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -332,11 +332,27 @@ WHERE table_id = {TABLE_ID})";
 }
 
 idx_t DuckLakeMetadataManager::GetNetDataFileRowCount(TableIndex table_id, DuckLakeSnapshot snapshot) {
-	// Compute sum(record_count) - sum(delete_count) in a single query
+	// Compute sum(record_count) - sum(delete_count) - inlined_deletions in a single query
 	// Delete files are only counted if their corresponding data file is still visible
 	// This handles TRUNCATE correctly: when a data file's end_snapshot is set,
 	// its associated delete files are no longer counted
-	string query = R"(
+
+	// Check if inlined deletion table exists
+	auto inlined_table_name = GetInlinedDeletionTableName(table_id, snapshot);
+
+	string inlined_deletion_subquery = "0";
+	if (!inlined_table_name.empty()) {
+		inlined_deletion_subquery = StringUtil::Format(R"(
+COALESCE((SELECT COUNT(*) FROM {METADATA_CATALOG}.%s del
+          JOIN {METADATA_CATALOG}.ducklake_data_file data ON del.file_id = data.data_file_id
+          WHERE del.begin_snapshot <= {SNAPSHOT_ID}
+            AND data.table_id = {TABLE_ID}
+            AND {SNAPSHOT_ID} >= data.begin_snapshot
+            AND ({SNAPSHOT_ID} < data.end_snapshot OR data.end_snapshot IS NULL)), 0))",
+		                                               inlined_table_name);
+	}
+
+	string query = StringUtil::Format(R"(
 SELECT
   COALESCE((SELECT SUM(record_count) FROM {METADATA_CATALOG}.ducklake_data_file
             WHERE table_id = {TABLE_ID}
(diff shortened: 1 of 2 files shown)


src/functions/ducklake_snapshots.cpp
```diff
@@ -140,7 +140,9 @@ vector<Value> DuckLakeSnapshotsFunction::GetSnapshotValues(const DuckLakeSnapsho
 	PushIDChangeList(change_keys, change_values, other_changes.altered_views, "views_altered");
 	PushIDChangeList(change_keys, change_values, other_changes.tables_inserted_inlined, "inlined_insert");
 	PushIDChangeList(change_keys, change_values, other_changes.tables_deleted_inlined, "inlined_delete");
-	PushIDChangeList(change_keys, change_values, other_changes.tables_flushed_inlined, "flushed_inlined");
+	PushIDChangeList(change_keys, change_values, other_changes.tables_flushed_inlined, "inline_flush");
```
Mytherin: Is this change on purpose (`flush_inlined -> inline_flush`)? Would we not have to do a migration to change anything here (or was this never part of a published spec version?)

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -80,6 +80,7 @@ CREATE TABLE {METADATA_CATALOG}.ducklake_partition_column(partition_id BIGINT, t
 CREATE TABLE {METADATA_CATALOG}.ducklake_file_partition_value(data_file_id BIGINT, table_id BIGINT, partition_key_index BIGINT, partition_value VARCHAR);
 CREATE TABLE {METADATA_CATALOG}.ducklake_files_scheduled_for_deletion(data_file_id BIGINT, path VARCHAR, path_is_relative BOOLEAN, schedule_start TIMESTAMPTZ);
 CREATE TABLE {METADATA_CATALOG}.ducklake_inlined_data_tables(table_id BIGINT, table_name VARCHAR, schema_version BIGINT);
+CREATE TABLE {METADATA_CATALOG}.ducklake_inlined_deletion_tables(table_id BIGINT, table_name VARCHAR);
```
Mytherin: The reason this table exists for inlined insertions is that there can be multiple versions we might need to read from if the table has been altered. Given that that's not the case for inlined deletions, we could skip having this table and instead check the schema for whether or not the table with a specific name (`ducklake_inlined_delete_ {table_id}`) exists.

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1985,6 +2099,150 @@ WHERE row_id=deleted_row_id;
...
+		if (chunk && chunk->size() > 0) {
+			auto table_name = chunk->GetValue(0, 0).ToString();
+			delete_inlined_table_name_cache[table_id.index] = table_name;
+			return table_name;
+		}
+	}
+
+	// Table doesn't exist
+	if (!create_if_not_exists) {
+		return string();
```
Mytherin: If we get rid of `ducklake_inlined_deletion_tables` we could do a `CREATE TABLE IF NOT EXISTS` instead of first checking for existence here (when `create_if_not_exists` is true).

Mytherin (commented):
Thanks for the PR! Great work - left some comments

src/storage/ducklake_metadata_manager.cpp
pdet: Yeah, fair point.

src/functions/ducklake_snapshots.cpp
pdet: Ah true, we have this in the catalog

PR #742, opened 2026-02-03
metadata manager register

qsliu2017:


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -28,11 +28,26 @@ DuckLakeMetadataManager::~DuckLakeMetadataManager() {
...
+	metadata_managers[name] = create;
+}
+
 unique_ptr<DuckLakeMetadataManager> DuckLakeMetadataManager::Create(DuckLakeTransaction &transaction) {
+	lock_guard<mutex> lock(metadata_managers_lock);
 	auto &catalog = transaction.GetCatalog();
 	auto catalog_type = catalog.MetadataType();
-	if (catalog_type == "postgres" || catalog_type == "postgres_scanner") {
-		return make_uniq<PostgresMetadataManager>(transaction);
+	auto create = metadata_managers[catalog_type];
```
pdet: [] will insert a null entry, is that why you have the if check?
If so, should we have a find() and a throw?

pdet:
Thanks, LGTM, when CI is green I'll merge it!

PR #743, opened 2026-02-03
Sort expressions for compaction and inline flush

Alex-Monahan:
This PR removes one limitation of #642 : We can now `SET SORTED BY` expressions in addition to column names. These expressions can be nested and they can even be DuckLake macros. I also added a test using spatial's `ST_HILBERT` function (which is currently skipped in CI, but runs correctly in local testing!) which was the inspiration for this chain of PRs in the first place! (Rebuilding this blog post in DuckLake: https://duckdb.org/2025/06/06/advanced-sorting-for-fast-selective-queries )

The sort expressions are validated when the `SET SORTED BY` statement is executed and an error message will return all of the columns that were not found. 

The PR also includes a few artifacts from running `make format`. Let me know if you want those reverted.

I added a fairly wide variety of tests, but I am always open to adding others! Feedback welcome as well!

While Claude did help me with this, I have reviewed the code extensively and written some of my own tests manually as well. I also forced it to combine some functions when it decided to create extra ones for fun... So I aimed to keep things modular. Such are the times we live in!

### DIFF
--- a/test/sql/sorted_table/merge_adjacent_sorted_nested_expression.test
+++ b/test/sql/sorted_table/merge_adjacent_sorted_nested_expression.test
@@ -1,5 +1,5 @@
 # name: test/sql/sorted_table/merge_adjacent_sorted_nested_expression.test
-# description: test ducklake merge adjacent files with SET SORTED BY using nested expressions
+# description: test ducklake merge adjacent files with SET SORTED BY using a nested expression
 # group: [sorted_table]
 
 require ducklake


Alex-Monahan:
Moving to draft until my local CI is passing

Alex-Monahan:
I narrowed down the issue. It appears to be caused by when a macro is created and then dropped within the same transaction. I can reproduce it without any sorting or inlining. I'm going to remove it from the test and add a separate issue for it. I'll link it here.

Alex-Monahan:
Unrelated bug I encountered while building tests for this (creating then dropping a macro in the same transaction) is here: https://github.com/duckdb/ducklake/issues/744

Alex-Monahan:
I believe the failure is unrelated - tests are passing on my fork here: https://github.com/Alex-Monahan/ducklake/actions .

Alex-Monahan:
Ah shoot. Looks like the spatial test is getting skipped. Is there any way I can test that in CI in some way? I have not been able to get spatial building locally yet, but I can also try that route as well if you recommend it.

pdet:
> * Main Extension Distribution Pipeline / Build extension binaries / Windows (windows_amd64, windows-latest, x64-windows-static-release, x64-windows-static-release, t... (pull_request)

failure is a missing order by, i'll fix it in a bit

pdet:
> Ah shoot. Looks like the spatial test is getting skipped. Is there any way I can test that in CI in some way? I have not been able to get spatial building locally yet, but I can also try that route as well if you recommend it.

yeah the geo tests are currently not executed in CI because they require the geo extension to be built and neither me or @Maxxen had the time yet to fiddle with that yet :(

Alex-Monahan:
> > Ah shoot. Looks like the spatial test is getting skipped. Is there any way I can test that in CI in some way? I have not been able to get spatial building locally yet, but I can also try that route as well if you recommend it.
> 
> yeah the geo tests are currently not executed in CI because they require the geo extension to be built and neither me or @Maxxen had the time yet to fiddle with that yet :(

Is there a patch I could use or a specific duckdb hash I should aim for on the Spatial side @Maxxen? I have tried to build it locally with success on 1.4, but not on 1.5 (to align it with DuckLake dev). Many thanks!!

Alex-Monahan:
@Maxxen , I actually got it working locally! Claude was able to figure out patching on my behalf. This spatial stuff is pretty amazing that it compiles this many things and works this smoothly. I always appreciate it!!

@pdet , this is all set for a review whenever you have a chance (no pressure!!). I was able to run the spatial test locally with success, so I feel fine without that running in CI.

pdet:
Hi @Alex-Monahan I think this is looking pretty good, I'm missing a test where we drop a macro before flushing, I guess that should throw an error? Same if we create a macro, set sort, drop macro, create macro again (but now invalid for our sorting), then try to flush.

Alex-Monahan:
Good feedback! I added those tests for the unhappy paths with macros. The errors we got back seem nice and helpful, so no code changes needed, just those tests! Let me know what else would be good!

Alex-Monahan:
It looks like CI did not rerun - maybe due to the GitHub outage yesterday? I can push a dummy commit if desired!

pdet:
> It looks like CI did not rerun - maybe due to the GitHub outage yesterday? I can push a dummy commit if desired!

Sometimes this happens, an empty commit should be enough! :-)

pdet:
Thanks! Looks great!

PR #751, opened 2026-02-06
Fix view binding thread-safety with recursive_mutex

rgernhardt:
Use std::recursive_mutex to protect IsBound(), Bind(), and GetQuery() methods in DuckLakeViewEntry. This prevents:
1. Concurrent access from multiple threads causing memory corruption
2. Potential deadlock from recursive calls during view binding

Includes test for concurrent catalog scans that trigger view binding.

This fix is only needed in v.1.4 because view binding works very differently in v1.5. and this concurrency issue was fixed there.


Mytherin:
Thanks!

PR #731, opened 2026-01-29
Extends `DuckLakeMetadataManager` to be easier overridden

qsliu2017:
Redo #638 and #699.

Specifically, this PR:

1. add `DuckLakeMetadataManager::Register` for custom MetadataManager. #742
2. move queries on metadata to `DuckLakeMetadataManager::Query`/`DuckLakeMetadataManager::Execute`; `DuckLakeTransaction::Query` now only for queries within DuckDB.
3. adopt metadata queries to be compatible with Postgres, and the mainly changes include:
    1. replace `LIST({'key1': val1, 'key2': val2, ...})` with `WrapWithListAggregation`, and later parsed by `LoadTags`, `LoadInlinedDataTables`, etc. #748
    2. replace `MAX_BY` with `SELECT DISTINCT ON ... ORDER BY`. #754
    3. divide `GetOrphanFilesForCleanup` into `GetActiveFiles`(on metadata db) and get all files by DuckDB's read_blob. #755
    4. `ReadInlinedData*` returns the raw result from metadata db, and `TransformInlinedData` converts it to DuckDB's type. #753
    5. carefully rename columns to avoid duplicate. #757
    6. move metadata related variables replacement into `DuckLakeMetadataManager`, e.g., `{METADATA_CATALOG_NAME_LITERAL}`.

This PR will be helpful to implement custom metadata manager in embedded DuckDB.

<details>

<summary>Our use case in <a href="https://github.com/relytcloud/pg_ducklake">pg_ducklake</a></summary>
<p />
<p>
pg_ducklake is a PostgreSQL extension that embed DuckDB with DuckLake. We implement a custom `PgDuckLakeMetadataManager`, which executes the queries directly inside PG, instead of PG connection. And we plan to override some hot-path queries directly on PG heap table layer.
</p>
<p>
ducklake is such a fast growing community. It would be hard to rewrite every methods of `DuckLakeMetadataManager`. And since DuckDB is widely embedded in other DBs, DuckLake extension could bring native lakehouse capability to these systems. Hopefully our work would be the example implementation.
</p>

</details>

PR #753, opened 2026-02-09
Inline data handling

qsliu2017:
part of #731 

`ReadInlinedData*` now returns the raw result from metadata db. For unsupported type of metadata database, column might be stored in a different type. For example, `map` is stored as json bytes in PostgreSQL.
`TransformInlinedData` converts it to DuckDB's type, so that subclasses can override its action according to its stored type.

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -106,7 +106,7 @@ public:
 	virtual bool TypeIsNativelySupported(const LogicalType &type);
 
 	virtual string GetColumnTypeInternal(const LogicalType &column_type);
-	virtual string CastColumnToTarget(const string &stats, const LogicalType &type);
+	virtual string CastColumnToTarget(const string &column, const LogicalType &type);
 
 	DuckLakeMetadataManager &Get(DuckLakeTransaction &transaction);
 
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -969,8 +969,8 @@ string DuckLakeMetadataManager::CastStatsToTarget(const string &stats, const Log
 	return stats;
 }
 
-string DuckLakeMetadataManager::CastColumnToTarget(const string &stats, const LogicalType &type) {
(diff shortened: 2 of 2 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -969,6 +969,10 @@ string DuckLakeMetadataManager::CastStatsToTarget(const string &stats, const Log
 	return stats;
 }
 
+string DuckLakeMetadataManager::CastColumnToTarget(const string &stats, const LogicalType &type) {
+	return stats + "::" + type.ToString();
+}
```
pdet: Why is this called stats? It seems is just an expression, no?

src/include/storage/ducklake_metadata_manager.hpp
```diff
@@ -196,20 +197,22 @@ class DuckLakeMetadataManager {
...
+	                                                          const vector<string> &columns_to_read);
+	virtual unique_ptr<QueryResult> ReadInlinedDataDeletions(DuckLakeSnapshot start_snapshot,
+	                                                         DuckLakeSnapshot end_snapshot,
+	                                                         const string &inlined_table_name,
+	                                                         const vector<string> &columns_to_read);
+	virtual unique_ptr<QueryResult> ReadAllInlinedDataForFlush(DuckLakeSnapshot snapshot,
+	                                                           const string &inlined_table_name,
+	                                                           const vector<string> &columns_to_read);
+	virtual shared_ptr<DuckLakeInlinedData> TransformInlinedData(QueryResult &result,
+	                                                             const vector<LogicalType> &expected_types);
```
pdet, line 214: I guess that adding the `expected_types` are for the subclasses, but maybe you can have a detailed explanation on the PR description about this change?

qsliu2017: done

src/storage/ducklake_metadata_manager.cpp
qsliu2017: addressed

pdet:
Thanks!

PR #755, opened 2026-02-09
split GetOrphanFilesForCleanup

qsliu2017:
part of #731 

split `GetOrphanFilesForCleanup` into `GetActiveFiles`(on metadata db) and get all files by DuckDB's `read_blob`.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -3937,12 +3937,8 @@ FROM {METADATA_CATALOG}.ducklake_files_scheduled_for_deletion
 	}
 	return result;
 }
-vector<DuckLakeFileForCleanup> DuckLakeMetadataManager::GetOrphanFilesForCleanup(const string &filter,
-                                                                                 const string &separator) {
-	auto query = R"(SELECT filename
-FROM read_blob({DATA_PATH} || '**')
-WHERE suffix(filename, '.parquet')
-AND filename NOT IN (
+unordered_set<string> DuckLakeMetadataManager::GetActiveFiles(const string &separator) {
+	string query = R"(
 SELECT REPLACE(
            CASE
                WHEN NOT file_relative THEN file_path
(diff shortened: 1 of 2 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -3542,11 +3542,9 @@ FROM {METADATA_CATALOG}.ducklake_files_scheduled_for_deletion
 	}
 	return result;
 }
-vector<DuckLakeFileForCleanup> DuckLakeMetadataManager::GetOrphanFilesForCleanup(const string &filter,
-                                                                                 const string &separator) {
-	auto query = R"(SELECT filename
-FROM read_blob({DATA_PATH} || '**')
-WHERE filename NOT IN (
+
+unordered_set<string> DuckLakeMetadataManager::GetActiveFiles(const string &separator) {
```
pdet, line 3940: In general, we should avoid splitting up queries, as these will increase round-trips to the DBMS catalog, which can decrease performance.

pdet (commented):
Thanks for the PR. I have one comment!

src/storage/ducklake_metadata_manager.cpp
qsliu2017: @pdet 

`read_blob({DATA_PATH} || '**')` is a duckdb specific function.

actually I think this split will eventually reduce round-trips:
- original query: contains read_blob, must be executed in duckdb. duckdb reads each related catalog table by one `select * from table`. # of round-trips == # of catalog table in query
- new query: can be executed as one query in catalog. # of round-trips == 1

qsliu2017:
Hi @pdet , just checking in on this PR when you have a moment. Let me know if there's anything else you need from my side. Thanks!

pdet:
> Hi @pdet , just checking in on this PR when you have a moment. Let me know if there's anything else you need from my side. Thanks!

Hi, I'll have a look this afternoon, but I'm holding off on merges for a bit for the release!

pdet:
Closing this as it is superseeded by https://github.com/duckdb/ducklake/pull/1196

PR #759, opened 2026-02-10
Merge v1.4-andium into main

Mytherin:


pdet:
It seems that the failing test is just a missing skip?

Mytherin:
Yeah, defining `skip_error_messages` overrides them. I'm guessing this test was previously skipped by the old set of `skip_error_messages`. Should pass now hopefully.

PR #762, opened 2026-02-10
Add override for GetName() for ducklake extension operators

rgernhardt:
Add override for GetName() for ducklake extension operators compact and flush.
The motivation is so that custom optimizer extensions can recognize these operators.


pdet:
Thanks!

PR #765, opened 2026-02-12
Fix filter pushdown

timbess:
This is a continuation of: https://github.com/duckdb/duckdb/pull/20816.

Optimizations like late materialization create more complex query plans that allow certain join branches to skip filters critical to limiting the number of unnecessary files pulled from the source. I'm not sure if `Bind()` is the right place to update that config option, but I confirmed that this fixes the issue locally. I'll figure out how to write a test for this, but I just want to be sure this approach is good while I'm working on it.

### DIFF
--- a/src/include/storage/ducklake_multi_file_list.hpp
+++ b/src/include/storage/ducklake_multi_file_list.hpp
@@ -36,6 +36,10 @@ public:
 	                                                const vector<column_t> &column_ids,
 	                                                TableFilterSet &filters) const override;
 
+	unique_ptr<MultiFileList> ComplexFilterPushdown(ClientContext &context, const MultiFileOptions &options,
+	                                                MultiFilePushdownInfo &info,
+	                                                vector<unique_ptr<Expression>> &filters) const override;
+
 	vector<OpenFileInfo> GetAllFiles() const override;
 	FileExpandResult GetExpandResult() const override;
 	idx_t GetTotalFileCount() const override;
@@ -63,6 +67,7 @@ private:
 	void GetFilesForTable() const;
 	void GetTableInsertions() const;
 	void GetTableDeletions() const;
+	void AddFilterToPushdownInfo(FilterPushdownInfo &pushdown_info, column_t column_id, unique_ptr<TableFilter> filter) const;
 
 private:
 	mutable mutex file_lock;
--- a/src/storage/ducklake_multi_file_list.cpp
+++ b/src/storage/ducklake_multi_file_list.cpp
@@ -11,14 +11,7 @@
 #include "duckdb/main/query_profiler.hpp"
 #include "duckdb/main/secret/secret_manager.hpp"
 #include "duckdb/optimizer/filter_combiner.hpp"
-#include "duckdb/parser/expression/function_expression.hpp"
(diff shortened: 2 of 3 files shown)


Mytherin:
Thanks for the PR!

DuckLake doesn't rely on hive partitioning internally - it just writes files out in hive partitioned format by convention for easier operability with other tools. I think this is the wrong way to go about solving this as this change makes hive partitioning a hard requirement for partitioned DuckLake tables. Users might not want to write files in a hive partitioned manner, e.g. if they are encrypted and the partitioning keys contain sensitive information.

timbess:
Good point, I was sorta diffing the implementation against the superclass and had a bit of tunnel vision for the issue I'm trying to solve. So basically, if I update the `FilterPushdownInfo` instead, it'll use DuckLake's built-in mechanisms to filter the file list and work with alternative table formats as well. I believe I got a version of this working, let me polish it up and update the PR.

timbess:
Ok, I believe I did it correctly this time, thank you for the tips and patience!

redox:
I didn't check all the implication of this diff, but we stumbled upon the same issue and could validate this branch fixes it.

I tested with this test:
```
diff --git a/test/sql/stats/topn_file_pruning.test b/test/sql/stats/topn_file_pruning.test
index 29c9e56a..a8dca8f1 100644
--- a/test/sql/stats/topn_file_pruning.test
+++ b/test/sql/stats/topn_file_pruning.test
@@ -86,3 +86,25 @@ query II
 EXPLAIN ANALYZE SELECT * FROM ducklake.events ORDER BY timestamp DESC NULLS FIRST LIMIT 1
 ----
 analyzed_plan	<REGEX>:.*1,800 rows.*
+
+# Test Top-N pruning with partitioned table - should only read files from matching partitions
+statement ok
+CREATE TABLE ducklake.partitioned_events(part_key INTEGER, timestamp TIMESTAMP);
+
+statement ok
+ALTER TABLE ducklake.partitioned_events SET PARTITIONED BY (part_key);
+
+# Insert 1 row per partition with different timestamps
+statement ok
+INSERT INTO ducklake.partitioned_events VALUES (1, TIMESTAMP '2026-01-01 00:00:00');
+
+statement ok
+INSERT INTO ducklake.partitioned_events VALUES (2, TIMESTAMP '2026-01-02 00:00:00');
+
+# Top-N query filtering to a specific partition - should only read files from that partition
+# The main TABLE_SCAN should only read 1 file from the matching partition
+query II
+EXPLAIN ANALYZE SELECT * FROM ducklake.partitioned_events WHERE part_key = 2 ORDER BY timestamp DESC LIMIT 1
+----
+analyzed_plan	<REGEX>:.*TABLE_SCAN.*Total Files Read: 1.*TABLE_SCAN.*Total Files Read: 1.*
+
```

with `main` it currently reads the 2 files instead of 1.

timbess:
@Mytherin Thanks for the help so far, please let me know if there's anything else I need to do to get this mergable.

buenaventure:
@pdet is there any chance to get this into the upcoming release?

Mytherin:
Thanks!

PR #769, opened 2026-02-16
Avoid linker conflicts with delta extension

ywelsch:
Note that this should target the DuckDB 1.5 branch (does not seem to exist on this repo)

Delta has the same definitions in the same scope.

The problem goes away when correctly scoping the MetadataFunctionInit and MetadataFunctionExecute with static (to the compilation unit), and avoiding the duplicate definitions of BaseMetadataFunction (by renaming to DuckLakeMetadataFunction and DeltaMetadataFunction, respectively).

See also https://github.com/duckdb/duckdb-delta/pull/283


benfleis:
@ywelsch I asked the ducklake experts to have a glance; my quick scan says these failures don't look related.

pdet:
Thanks for the PR. Can you merge with `main`?

PR #770, opened 2026-02-16
Update Geometry types support for v1.5

Maxxen:
This PR updates the `GEOMETRY` type support for DuckDB v1.5. This fixes a bunch of limitations and special-cases with the geometry type in DuckLake.

- `GEOMETRY` columns now return proper DuckDB geo stats, e.g. `SELECT stats(my_geom_columns)` now works. This means we now have geometry filter pushdown in DuckLake!
- `GEOMETRY` types can now be nested in structs/lists/maps
- `GEOMETRY` is now supported when inlining.
- All tests in `test/sql/types/geo` now run _without_ requiring the spatial extension to be loaded, making testing/building/maintenance of the DuckLake extension easier.

Also fixes a bug where we did not convert DuckLake `MAP` stats to DuckDB stats properly.

### DIFF
--- a/src/common/ducklake_types.cpp
+++ b/src/common/ducklake_types.cpp
@@ -76,8 +76,6 @@ static string ToStringBaseType(const LogicalType &type) {
 	throw InvalidInputException("Failed to convert DuckDB type to DuckLake - unsupported type %s", type);
 }
 
-
-
 bool DuckLakeTypes::RequiresCast(const LogicalType &type) {
 	// There are no types that requires casts as of DuckDB v1.5
 	return false;


pdet:
Hey @Maxxen, are geo types already working as inlined data?

Maxxen:
With this PR they do!

pdet:
Nice, can you resolve the conflict, CI should be good afterwards

Maxxen:
@pdet Huh I thought we inlined values as strings? Do you know why Postgres complains about the type missing/how I mark it as not supported in Postgres?

pdet:
> @pdet Huh I thought we inlined values as strings? Do you know why Postgres complains about the type missing/how I mark it as not supported in Postgres?

yeah only not supported values are stored as a string, you can flag that in the postgres metadata manager, i.e.,
`bool PostgresMetadataManager::TypeIsNativelySupported(const LogicalType &type)`

Maxxen:
@pdet I think this should be ready to merge now once the last CI deploy step is green!

PR #772, opened 2026-02-17
Fix for variant order switcharoo

pdet:


pdet:
@Tishj FYI I'm skipping a variant test here that might be a potential issue

PR #775, opened 2026-02-17
Making inlining default to ducklake.

pdet:
This PR makes data inlining the default behavior in DuckLake. Small inserts/deletes (up to 10 rows) are now stored directly in the catalog rather than creating separate Parquet files.

Due to this change, all tests are now also executed with inlining by default, so this PR implements many features and fixes related to inlining.

1.  ALTER TABLE Support with Inlined Data. We now support `ADD COLUMN`,`DROP COLUMN`, `RENAME TABLE`, `ALTER COLUMN TYPE`,`SET/DROP NOT NULL`.
2.  Expression Map Support in Inlined Data Reader.
3.  SQLite and Postgres now also support more data types and edge cases. In Postgres, varchars are now stored as `BYTEA` so we can support null bytes in them.

For testing, we adjusted tests whenever possible to make sure they can work with both inlining and not inlining, as in producing files in the correct numbers, in the same place, checking snapshots, and time-traveling. Due to the change in default behavior, a new CI was also added to ensure all tests are executed without using inlining.

### DIFF
--- a/test/configs/attach_ducklake.json
+++ b/test/configs/attach_ducklake.json
@@ -626,7 +626,7 @@
       "reason": "FIXME: Variant Bugs?",
       "paths": [
           "test/parquet/variant/variant_comparison.test",
-        " test/sql/storage/types/variant/variant_shredding_omit_untyped.test"
+        "test/sql/storage/types/variant/variant_shredding_omit_untyped.test"
       ]
     }
   ]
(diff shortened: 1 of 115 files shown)


test/configs/attach_ducklake.json
```diff
@@ -637,5 +637,24 @@
...
+    {
+      "reason": "FIXME: inline of GEOMETRY not yet supported",
+      "paths": [
+        "test/geoparquet/no_spatial.test",
+        "test/geoparquet/mixed.test",
+        "test/sql/types/geo/geometry_shred_empty.test",
+        "test/sql/types/geo/geometry.test",
+        "test/sql/types/geo/geometry_wkb.test",
+        "test/sql/types/geo/geometry_shred_fetch.test"
+      ]
```
pdet, line 657: @Maxxen FYI, in case this gets in before your geometry

PR #780, opened 2026-02-18
Respect `SET SORTED BY` for inserts

Alex-Monahan:
This removes another restriction from #642 - sorting can be configured to occur during inserts. There is an option to enable/disable this behavior (`sort_on_insert`), defaulting to `true` / enabled. It may be worth disabling sort on insert if insertion speed is the primary concern (since inline flush and compaction operations could still trigger sorting, but possibly in the background or at a more convenient time). 

The code couldn't be perfectly reused, since inserts operate on a physical plan rather than a logical plan like compaction and inline flush. However, I was able to reuse most of the logic thanks to splitting things out into helper functions.

There is also some slightly complex behavior when inlining. Data that is inserted inline will obey the `sort_on_insert` flag, except in the case that data that is inserted is larger than the inlining limit, in which case it will sort on insert anyway (to mirror the behavior of inline flush always sorting on insert). Let me know if this is not what we want and I can remove this piece and we can insert without sorting in that case instead.

I also don't do any deduplication of orderings in the plan, so if someone adds their own `ORDER BY` on their `INSERT` statement and also uses `SET SORTED BY`, I am guessing that the sort will happen twice. If that is an issue, let me know and I can see if I can learn how to handle that situation!

I did use Claude while building this, but I have reviewed the code and forced some refactoring when I didn't like Claude's approach and manually added more tests. Super open to feedback!

### DIFF
--- a/test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
+++ b/test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
@@ -1,5 +1,5 @@
 # name: test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
-# description: if column names have changed, inlining flush should fail until sorted by valid column names
+# description: if column names have changed, insert and flush should fail until sorted by valid column names
 # group: [sorted_table]
 
 require ducklake
@@ -14,7 +14,7 @@ test-env DATA_PATH __TEST_DIR__
 statement ok
 ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_inlining_flush_data_renamed', DATA_INLINING_ROW_LIMIT 10)
 
-# SET SORTED BY then rename columns - flush should error on missing columns
+# SET SORTED BY then rename columns - insert should error on missing columns
 statement ok
 CREATE TABLE ducklake.renamed_columns_test (unique_id INTEGER, sort_key_1 INTEGER, sort_key_2 VARCHAR);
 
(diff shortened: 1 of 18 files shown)


Alex-Monahan:
Edit: Fixed up the test to use a fresh DuckLake! Ready for review!

jprafael:
@Alex-Monahan does this work cover `MERGE INTO` operations?

I'm running into an issue where ducklake is attempting a _full-table_ hash join (spilling hundreds of GBs to tmp storage) instead of using partition information to run individual joins per partition, and using sort information to use a merge join instead of a hash join. Is this simply not supported yet?

Alex-Monahan:
> @Alex-Monahan does this work cover `MERGE INTO` operations?
> 
> I'm running into an issue where ducklake is attempting a _full-table_ hash join (spilling hundreds of GBs to tmp storage) instead of using partition information to run individual joins per partition, and using sort information to use a merge join instead of a hash join. Is this simply not supported yet?

This feature is separate actually! I think pushdown of partitions for merges would be awesome. This PR would only help performance if pushdown of non-partition columns were supported also. 

Could you start a separate issue with a test case that reproduces the behavior? Thank you!

test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
```diff
@@ -28,25 +28,32 @@ ALTER TABLE ducklake.renamed_columns_test RENAME COLUMN sort_key_1 TO sort_key_1
 statement ok
 ALTER TABLE ducklake.renamed_columns_test RENAME COLUMN sort_key_2 TO sort_key_2_changed;
 
-# Insert data (will be inlined initially)
-statement ok
+# Insert should fail because the sort expressions reference old column names
+statement error
```
Mytherin: I don't think this should fail - can we not update the sorted expression during the rename?

Mytherin (commented):
Thanks! Looks good to me - I left one comment

test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
Alex-Monahan: Ah, very cool idea! I had not even considered that. I will have a look. Thanks for the review Mark!

guillesd:
Hey @Alex-Monahan ! Will you have time to take a look at the comment and merge conflicts? It would be nice to get this one for the release

Alex-Monahan:
> Hey @Alex-Monahan ! Will you have time to take a look at the comment and merge conflicts? It would be nice to get this one for the release

Howdy!

I can have a look, but it may be next week. Is there a date I should shoot for? I may be able to rearrange things.

guillesd:
Hey @Alex-Monahan I merged the conflicts and gave the sorted expression on column rename thing a go. @pdet can you have a look at that last commit?

test/sql/sorted_table/data_inlining_flush_sorted_renamed.test
pdet: @guillesd I believe Mark's comment here still needs to be worked on.

Mytherin: I don't think that's really a dealbreaker, if the rest looks good to you we can merge it

guillesd: I think I did address it no? https://github.com/duckdb/ducklake/pull/780/commits/3786770909cae056cf6c0b9d1787fb420e33477e
I also changed the tests to instead of expect failure expect success

pdet:
Thanks @Alex-Monahan and @guillesd LGTM!

Alex-Monahan:
> Thanks @Alex-Monahan and @guillesd LGTM!

Thank you all for carrying this across the finish line for me!!! I really appreciate it.

guillesd:
Thanks @Alex-Monahan ! this was very nice work, we'll do a shut out in the announcement for 1.0!

PR #786, opened 2026-02-19
default catalog_type to duckdb

rgernhardt:
Small change to ducklake_settings function to default to "duckdb" catalog type if catalog is not one of the known alternatives (postgres, sqllite)

### DIFF
--- a/src/functions/ducklake_settings.cpp
+++ b/src/functions/ducklake_settings.cpp
@@ -21,14 +21,15 @@ static unique_ptr<FunctionData> DuckLakeSettingsBind(ClientContext &context, Tab
 	auto result = make_uniq<MetadataBindData>();
 	vector<Value> row_values;
 
-	// catalog_type: default to "duckdb" if not explicitly set (DuckDB is the default metadata storage)
+	// catalog_type: normalize the metadata type to a user-friendly name
+	// unknown types (including empty) default to "duckdb" since DuckDB is the default metadata storage
 	auto catalog_type = ducklake_catalog.MetadataType();
-	if (catalog_type.empty()) {
-		catalog_type = "duckdb";
-	} else if (catalog_type == "postgres_scanner") {
+	if (catalog_type == "postgres_scanner") {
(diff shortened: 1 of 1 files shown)


src/functions/ducklake_settings.cpp
```diff
@@ -21,14 +21,15 @@ static unique_ptr<FunctionData> DuckLakeSettingsBind(ClientContext &context, Tab
...
 	auto catalog_type = ducklake_catalog.MetadataType();
-	if (catalog_type.empty()) {
-		catalog_type = "duckdb";
-	} else if (catalog_type == "postgres_scanner") {
+	if (catalog_type == "postgres_scanner") {
 		catalog_type = "postgres";
 	} else if (catalog_type == "sqlite_scanner") {
 		catalog_type = "sqlite";
+	} else {
+		catalog_type = "duckdb";
```
carlopi: What would happen for say `mysql_scanner` ? That's not properly first class, but I could see that being used, and getting back `duckdb` would mean losing information.

Maybe it's just cleaner to add some extra aliases? Unsure.

rgernhardt: I see. That could be ok as well, I can alias motherduck to duckdb here, if we're ok with having that in the code here.

Like 
```
 if (catalog_type == "postgres_scanner") {
      catalog_type = "postgres";
  } else if (catalog_type == "sqlite_scanner") {
      catalog_type = "sqlite";
  } else if (catalog_type.empty() || catalog_type == "motherduck" || catalog_type == "md_server") {
      catalog_type = "duckdb";
  } 
 // else pass through
```

pdet: I think aliasing MotherDuck is preferable. I guess this eventually needs to move to the metadata class, which could be overridden. But for now, this would look fine to me!

rgernhardt: Done!

pdet:
Thanks

PR #793, opened 2026-02-23
Add missing spaces

szarnyasg:


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -203,7 +203,7 @@ void DuckLakeMetadataManager::ExecuteMigration(string migrate_query, bool allow_
 	}
 	auto result = transaction.Query(migrate_query);
 	if (result->HasError()) {
-		result->GetErrorObject().Throw("Failed to migrate DuckLake from v0.2 to v0.3:");
+		result->GetErrorObject().Throw("Failed to migrate DuckLake from v0.2 to v0.3: ");
```
szarnyasg, line 206: @pdet more importantly than the missing space character, this also gets printed when I try to migrate from DuckLake **v0.3 to v0.4**:

```text
DuckDB v1.5.0-dev7678 (Development Version, f961e485d6)
Enter ".help" for usage hints.
memory D install ducklake;
memory D ATTACH 'ducklake:https://datasets.ldbcouncil.org/snb-interactive-v1-ducklake/sf1.ducklake' (AUTOMATIC_MIGRATION);
Invalid Input Error:
Failed to migrate DuckLake from v0.2 to v0.3:Cannot execute statement of type "CREATE" on database "__ducklake_metadata_sf1" which is attached in read-only mode!
```

Issue #791, opened 2026-02-23
ducklake_add_data_files on partitioned table fails when partitioned column is also contained in Parquet file


gijshendriksen:
### What happens?

We have a set of Hive-partitioned Parquet files, in which the partition value is also included in the Parquet files themselves (so these columns can still be used after downloading individual files). When adding these Hive-partitioned files to a DuckLake table with the same partition columns, the `ducklake_add_data_files` fails. Specifically, when a column is found in the Parquet metadata, the column is no longer considered as a possible partition column.

I'm not sure what the expected behaviour would be, as I don't think it's standard practice to have a column in both the Hive path and in the Parquet files themselves. In theory, if this happens we could check whether the value in the Parquet file corresponds to the value in the partition column, and fail otherwise. I'd also be okay with adding a more informative error message, though.

Note that the issue is easily circumvented by removing the partition key from the table, since the Parquet file stats can still be used to prune the file.

### To Reproduce

First, make a Hive-partitioned directory:
```bash
mkdir 'partition_key=value'
```
Then, run the following SQL:
```sql
-- Add file with same column as the Hive partition
COPY (SELECT 'value' AS partition_key, 1 AS value) TO 'partition_key=value/data.parquet';

-- Create DuckLake table with same Hive partitioning
ATTACH 'ducklake:test.ducklake' AS dl;

CREATE TABLE dl.test AS FROM 'partition_key=value/data.parquet' WITH NO DATA;
ALTER TABLE dl.test SET PARTITIONED BY (partition_key);

-- Import the file
CALL ducklake_add_data_files('dl', 'test', 'partition_key=value/data.parquet');
```

This results in the following error:
```
Invalid Input Error:
File "partition_key=value/data.parquet" contains an invalid partition value for the table configuration.
```

Note that we can circumvent the issue by dropping the partition column:
```sql
ALTER TABLE dl.test RESET PARTITIONED BY;
CALL ducklake_add_data_files('dl', 'test', 'partition_key=value/data.parquet');
```

### OS:

Linux x86_64

### DuckDB Version:

1.4.3

### DuckLake Version:

41255d47 (core_nightly)

### DuckDB Client:

CLI

### Full Name:

Gijs Hendriksen

### Affiliation:

Radboud University

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a nightly build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

[...]

PR #798, opened 2026-02-24
Skip hive partitioned columns from parquet files

pdet:
If we add a parquet file to a hive partition scheme, where this file has the partition column in the file, we skip it during mapping.

Fix: https://github.com/duckdb/ducklake/issues/791


gijshendriksen:
Thanks for the quick fix @pdet! I've verified the fix on our live data as well, and it works perfectly :)

PR #801, opened 2026-02-24
fix: avoid parse-AST-serialize roundtrip in DuckLakeViewEntry::ToSQL() (#788)

### LINKED ISSUE #788: duckdb_views() is ~70x slower than duckdb_tables() due to unnecessary SQL parsing in DuckLakeViewEntry::ToSQL()
## Problem

`duckdb_views()` takes **1.4 seconds** on a DuckLake catalog with ~1,900 views, while `duckdb_tables()` takes **0.02 seconds** on the same catalog with ~1,100 tables. The slowdown is consistent on every call — there is no caching.

## Reproduction Query

```sql
SELECT sql FROM duckdb_views()
WHERE database_name = 'my_ducklake'
  AND schema_name = 'main'
  AND view_name = 'my_view';
```

The `WHERE` filters provide no benefit — `duckdb_views()` scans all views and calls `ToSQL()` on each one during init, before any filtering is applied.

## Profiling Evidence

Ran `EXPLAIN ANALYZE` on a production DuckLake instance (1,882 views, 624 tables, 9,914 columns):

| Query | Server-side Time | Rows |
|---|---|---|
| `duckdb_views()`  | **1.43s** | 1,882 |
| `duckdb_views()`  | **1.37s** | 1,882 |
| `duckdb_views()`  | **1.38s** | 1,882 |
| `duckdb_tables()` | **0.02s** | 1,100 |
| `duckdb_columns()` | **0.10s** | 80,889 |
| `duckdb_views()` after DDL (cold cache) | **1.38s** | 1,882 |
| `duckdb_views()` after DDL (warm cache) | **1.39s** | 1,882 |

### Full EXPLAIN ANALYZE output

```sql
EXPLAIN ANALYZE SELECT sql FROM duckdb_views()
WHERE database_name = 'my_ducklake'
  AND schema_name = 'main'
  AND view_name = 'my_view';
```
```
┌────────────────────────────────────────────────┐
│               Total Time: 1.43s               │
└────────────────────────────────────────────────┘
┌───────────────────────────┐
│           QUERY           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      EXPLAIN_ANALYZE      │
│           0 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│             #3            │
│           0 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           FILTER          │
│  ((database_name =        │
│  'my_ducklake') AND       │
│  (view_name =             │
│  'my_view') AND           │
│  (schema_name = 'main'))  │
│           0 rows          │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│        1,882 rows         │
│          (0.00s)          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         TABLE_SCAN        │
│         Function:         │
│        DUCKDB_VIEWS       │
│        1,882 rows         │
│          (1.43s)          │
└───────────────────────────┘
```

All 1.43s is spent in the `DUCKDB_VIEWS` TABLE_SCAN (which calls `ToSQL()` → `ParseSelectStatement()` on all 1,882 views). The FILTER node that applies the `WHERE` clause runs in 0.00s — but by then the expensive work is already done.

## Root Cause

`duckdb_views()` calls `view.ToSQL()` for every view (column 11). In `DuckLakeViewEntry`, this triggers:

```
ToSQL()
  → GetInfo()                          // creates a new CreateViewInfo each call
    → ViewCatalogEntry::GetInfo()      // copies fields, query = nullptr (DuckLake is lazy)
    → ParseSelectStatement()           // Full SQL parse from raw string!
      → Parser::ParseQuery(query_sql)  // Lexer + parser → AST
  → CreateViewInfo::ToString()         // Serialize AST back to string
```

This is a **parse→AST→serialize roundtrip to nowhere** — DuckLake already has the raw SQL string in `query_sql`, but `ToSQL()` parses it into an AST only to serialize it back to a string.

Additionally, **the parsed AST is not cached**. `GetInfo()` creates a new `CreateViewInfo` copy each time, sets `query` on that temporary copy, and discards it. The next call re-parses from scratch.

At ~0.7ms per view × 1,882 views = ~1.4s.

### Why `duckdb_tables()` doesn't have this problem

For tables, `GetInfo()` returns `CreateTableInfo` with eagerly loaded column definitions — no parsing needed. `ToString()` just serializes the column list (simple string concatenation).

### Why DuckLake views are lazy

When DuckLake loads views from the metadata catalog, it only stores the raw SQL string and does NOT parse it ([ducklake_catalog.cpp:396-407](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_catalog.cpp#L396-L407)):

```cpp
auto create_view_info = make_uniq<CreateViewInfo>(schema_entry, view.name);
create_view_info->aliases = view.column_aliases;
// NOTE: create_view_info->query is NOT set — only the raw SQL string is stored
auto view_entry = make_uniq<DuckLakeViewEntry>(..., std::move(view.sql), ...);
```

This lazy approach is correct for startup performance, but `ToSQL()` defeats it.

## Proposed Fix

Override `ToSQL()` in `DuckLakeViewEntry` to construct the CREATE VIEW string directly from `query_sql`, avoiding the parse roundtrip entirely:

```cpp
string DuckLakeViewEntry::ToSQL() const {
    string result = "CREATE VIEW ";
    if (schema.name != DEFAULT_SCHEMA) {
        result += KeywordHelper::WriteOptionallyQuoted(schema.name) + ".";
    }
    result += KeywordHelper::WriteOptionallyQuoted(name);
    if (!aliases.empty()) {
        result += "(";
        for (idx_t i = 0; i < aliases.size(); i++) {
            if (i > 0) {
                result += ", ";
            }
            result += KeywordHelper::WriteOptionallyQuoted(aliases[i]);
        }
        result += ")";
    }
    result += " AS ";
    result += query_sql;
    result += ";";
    return result;
}
```

This is safe because `query_sql` is always the output of `SelectStatement::ToString()` — either set at view creation time ([ducklake_schema_entry.cpp:144](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_schema_entry.cpp#L144): `info.query->ToString()`) or loaded from the metadata database (which stored the same serialized form).

**Expected improvement**: `duckdb_views()` should drop from ~1.4s to ~0.02s (similar to `duckdb_tables()`), making it ~70x faster on this catalog.

## Impact

This affects any DuckLake deployment with a large number of views. Our service uses `duckdb_views()` to look up view SQL during DDL operations (drop, rename, create-or-replace). With ~1,900 views, each call adds 1.4s of pure parsing overhead. Operations like rename-table that call it 2-3 times accumulate 4+ seconds of unnecessary latency.

## Environment

- DuckDB version: v1.3.x (latest DuckLake main)
- DuckLake metadata backend: PostgreSQL
- Catalog size: 624 tables, 1,882 views, 9,914 columns

suresh-summation:
## Summary

Fixes #788

- `duckdb_views()` was ~70x slower than `duckdb_tables()` because `DuckLakeViewEntry::ToSQL()` called `GetInfo()->ToString()`, which parsed `query_sql` into a full AST and serialized it back on **every invocation** (~0.7ms per view, no caching). On a catalog with ~1,900 views this caused ~1.4s latency.
- Override `ToSQL()` to directly construct the `CREATE VIEW` statement from already-stored fields (`name`, `aliases`, `query_sql`), bypassing the parser entirely. `query_sql` is guaranteed to be pre-serialized from `SelectStatement::ToString()` at view-creation time.
- `GetInfo()` and `GetQuery()` are unchanged — they still parse lazily when the AST is actually needed (e.g. binding, `Copy()`, `AlterEntry()`).

## Test plan

- [ ] Simple view SQL output matches expected `CREATE VIEW name AS query;`
- [ ] View with explicit column aliases renders correctly: `CREATE VIEW name (a, b) AS ...`
- [ ] Complex view (filter + join) serializes correctly
- [ ] View whose name is a reserved keyword gets double-quoted (e.g. `"select"`)
- [ ] After `ALTER VIEW ... RENAME TO`, `ToSQL()` reflects the new name
- [ ] View with a subquery serializes correctly

### DIFF
--- a/src/storage/ducklake_view_entry.cpp
+++ b/src/storage/ducklake_view_entry.cpp
@@ -59,9 +59,6 @@ unique_ptr<CreateInfo> DuckLakeViewEntry::GetInfo() const {
 }
 
 string DuckLakeViewEntry::ToSQL() const {
-	// Directly construct the CREATE VIEW statement from the stored query_sql,
-	// avoiding the expensive parse->AST->serialize roundtrip that GetInfo()->ToString() performs.
-	// query_sql is guaranteed to be pre-serialized from SelectStatement::ToString().
 	string result = "CREATE VIEW ";
 	result += KeywordHelper::WriteOptionallyQuoted(name);
 	if (!aliases.empty()) {
--- a/test/sql/issues/view_tosql_performance.test
+++ b/test/sql/issues/view_tosql_performance.test
@@ -14,7 +14,7 @@ test-env DATA_PATH __TEST_DIR__
 statement ok
 ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_view_tosql');
 
-# Simple view - verify sql column in duckdb_views()
+# simple view
 statement ok
 CREATE VIEW ducklake.simple_view AS SELECT 42 AS answer
 
@@ -23,13 +23,12 @@ SELECT sql FROM duckdb_views() WHERE view_name='simple_view' AND database_name='
 ----
 CREATE VIEW simple_view AS SELECT 42 AS answer;
 
-# Ensure the view is still queryable
 query I
 SELECT answer FROM ducklake.simple_view
 ----
 42
 
-# View with explicit column aliases
+# view with explicit column aliases
 statement ok
 CREATE VIEW ducklake.aliased_view(x, y) AS SELECT 1, 2
 
@@ -43,7 +42,7 @@ SELECT x, y FROM ducklake.aliased_view
 ----
 1	2
 
-# View with a more complex query (join + filter)
+# view with a filter
 statement ok
 CREATE TABLE ducklake.t1 (id INTEGER, val VARCHAR)
 
@@ -64,7 +63,7 @@ SELECT * FROM ducklake.complex_view
 2	world
 3	foo
 
-# View with a name that is a keyword - must be quoted
+# view with a keyword name must be quoted
 statement ok
 CREATE VIEW ducklake."select" AS SELECT 99 AS n
 
@@ -78,7 +77,6 @@ SELECT n FROM ducklake."select"
 ----
 99
 
-# Verify all views appear correctly in duckdb_views()
 query I
 SELECT view_name FROM duckdb_views() WHERE database_name='ducklake' ORDER BY view_name
 ----
@@ -87,7 +85,7 @@ complex_view
 select
 simple_view
 
-# After renaming a view, ToSQL() must reflect the new name
+# after rename, sql reflects the new name
 statement ok
 ALTER VIEW ducklake.simple_view RENAME TO renamed_view
 
@@ -101,13 +99,12 @@ SELECT answer FROM ducklake.renamed_view
 ----
 42
 
-# The old name should no longer appear
 query I
 SELECT count(*) FROM duckdb_views() WHERE view_name='simple_view' AND database_name='ducklake'
 ----
 0
 
-# View with a subquery
+# view with a subquery
 statement ok
 CREATE VIEW ducklake.subquery_view AS SELECT * FROM (SELECT id * 2 AS doubled FROM t1) sq WHERE doubled < 6
 


suresh-summation:
@pdet  The query still takes the same amount of time, especially when executed immediately after a DDL operation. Any thoughts on what might be causing this?

pdet:
Your analysis looked good. At a glance, I don't have much to add. I don't think this is causing frequent catalog round-trips as is (which is one of the main bottlenecks for DuckLake in systems like Postgres).

I could recommend using a profiler.

suresh-summation:
> Your analysis looked good. At a glance, I don't have much to add. I don't think this is causing frequent catalog round-trips as is (which is one of the main bottlenecks for DuckLake in systems like Postgres).
> 
> I could recommend using a profiler.


@pdet 
I was wrongly using debug build. The release build is lot quicker. I guess the change made significant performance difference. Would help to get this build out. Thanks for your help

The following numbers are from running it in my local. (Previously it used to take 1.2 sec in my local during cold cache scenario)

  ════════════════════════════════════════════════════════════════
    DuckLake RELEASE — Complex Views Post-DDL Timing
    Catalog: 2000 complex views  |  500 tables
  ════════════════════════════════════════════════════════════════

    Pre-DDL
      Query 1 (cold start)   : 1ms
      Query 2 (warm cache)   : 1ms

    DDL: CREATE TABLE repro_test

    Post-DDL
      Query 1 (cache busted) : 19ms
      Query 2 (warm cache)   : 1ms

    ──────────────────────────────────────────────
                               Pre-DDL   Post-DDL
    Query 1 (cold/reload)        1ms       19ms
    Query 2 (warm cache)         1ms        1ms
  ════════════════════════════════════════════════════════════════
  
  
  | Scenario         | System v1.4.4 (unpatched) | Our release build (patched) |
|------------------|----------------------------|------------------------------|
| Cold start       | 1121 ms                    | 2 ms                         |
| Warm cache       | 277 ms                     | 1 ms                         |
| Post-DDL reload  | 1122 ms                    | 19 ms                        |
| Post-DDL warm    | 252 ms                     | 1 ms                         |

pdet:
That looks good and is more in line with what I was expecting. Can you just resolve the conflict that occurred when retargeting it from `main` to `v1.5`?

suresh-summation:
@pdet  Done

suresh-summation:
@pdet Would really help to get this merged if it looks good. Thanks

pdet:
Hi @suresh-summation, we are in a merge-freeze process, so I'm only merging critical bug fixes to 1.5 for the next couple of days. After that, this will be good to go!

suresh-summation:
Thanks @pdet

Mytherin:
Thanks!

PR #802, opened 2026-02-24
Duckdb 1.5 test fixes

rgernhardt:
Test fixes
- merge_adjacent_sorted_spatial_hilbert.test‎: This test was missed when data inlining was made the default. The test needs to disable inling, because it looks for some data files explicitly
- merge_adjacent_max_files.test‎: Replace exact file_processed counts with invariant checks. The exact parquet file size of the small parquet files the test generates is platform dependent. How many files fill up the the 1KB limit is then also platform dependent (we've seen 5 locally and 6 in CI).
- filter_stress.test‎: The test generates parquet files using the COPY command and then has assertions on the number of files produced. The number of files depend on how many threads are running the copy. If just 1 thread then it's 54 files, but with more threads it can be less files (we've seen 13 in CI). Changing the test to be more flexible

### DIFF
--- a/test/sql/sorted_table/merge_adjacent_sorted_spatial_hilbert.test
+++ b/test/sql/sorted_table/merge_adjacent_sorted_spatial_hilbert.test
@@ -12,7 +12,7 @@ test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
 
 
 statement ok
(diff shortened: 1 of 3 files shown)


test/sql/sorted_table/merge_adjacent_sorted_spatial_hilbert.test
pdet, line 9: I think our CI skips this test because it requires spatial

pdet (commented):
Thanks, LGTM!

PR #806, opened 2026-03-03
More efficient AppendFiles

thalassemia:
When `table_changes.new_data_files` is empty, it is much faster to just move the entire `files` vector. Even when it is not empty, reserving memory and using `move_iterator` is faster and results in lower peak memory usage.

This makes a difference when using `ducklake_add_data_files` to bulk import tens or hundreds of thousands of existing Parquet files.


Mytherin:
Thanks!

PR #807, opened 2026-03-03
Use Appender API for DuckDB metadata manager

thalassemia:
The metadata manager currently constructs SQL strings to insert new metadata when writing new data files. If using DuckDB as the catalog, the Appender API eliminates most of that overhead and greatly improves bulk insert performance.

This makes a very big difference when using `ducklake_add_data_files` to bulk import tens to hundreds of thousands of existing Parquet files. Even beyond the performance improvement, peak memory usage also goes down significantly using this approach (no more massive SQL strings that have to be parsed into an AST).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ testext
 test/python/__pycache__/
 .Rhistory
 scripts/data/
+duckdb_benchmark_data/
--- a/benchmark/ingest/add_files_small_files.benchmark
+++ b/benchmark/ingest/add_files_small_files.benchmark
@@ -1,0 +1,5 @@
+# name: benchmark/ingest/add_files_small_files.benchmark
+# description: Benchmark adding many small parquet files using ducklake_add_data_files
+# group: [add_files_small]
+
+template benchmark/ingest/add_files_small_files.benchmark.in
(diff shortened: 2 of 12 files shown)


guillesd:
Hi @thalassemia! Thanks for the PR. I am not sure about this though. The nice thing about using SQL in the metadata manager is that it is a common layer to most catalogs (given that Standard SQL is used). The Appender API is exclusive to DuckDB. I'll ask internally though!

Also, since you mention performance, it is usually nice to provide some benchmark in the PR to justify this.

Thanks!

thalassemia:
Thanks for the follow up! I added a benchmark that creates 10,000 small Parquet files each with 20 columns and adds them to a DuckLake table using `ducklake_add_data_files`. On `v1.5-variegata` (7ea1564), this benchmark takes ~9 seconds. On this branch (4c88562), that goes down to ~3.8 seconds. My real dataset contains hundreds of thousands of files each with ~200 columns, so the problem (and performance discrepancy) is much bigger there.

YuweiXiao:
Interesting patch! The real power may lie in inline data. Previously, inline data acquires SQL assemble and then a large `insert into values` execution. With appender API, the inline is almost a native duckdb write.

In pg_ducklake (postgres + ducklake), we are working on a similar idea that utilizes native pg utility to speed up metadata execution. But indeed, compatibility across different catalog backend (and across versions) is an issue.

Mytherin:
Thanks!

PR #808, opened 2026-03-03
Eagerly process metadata when importing Parquet files

thalassemia:
Currently, `ducklake_add_data_files` reads all Parquet metadata into a vector of memory-intensive `ParquetFileMetadata` objects before adding them to the table. This PR processes metadata eagerly to reduce peak memory consumption.

This makes a difference when using `ducklake_add_data_files` to import tens or hundreds of thousands of existing Parquet files.

I also added a test for importing overlapping globs, a previously untested feature of `ducklake_add_data_files`. It turns out there is an issue specifically on Windows systems ([failing test log](https://github.com/thalassemia/ducklake/actions/runs/22649255427/job/65645759704)) that is resolved by replacing backslashes with forward slashes (aadf6ad).


Mytherin:
Thanks!

PR #809, opened 2026-03-03
Allow implicit type casting on Parquet import

thalassemia:
Currently, type checking is very strict when using `ducklake_add_data_files`. This PR loosens that type checking to allow implicit numerical type casts that are supported by DuckDB [according to this table](https://duckdb.org/docs/stable/sql/data_types/typecasting#casting-operations-matrix).


Mytherin:
Thanks - but this is not the correct list of type-casts to use. The supported list of type promotions for DuckLake can be found [here](https://ducklake.select/docs/stable/duckdb/usage/schema_evolution#type-promotion). This is less permissive than DuckDB itself due to having to handle stats (and is also kept compatible with the supported list of type promotions in Iceberg).

thalassemia:
Thanks for the link! Good to know about Iceberg using the same promotion rules. I'll make sure to use float/double in my Parquet files where appropriate going forwards.

PR #811, opened 2026-03-03
Implicit type casting and null handling for Hive partition keys when importing Parquet files

thalassemia:
Currently, `MapHiveColumn` receives the raw output of `HivePartitioning::Parse`, which is a map of strings. These get interpreted as type `VARCHAR`, which cannot be cast to any other type. As a result, `ducklake_add_data_file` fails to load Parquet files into DuckLakes with non-`VARCHAR` partitioning keys.

This PR uses `HivePartitioning::GetValue` to gracefully cast parsed partition key strings to their target column types before invoking `MapHiveColumn`. This also improves handling of null partition keys, thanks to duckdb/duckdb#20512.

I added a test for the new casting logic. This test revealed an edge case where string columns with no min/max stats get assigned a maximum length of 0. This causes materialization compression to mangle these columns.
```
1. test/sql/add_files/add_files_hive_partition_cast.test:57
Query unexpectedly failed (test/sql/add_files/add_files_hive_partition_cast.test:57)
 (test/sql/add_files/add_files_hive_partition_cast.test:57)!
SELECT id, category, amount FROM ducklake.test_null ORDER BY id;
Actual result:
Invalid Error: Unoptimized statement differs from original result!
Original Result:
id	category	amount	
INTEGER	VARCHAR	DECIMAL(10,2)	
[ Rows: 3]
1	o	99.99
2	NULL	50.00
3	v	NULL

Unoptimized:
id	category	amount	
INTEGER	VARCHAR	DECIMAL(10,2)	
[ Rows: 3]
1	electronics	99.99
2	NULL	50.00
3	home appliances	NULL
```
I fixed this by using `StringStats::CreateUnknown` to handle this specific case, which leaves the max length unbounded instead of setting it to 0 like `StringStats::CreateEmpty`.

It's worth noting that I encountered [a weird internal error](https://github.com/thalassemia/ducklake/actions/runs/22658324324/job/65672761590) in one of the new tests using the SQLite catalog. It appears to be something in the `sqlite_scanner` extension and only shows up with `pragma enable_verification`. I turned off query verification for that one test to work around this.

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -1147,9 +1147,13 @@ void DuckLakeFileProcessor::MapPartitionColumns(ParquetFileMetadata &file) {
 			// key not found in the file path
 			continue;
 		}
-		auto hive_value = HivePartitioning::GetValue(context, partition_key_name, hive_entry->second, field_id->Type());
+		// Get the correct type for the partition key based on the transform
+		// For YEAR/MONTH/DAY/HOUR transforms, the type is BIGINT, not the source column type
+		auto partition_key_type =
+		    DuckLakePartitionUtils::GetPartitionKeyType(partition_field.transform.type, field_id->Type());
+		auto hive_value = HivePartitioning::GetValue(context, partition_key_name, hive_entry->second, partition_key_type);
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #822, opened 2026-03-10
Fix: rewrite_data_files skips large files with delete files

### LINKED ISSUE #821: ducklake_rewrite_data_files skips data files larger than target_file_size
## Problem

`ducklake_rewrite_data_files` silently skips data files that have associated delete files when the data file size exceeds `target_file_size` (default 512 MB).

## Root Cause

In `src/functions/ducklake_compaction_functions.cpp`, `DuckLakeCompactor::GenerateCompactions` (line ~225):

```cpp
if (candidate.file.data.file_size_bytes >= target_file_size) {
    // this file by itself exceeds the threshold - skip merging
    continue;
}
```

This check is applied to **all** compaction types, including `CompactionType::REWRITE_DELETES`. It was designed for `MERGE_ADJACENT_TABLES` (where merging already-large files doesn't make sense), but incorrectly also gates the rewrite-deletes path.

As a result, any data file ≥ 512 MB with associated delete files will be ignored by `ducklake_rewrite_data_files`, even with `delete_threshold => 0`. The delete files persist indefinitely and cannot be eliminated through rewrite.

## Expected Behavior

`ducklake_rewrite_data_files` should rewrite all data files that have associated delete files (subject to `delete_threshold`), regardless of the data file size.

## Suggested Fix

Guard the file size check so it only applies to `MERGE_ADJACENT_TABLES`:

```cpp
if (candidate.file.data.file_size_bytes >= target_file_size && type != CompactionType::REWRITE_DELETES) {
    continue;
}
```

KumoSiunaus:
## Summary

- `ducklake_rewrite_data_files` silently skips data files ≥ `target_file_size` (default 512 MB), even when they have associated delete files
- The file size check was intended only for `MERGE_ADJACENT_TABLES` but was incorrectly applied to `REWRITE_DELETES` as well
- One-line fix: guard the size check with `type != CompactionType::REWRITE_DELETES`

Fixes #821

## Test plan

- [ ] Existing compaction tests still pass
- [ ] Verify that `ducklake_rewrite_data_files` now rewrites data files larger than `target_file_size` when they have delete files

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -222,8 +222,9 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 	compaction_map_t<DuckLakeCompactionCandidates> candidates;
 	for (idx_t file_idx = 0; file_idx < files.size(); file_idx++) {
 		auto &candidate = files[file_idx];
-		if (candidate.file.data.file_size_bytes >= target_file_size) {
+		if (candidate.file.data.file_size_bytes >= target_file_size && type != CompactionType::REWRITE_DELETES) {
 			// this file by itself exceeds the threshold - skip merging
+			// (does not apply to REWRITE_DELETES - delete files must be rewritten regardless of data file size)
 			continue;
 		}
 		if ((!candidate.delete_files.empty() && type == CompactionType::MERGE_ADJACENT_TABLES) ||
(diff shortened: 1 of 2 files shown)


guillesd:
hi @KumoSiunaus could you please add a test?

guillesd:
@KumoSiunaus sorry I also forgot to mention: could you retarget `v1.5-variegata` branch? Otherwise this fix won't be visible till we release the next minor version of DuckDB

Mytherin:
Thanks!

KumoSiunaus:
@Mytherin May I ask when this fix will be officially available — i.e., when the DuckLake extension containing this fix can be installed via DuckDB?

guillesd:
With DuckDB 1.5.1 targeted for 2026-03-23

Issue #828, opened 2026-03-11
DuckDB UI query fails with segfault


jprafael:
### What happens?

Testing duckdb-ui together with ducklake.
The process crashes as soon as duckdb-ui issues the following query:

```sql
SELECT
    schemata.schema_name as schema_name,
    tables.table_name as table_name,
    tables.table_type as table_type,
    tables.table_comment as table_comment,
    table_metadata.estimated_size as table_estimated_size,
    columns.ordinal_position as column_position,
    columns.column_name as column_name,
    columns.data_type as column_data_type,
    columns.column_comment as column_comment
FROM system.information_schema.schemata as schemata
LEFT JOIN system.information_schema.tables as tables
ON
    tables.table_catalog = schemata.catalog_name AND
    tables.table_schema = schemata.schema_name
LEFT JOIN duckdb_tables() as table_metadata 
ON
    table_metadata.table_name = tables.table_name AND
    table_metadata.schema_name = tables.table_schema AND
    table_metadata.database_name = tables.table_catalog
LEFT JOIN system.information_schema.columns as columns
ON
    columns.table_catalog = tables.table_catalog AND
    columns.table_schema = tables.table_schema AND
    columns.table_name = tables.table_name
WHERE
    schemata.catalog_name = 'lake' AND
    schemata.schema_name NOT IN ('information_schema', 'pg_catalog')
ORDER BY schemata.schema_name, tables.table_name, columns.ordinal_position;
```

### To Reproduce

Standalone reproducible:

```python
#!/usr/bin/env python3

import os
import shutil
import duckdb


def main() -> None:
    shutil.rmtree('/tmp/test-ducklake/', ignore_errors=True)
    os.makedirs('/tmp/test-ducklake/data/')

    db = duckdb.connect()
    db.execute("CALL enable_logging(storage => 'stdout', level=> 'INFO')")
    db.execute("INSTALL ducklake;")
    db.execute("INSTALL sqlite;")
    db.execute(
        f"""
        ATTACH 'ducklake:sqlite:/tmp/test-ducklake/metadata.sqlite'
        AS lake (
            DATA_PATH '/tmp/test-ducklake/data/'
        );
        """
    )
    db.execute(
        """
        SELECT
            schemata.schema_name as schema_name,
            tables.table_name as table_name,
            tables.table_type as table_type,
            tables.table_comment as table_comment,
            table_metadata.estimated_size as table_estimated_size,
            columns.ordinal_position as column_position,
[...]

PR #829, opened 2026-03-12
All access paths to table_data_changes go through the guard lock

jprafael:
Even for read-only paths, the access to `table_data_changes` should be guarded by the `table_data_changes_lock`.
This change plus https://github.com/duckdb/duckdb/pull/21334 ensures the added unit tests pass.

The unit test covers a specific query used by [duckdb-ui](https://github.com/duckdb/duckdb-ui) allowing it to be used with ducklake


jprafael:
Fixes #828

Mytherin:
Thanks! LGTM - future work would be to move this to a wrapper class so that grabbing the lock is enforced at an API level, but for now just fixing the issue is fine. Could you just retarget to `v1.5-variegata` so this can land in the v1.5.1 release?

Mytherin:
Thanks - test failures are unrelated (caused by an upstream bug).

Mytherin:
Actually the deadlock is most likely related

jprafael:
> Actually the deadlock is most likely related

Could you point me to the test that is failing with the deadlock?

Mytherin:
See [the CI run](https://github.com/duckdb/ducklake/actions/runs/23027947125/job/66927361099?pr=829)

jprafael:
@Mytherin I'm sorry I still can't open the [MinIO CI logs](https://github.com/duckdb/ducklake/actions/runs/23027947125) in github.
I ran all the tests locally without a deadlock, so it might be a flaky test that I need to focus testing on.
Do you know the name of the test?

Mytherin:
The CI run has likely lapsed - logs are only available temporarily. You can also look at logs on your local fork's CI run.

In any case, I've picked up your changes and have done the refactor I proposed. I think this will likely fix the deadlock as the API is now much saner / better separated which should prevent accidental locking mistakes. This is ready to go now after CI passes.

PR #830, opened 2026-03-12
Keep shared state alive

jprafael:
This is an alternative solution to https://github.com/duckdb/duckdb/pull/21334.
This makes the state alive for longer than the duration of the methods, which return reference objects.


jprafael:
The shared state edits are done on top of https://github.com/duckdb/ducklake/pull/829

Mytherin:
Thanks - are you sure these changes are necessary? The added test already passes without these changes - and the various `CreateInfo` members should all be consumed in the constructor, so keeping them alive does not make that much sense to me. Can you double check if the issue persists after https://github.com/duckdb/ducklake/pull/829?

jprafael:
@Mytherin i believe your changes to #829 made this redundant, although I'm having trouble pinpointing exactly which change made it so that the references are guaranteed to be alive after Scan() finishes.
Its ok for me to close this since the unit test consistently passes now

Mytherin:
I don't think this was ever a lifetime issue, I think the only bug here was the race condition. The lifetime issue was likely an AI hallucination.

PR #835, opened 2026-03-14
Avoid new schema version if partitions don't change

jprafael:
**Context**
It is common to have init scripts that look like this:

```sql
CREATE TABLE IF NOT EXISTS schema.table (
   field_1 VARCHAR,
   ....
);

ALTER TABLE schema.table
SET PARTITIONED BY (field_1)
```

Currently, each time this executes the table creation is skipped but the `ALTER TABLE` operation still creates a new table schema version.
This is undersirable because it prevents parquet file merging since the files have different schema versions.

**Proposed change**
Make the `ALTER TABLE` a no-op if the parititioning doesn't actually change.


pdet:
Thanks for the PR! LGTM!

Issue #813, opened 2026-03-03
ducklake_delete_orphaned_files deletes entire metadata database on an empty datalake


jprafael:
### What happens?

Invoking `CALL ducklake_delete_orphaned_files('lake', cleanup_all => true);` on an empty data lake also deletes all tables from the metadata database.

Following operations fail because of missing metadata tables, e.g.:

```
Catalog Error: Failed to get files scheduled for deletion from DuckLake: Table with name ducklake_files_scheduled_for_deletion does not exist!
```

### To Reproduce

```python
import os
import shutil
import duckdb

shutil.rmtree('//home/ubuntu/testlake-2')
os.makedirs('/home/ubuntu/testlake-2')

db = duckdb.connect()

db.execute(
    """
    FORCE INSTALL ducklake;
    FORCE INSTALL sqlite;
    """
)

db.execute(
    """
    ATTACH 'ducklake:sqlite:/home/ubuntu/testlake-2/metadata.sqlite'
    AS lake (
        DATA_PATH '/home/ubuntu/testlake-2/'
    );
    """
)

db.execute(
    """
    CALL ducklake_delete_orphaned_files('lake', cleanup_all => true);
    CALL ducklake_cleanup_old_files('lake', cleanup_all => true);
    """
)
```

Result:
```
Catalog Error: Failed to get files scheduled for deletion from DuckLake: Table with name ducklake_files_scheduled_for_deletion does not exist!
```

Expected: No error

### OS:

ubuntu 24.04 LTS, x86_64

### DuckDB Version:

v1.5.0-dev7422

### DuckLake Version:

e016fc8

### DuckDB Client:

python

### Full Name:

João Pedro Maia Rafael

### Affiliation:

Upper Delta, Unipessoal LDA

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a nightly build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

guillesd:
The issue is pretty straight forward (although dumb on DuckLake's side). When we do ducklake_delete_orphaned_files we rely on directory listings. Because your data path is also where your sqlite db file is, and this is not associated with any ducklake data file, ducklake removes it. I think we should do something smarter than this!

[...]

PR #839, opened 2026-03-16
ducklake_delete_orphaned_files now only deletes parquet files

c-herrewijn:
contains a fix for: https://github.com/duckdb/ducklake/issues/813

In case the ducklake catalog resides in the data path, it would be deleted by function `ducklake_delete_orphaned_files()`
We now prevent this by only deleting parquet files


jprafael:
@c-herrewijn does this also remove empty folders in case of partitioned tables?

c-herrewijn:
> @c-herrewijn does this also remove empty folders in case of partitioned tables?

No, the function only deletes files. That doesn't change with this PR. This PR just makes sure we only delete parquet files that are not registered and don't touch anything else.

Mytherin:
Thanks!

PR #842, opened 2026-03-16
Delete compactions might create 0 new files if the source data is fully deleted

jprafael:
In the normal path where data for a parquet file is fully deleted through a `DELETE` statement, ducklake doesn't create a delete file. `DuckLakeCompaction` is assuming that when merging a source and delete file, it must always create a new file because at least some live row must exist after compaction.

This is however not true after `ducklake_flush_inlined_data`.
This function writes the inlined data into a parquet file, and also writes inlined deletes to a delete file.
This delete file can delete all rows in the source data.

Note that an alternative approach where the source/delete parquet files are never created is not allowed, because we need to support time travel queries and `ducklake_flush_inlined_data` immediately deletes inline data.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -93,11 +93,17 @@ SourceResultType DuckLakeCompaction::GetDataInternal(ExecutionContext &context, 
 	}
 	source_state.returned_result = true;
 
+	if (!this->sink_state) {
+		throw InternalException("DuckLakeCompaction - missing sink state while producing result");
+	}
+	auto &gstate = this->sink_state->Cast<DuckLakeInsertGlobalState>();
+	auto files_created = gstate.written_files.size();
+
 	chunk.SetCardinality(1);
 	chunk.SetValue(0, 0, Value(table.schema.name));
 	chunk.SetValue(1, 0, Value(table.name));
 	chunk.SetValue(2, 0, Value::BIGINT(static_cast<int64_t>(source_files.size())));
-	chunk.SetValue(3, 0, Value::BIGINT(1)); // Each compaction creates 1 output file
+	chunk.SetValue(3, 0, Value::BIGINT(static_cast<int64_t>(files_created)));
 	return SourceResultType::FINISHED;
 }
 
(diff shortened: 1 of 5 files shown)


pdet:
Hi @jprafael, thanks for your PR.

I believe this would break time-travel, no?

If not, could you add it to the test where you go to each snapshot and verifies the data is correct?

Also, could you retarget it to v1.5?

jprafael:
@pdet You are right. The current version breaks time travel.
I think I need to use the alternative approach.
> Alternative to this implementation is that the contract is relaxed and instead ducklake_rewrite_data_files needs to be able to return "0 written files".

jprafael:
@pdet could you review the current version? all tests passing now

pdet:
I think we should have the same behavior as the `TryDropFullyDeletedFile()` optimization for the flush. Instead of writing a deletion file, we should set the `end_snapshot` of the data file.

jprafael:
@pdet here's my understanding of how things are working right now. Please let me know if these assumptions are wrong :pray:.

When data files are merged:
 - The original data files are kept in place, their metadata's `end_snapshot` is updated.
 - New files *might* be created with `start_snapshot` set and no `end_snapshot`.
 - Later,  `ducklake_cleanup_old_files` deletes the files where the `end_snapshot` has been expired.
 
 When inline data is flushed:
  - New files are always created with `start_snapshot` set and no `end_snapshot`.
  - Inline data/deletes are "immediately" deleted from the metadata database (in the same transaction as the metadata updates so that no data is lost on crash).
  - The `end_snapshot` column for inline data is actually never used.

Here's one approach to implement the optimization you mentioned:
 - Inline data is no longer deleted on flush.
 - `end_snapshot` is set for inline data after flushing.
   - Currently this is only possible on a per-row level.
   - Ideally, this should be possible for the entire entry ( `start_snapshot`/`end_snapshot` in `ducklake_inlined_data_tables`) just like it is possible for data files (`start_snapshot`/`end_snapshot` in `ducklake_data_file`).
 - Another function, like `ducklake_cleanup_old_files`, would actually delete the inline data.
 
 Is this what you mean?

pdet:
Prompted by your question I thought a bit more about it, and I think your solution is great, optimizing for the full deletion file seems like an unnecessary optimization for the deletion inlining at this point. Thanks for the great work!

Issue #841, opened 2026-03-16
404 when attempting to installing ducklake nightly extension


alastairrushworth:
### What happens?

I'm seeing a 404 error when attempting to installing ducklake nightly extension.  I'm not sure if I'm calling this incorrectly but the below doesn't work.  

```
import duckdb

con = duckdb.connect()
con.sql("FORCE INSTALL ducklake from core_nightly;")
```

results in 
```
HTTPException: HTTP Error: Failed to download extension "ducklake" at URL "http://nightly-extensions.duckdb.org/v1.5.0/windows_amd64/ducklake.duckdb_extension.gz" (HTTP 404)
Extension "ducklake" is an existing extension.
```

### To Reproduce

```
import duckdb

con = duckdb.connect()
con.sql("FORCE INSTALL ducklake from core_nightly;")
```

### OS:

MacOS & Windows

### DuckDB Version:

1.5.0

### DuckLake Version:

0.4

### DuckDB Client:

python

### Full Name:

Alastair Rushworth

### Affiliation:

Create Data Tech

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

pdet:
Fixed by #857 for 1.5.1

PR #844, opened 2026-03-17
Drop deletion inline table at flush, set nightly and build to v1.5.0

pdet:
Fix: https://github.com/duckdb/ducklake/issues/841

### DIFF
--- a/.github/duckdb-version
+++ b/.github/duckdb-version
@@ -1,1 +1,1 @@
-v1.5-variegata
+v1.5.0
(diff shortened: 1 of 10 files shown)


jprafael:
@pdet shouldn't  `DuckLakeMetadataManager::DeleteInlinedData` also call `DROP TABLE`?

pdet:
> @pdet shouldn't `DuckLakeMetadataManager::DeleteInlinedData` also call `DROP TABLE`?

In the end, I'm not going to follow through this PR, deleting the inlined table is a bad idea, as it will cause more conflicts in concurrent scenarios.

PR #846, opened 2026-03-18
Fix inlined file deletes lost on concurrent commit retry

qsliu2017:
## Summary
- **Bug 1**: `GetNewInlinedFileDeletes()` used `std::move` on the source `file_deletes` map, emptying it on the first commit attempt. When `FlushChanges` retried, the inlined file deletion data was gone — the DELETE appeared to succeed but rows were not removed.
- **Bug 2**: `delete_inlined_table_cache` (and `insert_inlined_table_name_cache`) were not cleared after rollback. On retry, the stale cache caused `GetInlinedDeletionTableName` to skip `CREATE TABLE IF NOT EXISTS` for the inlined deletion table, so the subsequent INSERT silently failed.
- Also removes the unused `inlined_table_name_cache` field (dead code).

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -311,9 +311,10 @@ private:
 public:
 	//! Read inlined file deletions for regular table scans (no snapshot info per row)
 	map<idx_t, set<idx_t>> ReadInlinedFileDeletions(TableIndex table_id, DuckLakeSnapshot snapshot);
+	//! Clear inlined table caches (needed after rollback so retry re-creates the tables)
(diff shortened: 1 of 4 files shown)


pdet:
Many thanks for the PR, this looks good to me, can you just retarget it to v1.5?

qsliu2017:
> Many thanks for the PR, this looks good to me, can you just retarget it to v1.5?

Done.

May I ask what's the branch policy for ducklake? I'm a little bit confused.

pdet:
Currently, bug fixes go to the `1.5` branch, and new features go to main. If it's a murky case (e.g., a performance improvement), I'll decide based on risk.

PR #861, opened 2026-03-19
fix(postgres): use postgres_query() for file column stats filter pushdown

### LINKED ISSUE #859: PostgreSQL metadata backend: full table scan of ducklake_file_column_stats on every query causes multi-second overhead
## Summary

When using a PostgreSQL metadata backend, DuckLake queries experience a fixed ~1-2s overhead regardless of query complexity. The root cause is that DuckDB's postgres scanner reads `ducklake_file_column_stats` using `COPY` with sequential `ctid` range batches across the entire table — even though `table_id` and `column_id` filters are correctly pushed down to PostgreSQL. With no index, this means a full sequential scan. With an index, each ctid batch can filter more efficiently but still touches every page range in the table.

## Environment

- DuckLake 0.3, DuckDB with postgres extension
- PostgreSQL 16 metadata backend (AWS RDS)
- ~800 tables, ~60K data files, ~6.7M rows in `ducklake_file_column_stats` (1.2 GB)
- One large table dominates: 30K files × 202 columns = 6.2M rows (93% of the stats table)

## Reproduction

1. Attach a DuckLake catalog with a PostgreSQL metadata store containing a non-trivial number of data files
2. Run any query against any table, e.g. `SELECT count(*) FROM ducklake.schema.some_small_table WHERE col > 'value'`
3. Observe ~1-2s fixed overhead regardless of query complexity or target table size

### ctid range batching defeats efficient index usage

In `ducklake_metadata_manager.cpp` (line ~1057), the filter pushdown code generates CTEs like:

```sql
col_0_stats AS NOT MATERIALIZED (
  SELECT data_file_id, max_value
  FROM {METADATA_CATALOG}.ducklake_file_column_stats
  WHERE column_id = 7 AND table_id = 19262
)
```

This references the table through DuckDB's attached postgres catalog. DuckDB's postgres scanner translates this into `COPY` queries with `ctid` range batches (1000 pages each):

```sql
-- Observed via pg_stat_activity during a DuckLake query:
COPY (SELECT "table_id", "column_id", "max_value", "data_file_id"
      FROM "public"."ducklake_file_column_stats"
      WHERE ctid BETWEEN '(48000,0)'::tid AND '(49000,0)'::tid
        AND "table_id" = '19262' AND "column_id" = '7'
) TO STDOUT (FORMAT "binary");
```

The `table_id` and `column_id` filters **are** pushed down, but the `ctid BETWEEN` clause forces PostgreSQL to scan every page range in the table sequentially. With a 134K-page table, this means issuing ~134 batch queries that each scan 1000 pages — effectively reading the entire table even though only a tiny fraction of rows match.

### Benchmarks

| Scenario | Time |
|----------|------|
| No index on `ducklake_file_column_stats` | ~2.0s |
| With `(table_id, column_id)` index | ~1.25s |
| Direct PostgreSQL query with same filters (no ctid batching) | <5ms |
| DuckLake `SELECT 1 FROM table LIMIT 1` (EXPLAIN ANALYZE execution only) | 29ms |

### Schema/stats caching works, but file_column_stats is loaded fresh every query

- Schema catalog: cached by `schema_version` ✓
- Global table stats: cached by `next_file_id` ✓
- File column stats (filter pushdown): **not cached, loaded every query** ✗

### Contrast with GetLatestSnapshotQuery

`postgres_metadata_manager.cpp` already uses `postgres_query()` for the snapshot query, which executes SQL directly on PostgreSQL without ctid batching. The metadata queries in the base class go through the DuckDB catalog scanner instead.

## Suggested Fix

**Option A: Use `postgres_query()` for filter pushdown CTEs.** Similar to how `GetLatestSnapshotQuery()` already works in `postgres_metadata_manager.cpp`, execute the `ducklake_file_column_stats` query directly on PostgreSQL rather than through the DuckDB catalog scanner. This would allow PostgreSQL to use an index scan (returning ~200 rows in <5ms) instead of scanning all 134K pages.

**Option B: Cache file column stats per `(table_id, snapshot)`.** The schema and global stats are already cached by `schema_version`/`next_file_id`. Extending this caching to per-file column stats would avoid repeated full scans within the same session.

For a catalog with 800 tables where one table has 30K files × 200 columns, Option A would reduce query overhead from ~1-2s to <50ms.

fuziontech:
## Summary

When DuckLake uses a PostgreSQL metadata backend, filter pushdown CTEs for `ducklake_file_column_stats` currently query via `{METADATA_CATALOG}`, which routes through DuckDB's postgres scanner. The scanner bulk-copies the entire table using `COPY` with ctid range batches, defeating any PostgreSQL indexes.

For large catalogs (e.g. 6.7M rows in `ducklake_file_column_stats`), this causes multi-second overhead on every query even when only a few hundred rows match the specific `table_id`/`column_id` filter.

This PR overrides `GenerateCTESectionFromRequirements` in `PostgresMetadataManager` to use `postgres_query()` — the same pattern already used by `GetLatestSnapshotQuery()`. This executes the filter directly on PostgreSQL where indexes can be leveraged, returning only the matching rows to DuckDB.

## Changes

- **CTE override**: Added `GenerateCTESectionFromRequirements` override to `PostgresMetadataManager` that generates CTEs using `postgres_query()` with `{METADATA_CATALOG_NAME_LITERAL}` and `{METADATA_SCHEMA_ESCAPED}` placeholders (matching the existing `GetLatestSnapshotQuery` pattern). This covers both `GetFilesForTable` and `GetExtendedFilesForTable` code paths since they both use `GenerateFilterPushdownComponents`.

- **Index creation**: Added a `(table_id, column_id)` index on `ducklake_file_column_stats` — both in `InitializeDuckLake` (new catalogs) and as a `MigrateV04` step (existing v0.4 catalogs). Without this index, PostgreSQL would still do a sequential scan even with `postgres_query()`. Bumps catalog version from 0.4 to 0.5.

## Scope / Known Limitations

- This fix targets the **CTE-based filter pushdown** path (static column filters). The new **dynamic filter LEFT JOINs** to `ducklake_file_column_stats` (added for Top-N optimization) still go through `{METADATA_CATALOG}` and would benefit from a similar treatment in a follow-up.

## Testing

- The existing `filter_pushdown.test` exercises the full filter pushdown path including CTE generation. When run with the postgres test config (`test/configs/postgres.json`), it exercises the `PostgresMetadataManager` code path including the new `GenerateCTESectionFromRequirements` override.
- The existing `migration.test` validates chain migration from all prior versions (v01 through v04-dev1), which now includes the v0.4 → v0.5 step. A `v04.db.gz` test fixture should be generated for explicit v0.4 → v0.5 migration testing.

## Context

We observed ~2s metadata resolution overhead per query on a production DuckLake catalog with ~6.7M rows in `ducklake_file_column_stats` (one table with 30K files × 202 columns). The postgres scanner was issuing ~134 `COPY` queries with ctid range batches to read the entire table, ignoring the `WHERE column_id = X AND table_id = Y` filter that would have returned only a few hundred rows via an index scan.

With this fix and the included index, PostgreSQL can serve each CTE lookup as a fast index scan instead of a full table copy.

Fixes #859


kelf:
any progress on this one?

pdet:
Hi @fuziontech, thanks for the PR. I think in its current state it is a bit problematic, as it introduces a spec change that affects all catalogs (i.e., index creation).

Even automatically adding the index just for Postgres requires careful consideration, as indexes also have drawbacks.

Maybe the correct way to support this would be to have `GetLatestSnapshotQuery()` change if the index exists. Creating the index would then be a deliberate user decision.

rschmukler:
Happy to see this issue and PR open, my team and I hit it as well. Is there any update on the PR or preferred approach?

fuziontech:
Didn't see the feedback until now so I'll start working on that to get this in fast 🏃

fuziontech:
Hey @pdet - rebased on current `main` and reworked the PR per your feedback:

- **No more automatic index creation, no catalog spec change, no version bump.** Index creation is left as a deliberate user decision.
- `PostgresMetadataManager::GenerateCTESectionFromRequirements` now lazily checks whether `ducklake_file_column_stats` has any index whose key columns include both `table_id` and `column_id`, and only switches to the `postgres_query()` CTE form when one exists. Without an index it falls through to the existing `DuckLakeMetadataManager` path. The check result is cached for the lifetime of the metadata manager (per-transaction), so a user who creates the index sees the fast path on their next transaction without needing to re-attach.
- Added `test/sql/stats/filter_pushdown_postgres_index.test` exercising both code paths (no index → base CTE, then `CREATE INDEX` → `postgres_query()` CTE) under the postgres test config. Gated behind `require postgres_scanner` and `require-env DUCKLAKE_CI`, matching `ducklake_settings_postgres.test`.

The branch was significantly behind `main` (catalog version refactor, V1.x migrations, etc.) so I reset onto `origin/main` and re-applied just the focused fix rather than fighting through the rebase. The unrelated schema-id remapping commits that were tangled into the original branch are dropped from this PR.

Diff is now `+155 / -2` across:
- `src/include/storage/ducklake_metadata_manager.hpp`
- `src/include/metadata_manager/postgres_metadata_manager.hpp`
- `src/metadata_manager/postgres_metadata_manager.cpp`
- `test/sql/stats/filter_pushdown_postgres_index.test`

Local verification:
- `test/sql/stats/filter_pushdown.test` passes on both default config and postgres config (the no-index fallback exercises the original code path).
- New test skips locally (postgres_scanner not statically linked into local unittest) and should run in CI.

Let me know if you'd prefer a different shape for the runtime detection (e.g. config flag, or detect once per catalog rather than per transaction).

pdet:
Hi @fuziontech thanks for the work here, I've changed my mind and decided this is a safe addition to a pre 1.1 release as the index can be manually built.

To make it to the next extension release, I took the liberty of picking up your work and fixing some details, as it would need to be merged today.

Closing this PR as is superseeded by: https://github.com/duckdb/ducklake/pull/1147

PR #866, opened 2026-03-20
fix: compaction should not inherit per_thread_output from lake config

### LINKED ISSUE #864: ducklake_rewrite_data_files and ducklake_merge_adjacent_files crash with per_thread_output enabled
### What happens?

Both `ducklake_rewrite_data_files()` and `ducklake_merge_adjacent_files()` throw `InvalidInputException: PRESERVE_ORDER is not supported with these parameters` when the DuckLake database has `per_thread_output` option set to `true`.

### To Reproduce

```sql
ATTACH 'ducklake:test.ducklake' AS db (DATA_PATH 'test_data/', DATA_INLINING_ROW_LIMIT 0);
CALL ducklake_set_option('db', 'per_thread_output', true);

CREATE TABLE db.t1 (id INT, value VARCHAR);
INSERT INTO db.t1 VALUES (1, 'a'), (2, 'b');
INSERT INTO db.t1 VALUES (3, 'c'), (4, 'd');
INSERT INTO db.t1 VALUES (5, 'e'), (6, 'f');

-- Crashes:
CALL ducklake_merge_adjacent_files('db');
-- Invalid Input Error: PRESERVE_ORDER is not supported with these parameters

-- Also crashes (after DELETE creates delete files):
DELETE FROM db.t1 WHERE id = 1;
CALL ducklake_rewrite_data_files('db', delete_threshold => 0);
-- Invalid Input Error: PRESERVE_ORDER is not supported with these parameters
```

### Root Cause

Both compaction functions share a single code path through `GenerateCompactionPlan` in `src/functions/ducklake_compaction_functions.cpp`. This function copies `per_thread_output` from the lake's copy options (line 563) and then unconditionally sets `PRESERVE_ORDER` (line 574):

```cpp
copy->per_thread_output = copy_options.per_thread_output;  // line 563
// ...
copy->preserve_order = PreserveOrderType::PRESERVE_ORDER;  // line 574
```

DuckDB rejects the combination of `per_thread_output=true` + `PRESERVE_ORDER` in `plan_copy_to_file.cpp`.

The `per_thread_output` option was introduced for INSERT in commit 91ab44b6. That commit changed `GetCopyOptions` from hardcoded `per_thread_output = false` to reading from the lake config, but did not update the compaction code path which shares the same `GetCopyOptions`.

Lines 575-576 already override `file_size_bytes` and `rotate` for compaction — `per_thread_output` was missed.

There are three call sites for `GetCopyOptions` in the codebase: INSERT (`ducklake_insert.cpp`), compaction (`ducklake_compaction_functions.cpp`), and flush inlined data (`ducklake_flush_inlined_data.cpp`). Only the compaction path sets `PRESERVE_ORDER`, so only it is affected.

### Suggested Fix

Override `per_thread_output` to `false` for compaction, same pattern as the existing overrides:

```cpp
copy->per_thread_output = false;
```

### OS:

macOS 15.4, arm64

### DuckDB Version:

1.4.4

### DuckLake Version:

current main (issue introduced in commit 91ab44b6)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Ilia Ablamonov

### Affiliation:

Rick.ai

- [x] Yes, I have

- [x] Yes, I have

Flamefork:
## Summary

- Compaction (`ducklake_merge_adjacent_files` / `ducklake_rewrite_data_files`) crashes with `PRESERVE_ORDER is not supported with these parameters` when `per_thread_output` is enabled on the lake
- Root cause: `GenerateCompactionPlan` copies `per_thread_output` from lake config but unconditionally sets `PRESERVE_ORDER` — DuckDB rejects this combination
- Fix: override `per_thread_output` to `false` for compaction, same pattern as the existing `file_size_bytes` and `rotate` overrides

Fixes #864

## Test plan

- [x] New test `test/sql/compaction/compaction_per_thread_output.test` — verifies both `merge_adjacent_files` and `rewrite_data_files` succeed with `per_thread_output` enabled and data integrity is preserved
- [x] All existing compaction tests pass

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -574,7 +574,7 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
 	copy->filename_pattern = std::move(copy_options.filename_pattern);
 	copy->file_extension = std::move(copy_options.file_extension);
 	copy->overwrite_mode = copy_options.overwrite_mode;
-	copy->per_thread_output = copy_options.per_thread_output;
+	copy->per_thread_output = false;
 	copy->file_size_bytes = copy_options.file_size_bytes;
 	copy->rotate = copy_options.rotate;
 	copy->return_type = copy_options.return_type;
--- a/test/sql/compaction/compaction_per_thread_output.test
+++ b/test/sql/compaction/compaction_per_thread_output.test
@@ -1,0 +1,65 @@
+# name: test/sql/compaction/compaction_per_thread_output.test
+# description: test compaction with per_thread_output enabled
+# group: [compaction]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_compaction_per_thread_output')
+
+statement ok
+CALL ducklake.set_option('per_thread_output', 'true');
+
+statement ok
+CREATE TABLE ducklake.test(i INTEGER);
+
+statement ok
+INSERT INTO ducklake.test VALUES (1);
+
+statement ok
+CALL ducklake_flush_inlined_data('ducklake')
+
+statement ok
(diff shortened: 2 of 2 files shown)


pdet:
Hi @Flamefork LGTM, can you retarget it to v1.5?

Flamefork:
Hi @pdet done (if I understood you correctly)
Thanks!

pdet:
Yes, that was it! Thanks!

PR #868, opened 2026-03-20
Cache existence of GetInlinedDeletionTableName

rgernhardt:
Cache the existence of GetInlinedDeletionTables.

# Caching

Before this change, each query would check for the existence of the deletion inlining table using 
`SELECT NULL FROM {METADATA_CATALOG}.%s LIMIT 1`.
For catalogs like remote postgres, this is an additional roundtrip to the catalog.
This changes introduces a cache on the catalog to remember the existence or non-existence. 

The cache distinguishes "known to exist" (permanent — these tables are never dropped) from "known to not exist" (valid until the snapshot advances, since a new delete could create the table).
 
# Existence check and error handling

Separately, fix three callers ReadInlinedFileDeletions, GetFileIdsWithInlinedDeletions, ReadInlinedFileDeletionsForRange. These callers would swallow errors in their queries. These callers already know the inlining table exists when they issue their queries, so errors should always be unexpected at that point.

One issue remains with the approach of using an error check for the existence check in GetInlinedDeletionTableName: A transient error on the query could lead us to assume that the deletion inlining table does not exist, even if it does. This could lead to wrong results. 
This is a preexisting issue, but I think this needs to be addressed

# Benchmark

  ClickBench (43 queries, 100M rows) against a Postgres-backed DuckLake catalog with no inlined deletions:
```
  ┌─────────────┬───────────┐
  │             │ Total (s) │
  ├─────────────┼───────────┤
  │ With fix    │ ~20.2     │
  ├─────────────┼───────────┤
  │ Without fix │ ~21.4     │
  ├─────────────┼───────────┤
  │ Speedup     │ ~6%       │
  └─────────────┴───────────┘
```

The improvement is consistent across runs. Fast queries benefit most in relative terms since the fixed-cost metadata round-trip is a larger fraction of their total runtime.

Q7 Speedup 1.18x
Q17 Speedup 1.29x
Q26 Speedup 1.27x

All of these queries have a small compute time so the saved round-trip time is a bigger fraction of total runtime.

**Note:** I did not measure significant benefit when running against a local duckdb catalog, where the there is no network round-trip for querying the metadata

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -171,6 +171,12 @@ public:
 	//! Return the schema for the given snapshot - loading it if it is not yet loaded
 	DuckLakeCatalogSet &GetSchemaForSnapshot(DuckLakeTransaction &transaction, DuckLakeSnapshot snapshot);
 
+	//! Check if an inlined deletion table is known to exist or not exist for the given table and snapshot
+	//! Returns: 1 = known to exist, -1 = known to not exist, 0 = unknown (need to query)
+	int CheckInlinedDeletionTableCache(TableIndex table_id, DuckLakeSnapshot snapshot);
+	//! Cache the result of an inlined deletion table existence check
+	void CacheInlinedDeletionTableResult(TableIndex table_id, DuckLakeSnapshot snapshot, bool exists);
+
 private:
 	void DropSchema(ClientContext &context, DropInfo &info) override;
 	unique_ptr<DuckLakeCatalogSet> LoadSchemaForSnapshot(DuckLakeTransaction &transaction, DuckLakeSnapshot snapshot);
@@ -201,6 +207,13 @@ private:
 	string metadata_type;
 	//! Whether or not the catalog is initialized
 	bool initialized = false;
+	//! Cache for inlined deletion table existence checks
+	mutex inlined_deletion_cache_lock;
+	//! Table IDs where the inlined deletion table is known to exist (permanent - never invalidated)
+	unordered_set<idx_t> inlined_deletion_exists;
+	//! Table IDs where the inlined deletion table is known to NOT exist, with the snapshot_id at which we checked
+	//! Valid as long as current snapshot.snapshot_id <= cached snapshot_id
+	unordered_map<idx_t, idx_t> inlined_deletion_not_exists;
 	//! The id of the last committed snapshot, set at FlushChanges on a successful commit
 	mutable mutex commit_lock;
 	optional_idx last_committed_snapshot;
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -25,6 +25,8 @@ struct DuckLakeConfigOption;
 struct DeleteFileMap;
 class LogicalGet;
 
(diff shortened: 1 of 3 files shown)


src/include/storage/ducklake_catalog.hpp
```diff
@@ -171,6 +171,12 @@ class DuckLakeCatalog : public Catalog {
 	//! Return the schema for the given snapshot - loading it if it is not yet loaded
 	DuckLakeCatalogSet &GetSchemaForSnapshot(DuckLakeTransaction &transaction, DuckLakeSnapshot snapshot);
 
+	//! Check if an inlined deletion table is known to exist or not exist for the given table and snapshot
+	//! Returns: 1 = known to exist, -1 = known to not exist, 0 = unknown (need to query)
+	int CheckInlinedDeletionTableCache(TableIndex table_id, DuckLakeSnapshot snapshot);
```
pdet: Should we have an enum here?

src/include/storage/ducklake_catalog.hpp
```diff
@@ -201,6 +207,13 @@ class DuckLakeCatalog : public Catalog {
 	string metadata_type;
 	//! Whether or not the catalog is initialized
 	bool initialized = false;
+	//! Cache for inlined deletion table existence checks
+	mutex inlined_deletion_cache_lock;
+	//! Table IDs where the inlined deletion table is known to exist (permanent - never invalidated)
+	unordered_set<idx_t> inlined_deletion_exists;
+	//! Table IDs where the inlined deletion table is known to NOT exist, with the snapshot_id at which we checked
+	//! Valid as long as current snapshot.snapshot_id <= cached snapshot_id
+	unordered_map<idx_t, idx_t> inlined_deletion_not_exists;
```
pdet, line 217: I understand why we would want an `inlined_deletion_exists` at the catalog level, but why do we need the negation, does it provide a caching benefit that we don't achieve with the inlined_deletion_exists already?

pdet (commented):
Hi @rgernhardt thanks for the PR! I just had a question about the code and a straight-forward comment.

src/include/storage/ducklake_catalog.hpp
rgernhardt: I think it does. 
For example I had a table that I created from ingesting some parquets and then started querying it. The table didn't have any inlined delete. 
Every query did the `SELECT NULL FROM {METADATA_CATALOG}.%s LIMIT 1 ` check (sometimes multiple times). 
With the negative cache, only the first query does the check and then all other queries used the cached information

pdet: Ah, you can't use `inlined_deletion_exists` because that's not reliably filled with data from the get-go, is that it?

rgernhardt: That's right. The absence from inlined_deletion_exists could mean two things: Either the table does not exists OR we haven't checked it yet. That's why we need the negative version as well.

src/include/storage/ducklake_catalog.hpp
rgernhardt: Agree, will fix

pdet:
Thanks!

PR #870, opened 2026-03-21
refactor: replace IN subqueries with LEFT JOIN in table_changes macro

### LINKED ISSUE #869: table_changes macro uses O(n²) IN subquery, hangs on large UPDATE operations
## Problem

The `table_changes` macro in `ducklake_table_changes.cpp` uses `IN` subqueries to classify rows as insert/update/delete:

```sql
SELECT ... CASE WHEN (snapshot_id, rowid) IN (SELECT snapshot_id, rowid FROM ducklake_table_deletions(...))
       THEN 'update_postimage' ELSE 'insert' END AS change_type, *
FROM ducklake_table_insertions(...)
UNION ALL
SELECT ... CASE WHEN (snapshot_id, rowid) IN (SELECT snapshot_id, rowid FROM ducklake_table_insertions(...))
       THEN 'update_preimage' ELSE 'delete' END AS change_type, *
FROM ducklake_table_deletions(...)
```

The `IN` subquery results in a nested loop scan, making the complexity O(n²) where n is the number of changed rows. For a table with ~176k updated rows, `table_changes` hangs indefinitely.

## Reproduction

1. Create a DuckLake table with ~176k rows
2. Run an `UPDATE` that touches all rows (e.g. setting a column value)
3. Call `table_changes` for that snapshot — it hangs

## Proposed fix

Replace the `IN` subqueries with `LEFT JOIN`, which uses hash joins for O(n) complexity:

```sql
SELECT i.snapshot_id, i.rowid,
       CASE WHEN d.rowid IS NOT NULL THEN 'update_postimage' ELSE 'insert' END AS change_type, i.*
FROM ducklake_table_insertions(...) i
LEFT JOIN ducklake_table_deletions(...) d ON i.snapshot_id = d.snapshot_id AND i.rowid = d.rowid
UNION ALL
SELECT d.snapshot_id, d.rowid,
       CASE WHEN i.rowid IS NOT NULL THEN 'update_preimage' ELSE 'delete' END AS change_type, d.*
FROM ducklake_table_deletions(...) d
LEFT JOIN ducklake_table_insertions(...) i ON d.snapshot_id = i.snapshot_id AND d.rowid = i.rowid
```

PR with fix incoming.

jcolot:
## Summary

- Replace `IN` subqueries with `LEFT JOIN` in the `ducklake_table_changes` macro
- Makes the join strategy explicit rather than relying on optimizer decorrelation
- More readable and easier to reason about

## Test plan

- [x] Existing `test/sql/table_changes/ducklake_table_changes.test` passes unchanged (same output, different execution plan)

### DIFF
--- a/src/functions/ducklake_table_changes.cpp
+++ b/src/functions/ducklake_table_changes.cpp
@@ -6,15 +6,19 @@ namespace duckdb {
 // clang-format off
 static const DefaultTableMacro ducklake_table_macros[] = {
 	{DEFAULT_SCHEMA, "ducklake_table_changes", {"catalog", "schema_name", "table_name", "start_snapshot", "end_snapshot", nullptr}, {{nullptr, nullptr}},  R"(
-SELECT snapshot_id, rowid, CASE WHEN (snapshot_id, rowid) in (SELECT snapshot_id, rowid FROM ducklake_table_deletions(catalog, schema_name, table_name, start_snapshot, end_snapshot))
-       THEN 'update_postimage'
-       ELSE 'insert'
-       END AS change_type, * FROM ducklake_table_insertions(catalog, schema_name, table_name, start_snapshot, end_snapshot)
+SELECT i.snapshot_id, i.rowid,
+       CASE WHEN d.rowid IS NOT NULL THEN 'update_postimage' ELSE 'insert' END AS change_type,
+       i.*
+FROM ducklake_table_insertions(catalog, schema_name, table_name, start_snapshot, end_snapshot) i
+LEFT JOIN ducklake_table_deletions(catalog, schema_name, table_name, start_snapshot, end_snapshot) d
+ON i.snapshot_id = d.snapshot_id AND i.rowid = d.rowid
 UNION ALL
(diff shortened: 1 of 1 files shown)


pdet:
Hi @jcolot this LGTM, can you retarget it v1.5?

jcolot:
Retargeted to v1.5-variegata and rebased. Thanks!

pdet:
Hi @jcolot, can you have a look in the failing CI?

jcolot:
The CI failure was due to the branch being behind v1.5-variegata. Rebased onto the latest should be resolved now.

jcolot:
Rebased onto the latest v1.5-variegata — the CI failure should be resolved now.

jcolot:
**Update after investigation:** I did further testing to try to reproduce the O(n²) behavior described in the original PR, but was unable to confirm it. The DuckDB optimizer consistently converts the `IN` subquery into a `HASH_JOIN (MARK)` — which is O(n) — across all tested scenarios (DuckDB 1.4.2 and 1.5-dev, in-memory, Postgres metadata, and S3 data storage).

The original hang I experienced was caused by a different factor in my code, downstream of the `table_changes` call, not by the `IN` subquery itself.

That said, the `LEFT JOIN` rewrite is still a worthwhile change:
- It makes the join strategy explicit rather than relying on optimizer decorrelation
- It's more readable and easier to reason about
- It's a defensive safeguard against any edge case where the optimizer might not decorrelate the `IN` subquery

pdet:
Thanks!

PR #878, opened 2026-03-24
Log type for metadata queries

rgernhardt:
Add a new log type for ducklake metadata queries. 

Metadata queries can be an important factor in performance, especially for remote catalogs. This changes adds a log type with fields catalog, query and elapsed_ms to debug metadata queries.

The log can be enabled like this
```
CALL enable_logging('DuckLakeMetadata', storage='shell_log_storage');
```

Note that on the INFO level, there is already a QueryLog, which logs every query on every connection including metadata queries. I think it's still worthwhile to have a separate log type for metadata queries, so you can filter on those alone (and have timing information in the elapsed_ms field).

This change also adds a small hook to DuckLakeCatalog that applications that embed ducklake can use to receive the metadata queries with their time, for example in order to add them to their observability pipelines.

### DIFF
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -5,14 +5,19 @@
 #include "storage/ducklake_storage.hpp"
 #include "functions/ducklake_table_functions.hpp"
 #include "storage/ducklake_secret.hpp"
+#include "duckdb/logging/log_manager.hpp"
 #include "duckdb/storage/storage_extension.hpp"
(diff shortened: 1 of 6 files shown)


pdet:
Hi @rgernhardt thanks for the PR! Can you also add tests?

rgernhardt:
> Hi @rgernhardt thanks for the PR! Can you also add tests?

Added!

pdet:
Thanks!

PR #903, opened 2026-03-25
Fix single quote escaping in variant stats TrySerialize

### LINKED ISSUE #902: VARIANT column with PostgreSQL catalog fails when shredded string stats contain single quotes
### What happens?

VARIANT columns with a Postgres metadata catalog blow up on commit if a shredded string field's min/max contains a single quote. The `extra_stats` JSON in `ducklake_table_column_stats` gets wrapped in single quotes without escaping, so Postgres chokes on a syntax error.

Interestingly the per-file stats (`ducklake_file_variant_stats`) in the same transaction *do* escape correctly (`'it''s broken'`), so it's just this one spot that was missed.

Same class of bug as #790 (fixed for data inlining).

### To Reproduce

Full repro with testcontainers: https://github.com/wideltann/ducklake-variant-bug

```sql
ATTACH 'ducklake:postgres:dbname=... host=... user=... password=...' AS dl (DATA_PATH '/tmp/files/');
USE dl;
CREATE TABLE t (v VARIANT);
INSERT INTO t VALUES ('[{"text": "hello"}]'::JSON::VARIANT);
-- works

INSERT INTO t VALUES ('[{"text": "it''s broken"}]'::JSON::VARIANT);
-- fails: syntax error at or near "s"
-- ...,"min":"hello","max":"it's broken",...
--                          ^ unescaped quote
```

Error from Postgres:

```
Failed to flush changes into DuckLake: Failed to execute query "...
VALUES (1, 1, false, NULL, NULL, NULL, '[{"field_name":"element.\"text\"","shredded_type":"varchar","null_count":0,"min":"hello","max":"it's broken","num_values":2,...}]')
...": ERROR:  syntax error at or near "s"
```

### Where the bug is

`TrySerialize()` in `src/storage/statistics/ducklake_variant_stats.cpp`:

```cpp
result = "'" + out + "'";
```

Should be using `DuckLakeUtil::SQLLiteralToString(out)` which does the `Replace("'", "''")`. Patch here: https://github.com/wideltann/ducklake-variant-bug/blob/main/fix.patch

### OS:

macOS 15, arm64

### DuckDB Version:

v1.5.1

### DuckLake Version:

latest (via `INSTALL ducklake`)

### DuckDB Client:

Python

### Full Name:

Tanner Dellett-Wion

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

wideltann:
`TrySerialize()` in `ducklake_variant_stats.cpp` wraps the serialized JSON in plain single quotes:

```cpp
result = "'" + out + "'";
```

When a shredded string field's min/max contains a single quote (e.g. `"max":"it's broken"`), this produces broken SQL for PostgreSQL catalogs. The per-file stats in `ducklake_file_variant_stats` already escape correctly via `StatsToString()`, so this was just missed in the table-level stats path.

Fix: use `DuckLakeUtil::SQLLiteralToString(out)` which does `Replace("'", "''")`.

Also adds a regression test with a single quote in variant string data.

Fixes #902

### DIFF
--- a/src/storage/statistics/ducklake_variant_stats.cpp
+++ b/src/storage/statistics/ducklake_variant_stats.cpp
@@ -10,6 +10,7 @@
 #include "duckdb/parser/keyword_helper.hpp"
 #include "storage/ducklake_metadata_info.hpp"
 #include "yyjson.hpp"
+#include "common/ducklake_util.hpp"
 
 namespace duckdb {
 
@@ -208,7 +209,7 @@ bool DuckLakeColumnVariantStats::TrySerialize(string &result) const {
 	string out(json, len);
 	free(json);
 	yyjson_mut_doc_free(doc);
-	result = "'" + out + "'";
+	result = DuckLakeUtil::SQLLiteralToString(out);
 	return true;
 }
 
--- a/test/sql/stats/variant_stats_special_chars.test
+++ b/test/sql/stats/variant_stats_special_chars.test
@@ -1,0 +1,50 @@
+# name: test/sql/stats/variant_stats_special_chars.test
+# description: test that single quotes in variant string stats don't break serialization
+# group: [stats]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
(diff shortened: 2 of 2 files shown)


pdet:
Hi @wideltann thanks for the PR, can you retarget it to the v1.5 branch?

pdet:
Thanks! LGTM!

PR #904, opened 2026-03-26
Don't scan duckdb_tables() during Initialize

rgernhardt:
Problem: We use a duckdb_tables() scan during ducklake initialize to verify whether the ducklake is initialozed. This loads all schemas of all attached ducklakes. This will prevent a ducklake attach if another unrelated ducklake database is corrupted. This also has negative performance impact, we might load more schemas than needed. 

Solution: Directly check for the existence of a ducklake metadata table we always expect to be there to check if the ducklake is initialized.

### DIFF
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -76,6 +76,7 @@ void DuckLakeInitializer::Initialize() {
 	// check if we are loading an existing DuckLake or creating a new one
 	// directly query a known ducklake metadata table to avoid scanning all attached catalogs via duckdb_tables()
 	// this prevents a corrupted ducklake catalog from blocking initialization of unrelated ducklake databases
+	// FIXME: verify that all ducklake tables are in the correct format
 	result = transaction.Query("SELECT NULL FROM {METADATA_CATALOG}.ducklake_metadata LIMIT 1");
 	if (result->HasError()) {
 		auto &error_obj = result->GetErrorObject();
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -74,22 +74,23 @@ void DuckLakeInitializer::Initialize() {
 	}
 	// after the metadata database is attached initialize the ducklake
 	// check if we are loading an existing DuckLake or creating a new one
-	// FIXME: verify that all tables are in the correct format instead
-	result = transaction.Query(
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_initializer.cpp
```diff
@@ -74,22 +74,23 @@ void DuckLakeInitializer::Initialize() {
 	}
 	// after the metadata database is attached initialize the ducklake
 	// check if we are loading an existing DuckLake or creating a new one
-	// FIXME: verify that all tables are in the correct format instead
```
pdet, line 77: Was the `FIXME` removed on purpose?

test/sql/issues/corrupted_catalog_fault_isolation.test
```diff
@@ -0,0 +1,34 @@
...
+# Create a ducklake catalog with a table
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS corrupt_dl (DATA_PATH '${DATA_PATH}/corrupt_dl/')
+
+statement ok
+CREATE TABLE corrupt_dl.t1 AS SELECT 1 AS id, 'hello' AS name
+
+# Corrupt the metadata: set an invalid column type
+statement ok
+UPDATE __ducklake_metadata_corrupt_dl.ducklake_column SET column_type = 'int32varcharvarchar' WHERE column_name = 'id'
```
pdet: Can we also detach and attach the `corrupt_dl` to verify it errors gracefully?

pdet (commented):
Hi @rgernhardt thanks for the PR, is looking good!

Can you also have a look at the failing CI?

src/storage/ducklake_initializer.cpp
rgernhardt: Ah, putting it back in

test/sql/issues/corrupted_catalog_fault_isolation.test
rgernhardt: Done. It's actually not the attach that will fail because of lazy loading. But the first query after the attach. Added that to the test

rgernhardt:
> Hi @rgernhardt thanks for the PR, is looking good!
> 
> Can you also have a look at the failing CI?

Done, I excluded this test from postgres and sqlite background now, since the test requires multiple independent metadata catalogs

pdet:
Thanks!

PR #906, opened 2026-03-27
fix: detect concurrent insert and delete transactions on the same table

hello-world-bfree:
**tl;dr**: Concurrent transactions that insert into and delete from the same table now conflict, preventing silent duplicate rows in upsert workloads. Both delete-and-insert & merge into. Inserts without deletes are unaffected

### Problem

`CheckForConflicts` had no cross-checks between insert and delete operations. Two concurrent transactions could both  read the same snapshot, both decide a row doesn't exist, and both insert it. This creates duplicates. Any concurrent upsert patterns are affected.

### Fix

Addition of bi-directional `ConflictCheck`s; inserts now conflict with concurrent deletes and vice versa

This happens at the table-level similar to the `tables_deleted_inlined` check. To do something similar at the file or row level would be a much larger change (from what I can tell) but could be the way the way to go if table-levvel is thought ot be too coarse.

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1236,16 +1236,22 @@ void DuckLakeTransaction::CheckForConflicts(const TransactionChangeInformation &
 	for (auto &table_id : changes.tables_inserted_into) {
 		ConflictCheck(table_id, other_changes.dropped_tables, "insert into table", "dropped it");
 		ConflictCheck(table_id, other_changes.altered_tables, "insert into table", "altered it");
+		ConflictCheck(table_id, other_changes.tables_deleted_from, "insert into table", "deleted from it");
+		ConflictCheck(table_id, other_changes.tables_deleted_inlined, "insert into table", "deleted inlined data from it");
 	}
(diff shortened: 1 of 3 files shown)


steven-luabase:
We just ran into a similar problem, both with wrapping DELETE and INSERT into the same transaction and using MERGE. The result returned duplicates in the target table.

pdet:
Hi @hello-world-bfree, thanks for the PR, can you retarget it to v1.5?

pdet:
Hi @hello-world-bfree thanks! With retarget I meant moving your commit to v1.5-variegata, I think here is currently looking like a merge from main?

hello-world-bfree:
> Hi @hello-world-bfree thanks! With retarget I meant moving your commit to v1.5-variegata, I think here is currently looking like a merge from main?

Ope! Sorry about that @pdet. I wasn't paying attention. Should be gtg now

tintamarre:
I spent the entire day yesterday debugging this problem before finally coming across this issue. I really hope it gets fixed quickly, it’s a crucial piece for synchronizing a source with a destination on Ducklake. Thanks for the work! 👍

pdet:
This looks great, thanks for the work!

PR #907, opened 2026-03-27
fix: use typed NULLs in UNION subqueries for SQL standard compliance

qsliu2017:
Bare `NULL` columns in UNION/subquery contexts lack explicit types, which can cause issues with strict SQL engines (PostgreSQL).

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1612,7 +1612,12 @@ LEFT JOIN LATERAL (
 	ORDER BY data_file_id, begin_snapshot DESC
 ) AS previous_delete
 USING (data_file_id), (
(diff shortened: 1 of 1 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1612,16 +1612,16 @@ LEFT JOIN LATERAL (
 	ORDER BY data_file_id, begin_snapshot DESC
 ) AS previous_delete
 USING (data_file_id), (
-	SELECT NULL path, NULL path_is_relative, NULL file_size_bytes, NULL footer_size, NULL encryption_key, NULL format
+	SELECT CAST(NULL AS VARCHAR) AS path, CAST(NULL AS BOOLEAN) AS path_is_relative, CAST(NULL AS BIGINT) AS file_size_bytes, CAST(NULL AS BIGINT) AS footer_size, CAST(NULL AS VARCHAR) AS encryption_key, NULL format
```
pdet: format is still untyped, is that on purpose?

qsliu2017: Rebase mistake. Addressed.

pdet:
Thanks!

PR #908, opened 2026-03-27
fix: return move projection

qsliu2017:
use std::move to convert unique_ptr<Derive> to unique_ptr<Base>.

This pattern has been fixed once in #689. Enabling the warning makes the compiler enforce it going forward.


pdet:
Thanks!

PR #909, opened 2026-03-27
fix: use GenerateUUID() instead of SQL UUID() for schema initialization

qsliu2017:
Replace the SQL `UUID()` function call with a pre-generated UUID passed as a format argument. This uses the existing `GenerateUUID()` method (which produces UUIDv7) and avoids reliance on the `UUID()` SQL function, which may not be available in all metadata backends.


pdet:
LGTM! Can you just fix the conflict?

qsliu2017:
> LGTM! Can you just fix the conflict?

Done.

PR #910, opened 2026-03-27
Adding reading and writing of Iceberg V3 Deletion Vectors (i.e., puffin files)

pdet:
This PR adds support for deletion vectors (roaring bitmaps) to ducklake.

This PR adds support to both read and write of these files (mostly by adapting the code from [duckdb/iceberg](https://github.com/duckdb/duckdb-iceberg/blob/main/src/core/deletes/iceberg_deletion_vector.cpp), and by bringing our old merge delete code back.

The writing of these files is currently only focused on testing, so we have added a  `write_deletion_vectors` option that defaults to `False` with the following description: `[EXPERIMENTAL - do not use outside testing] Whether to write Iceberg V3 deletion vectors (puffin) instead of positional delete files (parquet)`.

We should eventually converge to also write these by default, but that is a bit more involved, as it will require them to be able to write deletes of multiple snapshots in the same file.

The tests we are skipping are related to the way we now write deletion files with multiple snapshot information. Since we now only need to preserve one file for multi-snapshots, old files are eagerly deleted, which does not work with the merge style of single-snapshot deletions. This should not affect the read path, and implementing it would increase code complexity for an unrealistic test path.

### DIFF
--- a/src/storage/ducklake_delete.cpp
+++ b/src/storage/ducklake_delete.cpp
@@ -133,6 +133,7 @@ static DuckLakeDeleteFile WriteDeleteFileInternal(ClientContext &context, InputT
 	DuckLakeDeleteFile delete_file;
 	delete_file.data_file_path = input.data_file_path;
 	delete_file.file_name = delete_file_path;
+	delete_file.format = DeleteFileFormat::PARQUET;
 	delete_file.delete_count = stats.row_count;
 	delete_file.file_size_bytes = stats.file_size_bytes;
 	delete_file.footer_size = stats.footer_size_bytes.GetValue<idx_t>();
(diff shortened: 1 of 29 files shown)


src/storage/ducklake_delete.cpp
```diff
@@ -335,11 +364,55 @@ bool DuckLakeDelete::TryDropFullyDeletedFile(DuckLakeTransaction &transaction, c
...
+	auto &existing_deletes = existing_delete_data.deleted_rows;
+	all_deletes.insert(existing_deletes.begin(), existing_deletes.end());
+
+	// clear the deletes from the map
+	delete_map->ClearDeletes(filename);
+
+	// set the delete file as overwriting existing deletes
+	delete_file.overwrites_existing_delete = true;
+
+	if (TryDropFullyDeletedFile(transaction, delete_file, data_file_info, all_deletes.size())) {
```
Tishj, line 385: I think this is what's breaking your time travel, since the new files don't have per-position snapshot tracking, the old snapshots still need to use these files.

src/storage/ducklake_delete.cpp
```diff
@@ -335,11 +364,55 @@ bool DuckLakeDelete::TryDropFullyDeletedFile(DuckLakeTransaction &transaction, c
...
+
 void DuckLakeDelete::FlushDeleteWithSnapshots(DuckLakeTransaction &transaction, ClientContext &context,
                                               DuckLakeDeleteGlobalState &global_state, const string &filename,
                                               const DuckLakeFileListExtendedEntry &data_file_info,
                                               DuckLakeDeleteData &existing_delete_data,
                                               const set<idx_t> &sorted_deletes, DuckLakeDeleteFile &delete_file) const {
+	auto &catalog = table.catalog.Cast<DuckLakeCatalog>();
+	auto &schema = table.ParentSchema().Cast<DuckLakeSchemaEntry>();
+	if (catalog.WriteDeletionVectors(schema.GetSchemaId(), table.GetTableId())) {
+		FlushMergedDeletionVector(transaction, context, global_state, filename, data_file_info, existing_delete_data,
```
Tishj, line 411: Continutation of https://github.com/duckdb/ducklake/pull/910/changes#r3000461176:
Along with this?

src/storage/ducklake_delete_filter.cpp
```diff
@@ -107,9 +109,36 @@ idx_t DuckLakeDeleteFilter::Filter(row_t start_row_index, idx_t count, Selection
...
+		result.deleted_rows.push_back(pos);
+	}
+	return result;
+}
+
 DeleteFileScanResult DuckLakeDeleteFilter::ScanDeleteFile(ClientContext &context, const DuckLakeFileData &delete_file,
                                                           optional_idx snapshot_filter_min,
                                                           optional_idx snapshot_filter_max) {
+	// Check if this is a puffin deletion vector file
+	if (StringUtil::EndsWith(delete_file.path, ".puffin")) {
```
Tishj: Nit: don't `delete_file`'s also have a `format` to check, instead of checking just the file extension?

Tishj (approved):
Thanks, looks good to me, just had a comment regarding the detection and also a thought about what the cause of the time travel issues you mentioned might be.

PR #941, opened 2026-04-05
Fix SET NOT NULL error message

### LINKED ISSUE #940: SET NULL failure message seems incorrect
### What happens?

See repro SQL script.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake, DATA_PATH '/tmp/rename_bug');
memory D CREATE TABLE ducklake.tbl(val INT);
memory D INSERT INTO ducklake.tbl VALUES(NULL);
memory D ALTER TABLE ducklake.tbl ALTER COLUMN val SET NOT NULL; 
Catalog Error:
Cannot SET NULL on column val - the column has NULL values
```
"Cannot SET NULL" -> "Cannot SET NOT NULL"

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/ducklake/issues/940

The `AlterTable` function is working on `SetNotNull`, all error messages should be updated.
Exception after this PR
```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake, DATA_PATH '/tmp/rename_bug');
memory D CREATE TABLE ducklake.tbl(val INT);
memory D INSERT INTO ducklake.tbl VALUES(NULL);
memory D ALTER TABLE ducklake.tbl ALTER COLUMN val SET NOT NULL; 
Catalog Error:
Cannot SET NOT NULL on column val - the column has NULL values
```

### DIFF
--- a/src/storage/ducklake_table_entry.cpp
+++ b/src/storage/ducklake_table_entry.cpp
@@ -517,22 +517,24 @@ unique_ptr<CatalogEntry> DuckLakeTableEntry::AlterTable(DuckLakeTransaction &tra
 	auto stats = GetTableStats(transaction);
 	if (!stats) {
 		throw CatalogException(
-		    "Cannot SET NULL on table %s - the table has transaction-local changes or no stats are available", name);
+		    "Cannot SET NOT NULL on table %s - the table has transaction-local changes or no stats are available",
+		    name);
 	}
 
 	auto column_stats = stats->column_stats.find(field_id.GetFieldIndex());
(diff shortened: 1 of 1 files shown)


pdet:
Thanks!

PR #943, opened 2026-04-05
Fix schema creation exception

### LINKED ISSUE #942: Creating already existent schema doesn't error
### What happens?

Creating already existing schema doesn't fail.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake, DATA_PATH '/tmp/rename_bug');
memory D CREATE SCHEMA ducklake.myschema;
memory D CREATE SCHEMA ducklake.myschema;
memory D 
```
Somehow the existing schema doesn't seem to fail, which should expect exception like "schema already exists".

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/ducklake/issues/942

After this PR SQL statements become
```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake, DATA_PATH '/tmp/rename_bug');
memory D CREATE SCHEMA ducklake.myschema;
memory D CREATE SCHEMA ducklake.myschema;
Catalog Error:
Schema myschema already exists
```

### DIFF
--- a/src/storage/ducklake_schema_entry.cpp
+++ b/src/storage/ducklake_schema_entry.cpp
@@ -23,6 +23,12 @@ DuckLakeSchemaEntry::DuckLakeSchemaEntry(Catalog &catalog, CreateSchemaInfo &inf
       data_path(std::move(data_path_p)) {
 }
 
+unique_ptr<CreateInfo> DuckLakeSchemaEntry::GetInfo() const {
+	auto result = SchemaCatalogEntry::GetInfo();
+	result->on_conflict = OnCreateConflict::IGNORE_ON_CONFLICT;
+	return result;
+}
+
 bool DuckLakeSchemaEntry::HandleCreateConflict(CatalogTransaction transaction, CatalogType catalog_type,
                                                const string &entry_name, OnCreateConflict on_conflict) {
 	auto existing_entry = GetEntry(transaction, catalog_type, entry_name);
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_schema_entry.cpp
```diff
@@ -23,6 +23,12 @@ DuckLakeSchemaEntry::DuckLakeSchemaEntry(Catalog &catalog, CreateSchemaInfo &inf
       data_path(std::move(data_path_p)) {
 }
 
+unique_ptr<CreateInfo> DuckLakeSchemaEntry::GetInfo() const {
+	auto result = SchemaCatalogEntry::GetInfo();
+	result->on_conflict = OnCreateConflict::IGNORE_ON_CONFLICT;
```
dentiny, line 28: Without this change, CI fails with error message "Schema main already exists"
https://github.com/duckdb/ducklake/actions/runs/23999971205/job/69994315064?pr=943
This overriden only triggers when DuckDB serializes an already-existing catalog entry back into a CreateInfo object

For example: https://github.com/duckdb/duckdb/blob/538d289feaab26a6f599c7d0e79a50b6722955b9/src/execution/operator/persistent/physical_export.cpp#L36

PR #948, opened 2026-04-07
Normalize ducklake path joins

txbm:
Fix all unsafe path handling to prevent malformed paths from erroneously causing data loss and other mayhem


txbm:
This patch was verified via a full catalog setup + data insertion + merge + cleanup cycle and prevents incidental data loss and other problems caused by malformed paths persisting in the catalog

pdet:
Hi @txbm could you add tests that trigger the issue?

txbm:
> Hi @txbm could you add tests that trigger the issue?

Added reproducing test cases and rebased against `main`

PR #951, opened 2026-04-07
Explicitly set C++17 for DuckLake

txbm:
Seems like the build implicitly requires CXX 17 but it was not made explicit anywhere. This makes it explicit if that is the intention


pdet:
Hi @txbm thanks for the PR, the `main` branch was neglected for a while due to the `DuckLake 1.0` release, I've updated it today at https://github.com/duckdb/ducklake/pull/1008

PR #953, opened 2026-04-07
Respect DuckDB's static extension build flag: EXTENSION_STATIC_BUILD

txbm:
DuckDB has an existing flag for extensions to opt in or out of static building separate from loadable extension building. This patch adds support for honoring that existing flag

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,14 +16,16 @@ include_directories(${roaring_DIR}/../../include)
 add_subdirectory(src)
 set(EXTENSION_SOURCES ${ALL_OBJECT_FILES})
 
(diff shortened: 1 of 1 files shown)


carlopi (commented):
Hi, I am not sure I understand what's the goal here.

You want to build ducklake ONLY as loadable extension? But then you should already be good, you can just use a separate duckdb binary.

Basically, I am not sure I understand the use case, but I would think it's a better fit for a fix in duckdb/duckdb than in each extension. It doesn't seem to be `ducklake` specific.

carlopi:
I'll close for now, it seems to me this should be moved to duckdb, but I don't see the full picture / intended goal.

Issue #957, opened 2026-04-08
BUG: delete table/view in one transaction clears all new tables/views


dentiny:
### What happens?

Hi team, I noticed the follow SQL error doesn't seem to be reasonable
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D BEGIN;
memory D CREATE TABLE dl.main.t1 (x INT);
memory D CREATE TABLE dl.main.t2 (y INT);
memory D DROP TABLE dl.main.t1;
memory D SELECT * FROM dl.main.t2;
Catalog Error:
Table with name t2 does not exist!
Did you mean "information_schema.tables"?

LINE 1: SELECT * FROM dl.main.t2;
                      ^
```
Here we only delete table `t1` and `t2` should be visible.

### To Reproduce

The same issues happens for view
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D BEGIN;
memory D CREATE VIEW dl.main.v1 AS SELECT 1 AS a;
memory D CREATE VIEW dl.main.v2 AS SELECT 2 AS b;
memory D DROP VIEW dl.main.v1;
memory D SELECT * FROM dl.main.v2;
Catalog Error:
Table with name v2 does not exist!
Did you mean "information_schema.views"?

LINE 1: SELECT * FROM dl.main.v2;
                      ^
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #958, opened 2026-04-08
Fix drop table in single transaction

dentiny:
Closes https://github.com/duckdb/ducklake/issues/957

If I don't mis-understand, when a new table/view is created inside of a transaction, it's inserted into `new_tables` map

This is where we find the `DuckLakeCatalogSet`
https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L3048-L3052

And when we create a new entry, we add into the `CatalogSet` returned above
https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L2803-L2807

The bug is somehow when we delete a table/view, we remove the whole catalog set from the transaction-local `new_tables`, even if it's not empty:
- table: https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L2839
- view: https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L2854
- In my PR, I check the emptiness of the catalog set before removing

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -2844,7 +2844,9 @@ void DuckLakeTransaction::DropTable(DuckLakeTableEntry &table) {
 		// if we have written any files for this table - clean them up
 		auto context_ref = context.lock();
 		local_changes.CleanupFiles(*context_ref, table_id);
-		new_tables.erase(schema_entry);
+		if (schema_entry->second->GetEntries().empty()) {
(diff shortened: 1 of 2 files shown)


pdet:
Thanks!

Issue #961, opened 2026-04-08
BUG: setting default for multiple times in a single transaction crashes


dentiny:
### What happens?

A weird case: when I setting the default value for multiple times in a single transaction, later access to the same table crashes. See repro sql for details.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl (a INT);
memory D BEGIN;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 99;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 9;
memory D COMMIT;
memory D BEGIN;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 8;
Catalog Error:
Column with name a already exists!
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #962, opened 2026-04-08
Fix repeated column update in one transaction

dentiny:
Closes https://github.com/duckdb/ducklake/issues/961

When I was debugging the issue, I used the dummy way to add logging for `ducklake_column` system table, for the failure sql.
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl (a INT);
memory D BEGIN;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 99;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 9;
memory D COMMIT;
memory D BEGIN;
memory D ALTER TABLE dl.tbl ALTER a SET DEFAULT 8;
Catalog Error:
Column with name a already exists!
```

Here's what I found [here](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L2241-L2250):
- At commit phase, the column `a` is dropped twice, and re-created twice
- The drop column SQL: 
```sql
WITH dropped_cols(tid, cid) AS (
VALUES (1, 1), (1, 1)
)
UPDATE {METADATA_CATALOG}.ducklake_column
SET end_snapshot = {SNAPSHOT_ID}
FROM dropped_cols
WHERE table_id=tid AND column_id=cid AND end_snapshot IS NULL;
```
- The append column SQL: 
```sql
INSERT INTO {METADATA_CATALOG}.ducklake_column VALUES (1, {SNAPSHOT_ID}, NULL, 1, 1, 'a', 'int32', NULL, '99', true, NULL, 'literal', 'duckdb'),(1, {SNAPSHOT_ID}, NULL, 1, 1, 'a', 'int32', NULL, '9', true, NULL, 'literal', 'duckdb');
```
- As we can easily tell, the column `a` is dropped twice and appended twice, all the them are recorded in the `ducklake_column`
- Next time when we reconstruct the latest snapshot [here](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_catalog.cpp#L326) inside of [MetadataManager](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_metadata_manager.cpp#L529-L555), we get two columns, [inserting both into the same table](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_catalog.cpp#L376) leads to crash

In this PR, the goal I want to achieve is to only insert one single row (the latest value) into `ducklake_column`, so later access won't access single column for multiple times.

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1533,6 +1533,8 @@ void DuckLakeTransaction::GetNewTableInfo(DuckLakeCommitState &commit_state, Duc
 	}
 
 	set<FieldIndex> columns_handled_by_later_ops;
+	// Maps from field index to the number of alter operations for that field.
+	map<FieldIndex, idx_t> field_alter_count;
 	for (idx_t table_idx = 0; table_idx < tables.size(); table_idx++) {
 		auto &table = tables[table_idx].get();
 		auto local_change = table.GetLocalChange();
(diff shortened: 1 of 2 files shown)


pdet:
Thanks! LGTM!

Issue #964, opened 2026-04-08
BUG: dropping schema succeeds with table/view/macro inside


dentiny:
### What happens?

I found in certain cases, dropped a schema with table inside would succeed, with no `CASCADE` specification.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE SCHEMA dl.my_schema;
memory D BEGIN;
memory D CREATE TABLE dl.my_schema.new_table (i INTEGER);
memory D DROP SCHEMA dl.my_schema;
memory D COMMIT;
-- succeed
memory D 
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #965, opened 2026-04-08
Check transaction-local entities before drop

dentiny:
Closes https://github.com/duckdb/ducklake/issues/964

Hi team, I met this weird issue that dropping schema succeeds when there's table inside; after some investigation I found we currently don't consider transaction local entities on drop, only committed items.

### DIFF
--- a/test/sql/catalog/drop_schema_transaction_local_entries.test
+++ b/test/sql/catalog/drop_schema_transaction_local_entries.test
@@ -1,0 +1,77 @@
+# name: test/sql/catalog/drop_schema_transaction_local_entries.test
+# description: Test that DROP SCHEMA correctly checks transaction-local entries for dependents
+# group: [catalog]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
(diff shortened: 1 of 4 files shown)


test/sql/catalog/drop_schema_transaction_local_entries.test
```diff
@@ -0,0 +1,77 @@
...
+statement ok
+CREATE MACRO ducklake.my_schema.add_one(x) AS x + 1
+
+statement error
+DROP SCHEMA ducklake.my_schema
+----
+there are entries that depend on it
+
+statement ok
+DROP SCHEMA ducklake.my_schema CASCADE
```
pdet, line 84: Can we check everything was dropped?

test/sql/catalog/drop_schema_transaction_local_entries.test
```diff
@@ -0,0 +1,77 @@
...
+
+# Test when there's a transaction-local macro in the schema, it cannot be dropped.
+statement ok
+BEGIN
+
+statement ok
+CREATE SCHEMA ducklake.my_schema
+
+statement ok
+CREATE MACRO ducklake.my_schema.add_one(x) AS x + 1
```
pdet, line 76: Can we also have a table macro?

pdet (commented):
Thanks for the PR, just two small points.

dentiny:
> just two small points.

Test cases updated :)

pdet:
Thanks!

Issue #967, opened 2026-04-08
BUG: view creation containing catalog breaks incorrect result


dentiny:
### What happens?

When we create a view, we [replace](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_schema_entry.cpp#L153) catalog name inside of view creation sql with a special token, and [replace it back](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_view_entry.cpp#L88) when the view gets queries.

That brings a limitation: if the view creation sql contains catalog name we're screwed.

### To Reproduce

```sql
memory D ATTACH 'ducklake:dl_view.db' AS dl (DATA_PATH '/tmp/dl_view');
memory D CREATE TABLE dl.main.t(i INTEGER);
memory D CREATE VIEW dl.main.v AS SELECT 'dl.hello' AS val FROM dl.main.t;
memory D DETACH dl;
memory D ATTACH 'ducklake:dl_view.db' AS other_name (DATA_PATH '/tmp/dl_view');
memory D SELECT * FROM other_name.main.v;
┌─────────┐
│   val   │
│ varchar │
└─────────┘
  0 rows 
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #968, opened 2026-04-08
Fix sql string replacement for view

dentiny:
Closes https://github.com/duckdb/ducklake/issues/967

The bug is:
- When a view is created, DuckLake replaces the catalog name with a placeholder `{DUCKLAKE_CATALOG}` [here](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_schema_entry.cpp#L153); when the view gets queries we replace it back [here](https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_view_entry.cpp#L88)
- But the thing is, if the original sql contains a catalog name (in detail, `dl` is part of the view creation sql, and it's also the catalog name), string-based replacement doesn't work

The fix I proposed in this PR is to create a quote-aware string replacement.
Disclaimer: the string replacement function is written by opus.

### DIFF
--- a/src/common/ducklake_util.cpp
+++ b/src/common/ducklake_util.cpp
@@ -350,4 +350,46 @@ bool DuckLakeUtil::HasInlinedSystemColumnConflict(const vector<DuckLakeColumnInf
 	return false;
 }
 
+string DuckLakeUtil::ReplaceSkippingQuotes(const string &sql, const string &from, const string &to) {
+	if (from.empty()) {
+		return sql;
+	}
+	string result;
+	result.reserve(sql.size());
+	idx_t i = 0;
+	while (i < sql.size()) {
+		char c = sql[i];
+		if (c == '\'' || c == '"') {
+			// Now we're inside a quoted region, so copy everything until the matching close quote.
+			char quote = c;
+			result += c;
+			i++;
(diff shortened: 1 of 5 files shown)


src/common/ducklake_util.cpp
```diff
@@ -350,4 +350,46 @@ bool DuckLakeUtil::HasInlinedSystemColumnConflict(const vector<DuckLakeColumnInf
 	return false;
 }
 
+string DuckLakeUtil::ReplaceSkippingQuotes(const string &sql, const string &from, const string &to) {
+	if (from.empty()) {
```
pdet, line 355: Can we use Parser::Tokenize()?

pdet (commented):
Thanks for the PR, just one comment about it

dentiny:
I understand the urgency/necessity to get PRs in ASAP before release, I will address comments later today.

src/common/ducklake_util.cpp
dentiny: sure, updated

pdet:
Looks great! Thanks for the rewrite!

Issue #917, opened 2026-03-30
`DELETE` Silently Fails for Rows with Duplicate IDs Across Flush-Created Files


hello-world-bfree:
### What happens?

`DELETE FROM table WHERE <condition>` silently fails (no error, zero rows deleted) when matching rows have the same logical ID in multiple parquet files. SELECT correctly finds the same rows. Range DELETEs and UPDATEs work. The bug occurs on tables with files created by `ducklake_flush_inlined_data` that contain embedded `_ducklake_internal_snapshot_id` / `_ducklake_internal_row_id` columns. Reproduces on both v1.5.1 and latest main (`710adf3d05`).

## Environment

- DuckDB v1.5.1 (also reproduced on v1.5.2-dev71 / latest main)
- DuckLake with PostgreSQL metadata catalog
- S3-backed data storage
- `DATA_INLINING_ROW_LIMIT 700`
- `expire_snapshots` and `cleanup_old_files` have been run
- Table has ~936 active data files, ~25M rows
- Mix of file types: bulk-loaded (no embedded columns), flush-created (with embedded `_ducklake_internal_row_id` and `_ducklake_internal_snapshot_id`), and small CDC files

### Basic demonstration

```sql
-- Row exists in 2 files:
SELECT count(*) FROM user_settings_v2 WHERE id = 41307875;
-- Returns: 2

-- DELETE silently does nothing:
BEGIN;
DELETE FROM user_settings_v2 WHERE id = 41307875;
SELECT count(*) FROM user_settings_v2 WHERE id = 41307875;
-- Returns: 2 (unchanged)
ROLLBACK;
```

### What works vs what fails

| Operation | Result |
|-----------|--------|
| `DELETE WHERE id = X` (X has ≥2 copies across files) | Silent no-op ✗ |
| `DELETE WHERE rowid = <specific>` (row's id has dups) | Silent no-op ✗ |
| `DELETE WHERE rowid BETWEEN a AND b` (same rows) | Works ✓ |
| `UPDATE WHERE id = X` (X has ≥2 copies) | Finds rows ✓ |
| `SELECT WHERE id = X` (X has ≥2 copies) | Finds rows ✓ |
| `DELETE WHERE id = Y` (Y has 1 copy) | Works ✓ |
| `DELETE WHERE id = Y` (Y has 2 copies, one is txn-local) | Works ✓ |
| Bulk `DELETE WHERE rowid >= file_start AND rowid < file_end` | Works ✓ |

[...]

PR #969, opened 2026-04-08
fix: deletion bugs (fixes issue #917)

hello-world-bfree:
## Summary

This PR fixes four separate deletion bugs that showed up under the same [#917](https://github.com/duckdb/ducklake/issues/917) umbrella:

  - `ducklake_rewrite_data_files` skipped files whose deletes only existed in metadata-backed inlined delete tables
  - `DELETE` on some `ducklake_add_data_files` tables failed because the local delete-file directory was never created
  - legacy `v0.3 -> v0.4` files with `mapping_id = NULL` could break after later column renames
  - `ducklake_flush_inlined_data` could write wrong positional delete files for sorted tables

These were related in symptoms, but they were not the same bug.

### Rewrite path ignored metadata-only file deletes

`rewrite_data_files` only looked at rows in `ducklake_delete_file` when choosing rewrite candidates and validating output row counts.

If a file only had metadata-backed inlined file deletes, duckLake could treat it as if it had no deletes at all. That meant:

  - rewrite candidates got skipped
  - imported files from `ducklake_add_data_files` could stay stuck in the deleted-but-never-rewritten state
  - `merge_adjacent_files` could make decisions without seeing those deletes

### Delete writers assumed the table directory already existed

For local `ducklake_add_data_files` tables, the first write might actually be a delete file. If the table directory had never been created, writing the delete artifact failed.

### Legacy files without a mapping stopped surviving renames

Some `v0.3 -> v0.4` files can still have `mapping_id = NULL`. If those files also do not carry field IDs, and the table later goes through a rename, binding by current column names is wrong for the old file. That gives the "new rows delete, old rows do not" kind of failure.

There was also a secondary issue where `GetFilesExtended()` could seed the normal file list with partial metadata, which made mapping-related bugs easier to trigger.

### Sorted flush used the wrong row positions

`ducklake_flush_inlined_data` was building delete positions from:

```sql
ROW_NUMBER() OVER (ORDER BY row_id, begin_snapshot)
```

That only works if the written parquet file is also in row-id order. For sorted tables it is not. The file is in sort-key order, so the delete file could target the wrong rows.

## What this changes

  - treat metadata-backed inlined file deletes like regular delete files during rewrite planning and validation
  - pass inlined file deletions through the compaction scan path
  - count inlined deleted rows in rewrite row-count accounting
  - avoid merging files that still have any kind of delete state
  - create local delete directories before writing delete files or deletion vectors
  - add a legacy fallback name map based on the historical table schema at the file snapshot when mapping_id is missing
  - stop back-filling the normal file list from GetFilesExtended()
  - build flush delete positions from the actual written parquet file order by scanning _ducklake_internal_row_id and _ducklake_internal_snapshot_id

## Notes

  - The legacy mapping fix is about v0.3 -> v0.4 histories, not v0.2.
  - The sorted flush fix adds a readback of the written parquet file on the flush path. That is extra work, but it is the only safe way to build positional delete files once file order is no longer row-id order.
  - This PR does not finish the separate ducklake_table_deletions / ducklake_table_changes diagnostics issue for repeated row IDs in one partial file. That is a follow-up issue in the diagnostics path, not the normal read path.

  ## Tests

  - test/sql/delete/delete_add_files_creates_delete_dir.test
  - test/sql/delete/delete_legacy_missing_mapping_after_rename_add_files.test
  - test/sql/rewrite_data_files/test_rewrite_inlined_file_deletes.test
  - test/sql/rewrite_data_files/test_rewrite_inlined_file_deletes_add_files.test
  - test/sql/sorted_table/data_inlining_flush_sorted_sequential_updates.test
  - full DuckLake debug SQLLogicTest pass, including *.test_slow

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -238,9 +238,10 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 			// (does not apply to REWRITE_DELETES - delete files must be rewritten regardless of data file size)
 			continue;
 		}
-		if ((!candidate.delete_files.empty() && type == CompactionType::MERGE_ADJACENT_TABLES) ||
-		    candidate.file.end_snapshot.IsValid() || candidate.has_inlined_deletions) {
-			// Merge Adjacent Tables doesn't perform the merge if delete files are present
+		if (((!candidate.delete_files.empty() || !candidate.inlined_file_deletions.empty()) &&
+		     type == CompactionType::MERGE_ADJACENT_TABLES) ||
+		    candidate.file.end_snapshot.IsValid()) {
+			// Merge Adjacent Tables doesn't perform the merge if any deletes are present
 			continue;
 		}
 		// construct the compaction group for this file - i.e. the set of candidate files we can compact it with
@@ -415,9 +416,11 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
 	for (auto &source : source_files) {
 		DuckLakeFileListEntry result;
 		result.file = source.file.data;
+		result.file_id = source.file.id;
 		result.row_id_start = source.file.row_id_start;
 		result.snapshot_id = source.file.begin_snapshot;
 		result.mapping_id = source.file.mapping_id;
+		result.inlined_file_deletions = source.inlined_file_deletions;
 		switch (type) {
 		case CompactionType::REWRITE_DELETES: {
 			if (!source.delete_files.empty()) {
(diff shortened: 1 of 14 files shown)


hello-world-bfree:
@pdet I have a branch rebased to v1.5:variegata and ready to go as well. That branch doesn't have `DuckLakeTableEntry::GetPartitionSQLExpressions()` or `DuckLakePartitionUtils::BuildPartitionFilter()` though, so I had to reconstruct them. That's the only difference between the two.

pdet:
> @pdet I have a branch rebased to v1.5:variegata and ready to go as well. That branch doesn't have `DuckLakeTableEntry::GetPartitionSQLExpressions()` or `DuckLakePartitionUtils::BuildPartitionFilter()` though, so I had to reconstruct them. That's the only difference between the two.

Nice, can you push the 1.5? We will be releasing soon, will be cool to get this in as you managed to reproduce it.

PR #970, opened 2026-04-08
fix: deletion bugs variegata (fixes issue #917)

hello-world-bfree:
see https://github.com/duckdb/ducklake/pull/969


hello-world-bfree:
@pdet Much appreciated! 🙇

pdet:
Hi @hello-world-bfree thanks for the PR, given this would need to get in tomorrow max to go for the release, I took the liberty of extracting your tests and providing the fixes in here: https://github.com/duckdb/ducklake/pull/972

I've decided to skip the tests related to rewriting over inlined deletes, as this is not critical for the release (but definetly a great to have, for the next version). Maybe you can adjust this PR based on mine to have just the rewriting path, then we can have a look at it after the release! Many thanks again for your work!

hello-world-bfree:
> I've decided to skip the tests related to rewriting over inlined deletes, as this is not critical for the release (but definetly a great to have, for the next version). Maybe you can adjust this PR based on mine to have just the rewriting path, then we can have a look at it after the release! Many thanks again for your work!

Thanks @pdet ! Will do

hello-world-bfree:
Hey @pdet! Thanks for getting the fixes out. Looks like there is still some inline deleting that doesn't occur. We're still getting some dupes because of it (cdc use case). This PR resolves it and adds the test, sitting on top of what you've already got merged in. Thanks again!!

[image](https://github.com/user-attachments/assets/9f90a896-892e-4c63-be7a-8ad8e88f9297)

image

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1914,17 +1914,31 @@ ORDER BY data.begin_snapshot, data.row_id_start, data.data_file_id, del.begin_sn
 		file_entry.delete_files.push_back(std::move(delete_file));
 	}
 
-	// Check for inlined deletions and mark affected files
-	// Gather file IDs first, then query only for existence
-	vector<idx_t> file_ids;
-	file_ids.reserve(files.size());
+	// Load inlined deletions for active files so rewrite compaction can treat them the same as delete files.
+	auto inlined_deletions = ReadInlinedFileDeletions(table_id, snapshot);
```
pdet: This also runs for `MERGE_ADJACENT` right? If so, wouldn't that introduce a performance regression to it?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1807,9 +1807,9 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
 	string select_list = data_select_list + ", " + delete_select_list;
 	string deletion_threshold_clause;
 	if (type == CompactionType::REWRITE_DELETES) {
-		deletion_threshold_clause = StringUtil::Format(
-		    " AND CAST(del.delete_count AS FLOAT)/CAST(data.record_count AS FLOAT) >= %f and data.end_snapshot is null",
-		    deletion_threshold);
+		// Filter current data files in SQL, then apply the delete threshold in C++ so we can include
+		// metadata-only inlined file deletions as rewrite candidates.
+		deletion_threshold_clause = " AND data.end_snapshot is null";
```
pdet, line 1812: Can't we still do that in SQL?

pdet (commented):
Thanks for the fix! LGTM, just had two few comments that I think might create perf regressions.

src/storage/ducklake_metadata_manager.cpp
hello-world-bfree: For sure! Great catch. Should be resolved now. Only the `REWRITE_DELETES` compaction type runs the full row-id fetch.

src/storage/ducklake_metadata_manager.cpp
hello-world-bfree: unfortunately not without a correlated subquery since all of the inlined deletions are in the `ducklake__inlined_delete_*` table. Could do a hybrid to keep the common case ccheaper with the sql filter but that complexity might not be worth it. Wouldn't think the performance will be too much different

pdet:
Thanks!

Issue #973, opened 2026-04-09
BUG: SET DEFAULT then RENAME on the sort column in the same transaction fails


dentiny:
### What happens?

Hi team, I attach a SQL test file which unexpected breaks.

### To Reproduce

```sql
# name: test/sql/sorted_table/set_default_preserves_sort_key.test
# description: SET DEFAULT followed by RENAME on the sort column in the same transaction
# group: [sorted_table]

require ducklake

require parquet

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '__TEST_DIR__/set_default_preserves_sort_key/')

statement ok
USE ducklake;

statement ok
CREATE TABLE t (a INTEGER, b INTEGER);

statement ok
INSERT INTO t VALUES (3, 30), (1, 10);

statement ok
ALTER TABLE t SET SORTED BY (a);

statement ok
BEGIN;

statement ok
ALTER TABLE t ALTER COLUMN b SET DEFAULT 42;

statement ok
ALTER TABLE t RENAME COLUMN a TO a_renamed;

statement ok
COMMIT;

statement ok
INSERT INTO t VALUES (2, 20);

query II
SELECT * FROM t ORDER BY a_renamed
----
1	10
2	20
3	30
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #974, opened 2026-04-09
Fix missing assignment for sort data

dentiny:
Closes https://github.com/duckdb/ducklake/issues/973

I accidentally found the missing assignment when reading the code, should be straightforward.
Same as the other constructor: https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_table_entry.cpp#L73


pdet:
Thanks!

Issue #975, opened 2026-04-09
BUG: deletion files left behind


dentiny:
### What happens?

Hi team, I think when we rollback a transaction, we should cleanup both data files and deletion files.
https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L571-L573

How I find the issue: I asked opus to repeated do append and deletion in committed and rollbacked transaction, and found dangling files.

### To Reproduce

```sql
# name: test/sql/cleanup/drop_table_cleans_delete_files.test
# description: Dropping a transaction-local table must clean up both deletion files and data files
# group: [cleanup]

require ducklake

require parquet

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

test-env DATA_PATH __TEST_DIR__

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_drop_delete_cleanup', DATA_INLINING_ROW_LIMIT 0)

statement ok
BEGIN

statement ok
CREATE TABLE ducklake.tbl AS SELECT i id FROM range(1000) t(i);

# data file exists
query I
SELECT COUNT(*) FROM glob('${DATA_PATH}/ducklake_drop_delete_cleanup/main/tbl/*.parquet')
----
1

statement ok
DELETE FROM ducklake.tbl WHERE id % 2 = 0;

# delete file exists
query I
SELECT COUNT(*) FROM glob('${DATA_PATH}/ducklake_drop_delete_cleanup/main/tbl/*-delete.parquet')
----
1

statement ok
DROP TABLE ducklake.tbl

# both data files and delete files should be cleaned up
query I
SELECT COUNT(*) FROM glob('${DATA_PATH}/ducklake_drop_delete_cleanup/main/tbl/*.parquet')
----
0

query I
SELECT COUNT(*) FROM glob('${DATA_PATH}/ducklake_drop_delete_cleanup/main/tbl/*-delete.parquet')
----
0

statement ok
COMMIT
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

[...]

PR #976, opened 2026-04-09
Delete deletion files for local table change

dentiny:
Closes https://github.com/duckdb/ducklake/issues/975

The code change is borrowed from another file cleanup function here: https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_transaction.cpp#L67-L79


pdet:
Thanks!

Issue #984, opened 2026-04-10
BUG: macros exists after snapshot expiration


dentiny:
### What happens?

Hi team, I found macros (both scalar and table) exist even after they've been dropped and snapshot expired.
See attached SQL statements for details.

### To Reproduce

```sql
# name: test/sql/compaction/expire_snapshots_drop_macro.test
# description: test expire_snapshots cleans up dropped macro metadata (scalar and table macros)
# group: [compaction]

require ducklake

require parquet

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

test-env DATA_PATH __TEST_DIR__

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/expire_snapshots_drop_macro_files', METADATA_CATALOG 'metadata')

statement ok
USE ducklake;

statement ok
CREATE TABLE tbl(i INTEGER);

# snapshot: create scalar macro
statement ok
CREATE MACRO add_one(x) AS (x + 1);

# snapshot: create table macro
statement ok
CREATE MACRO get_all() AS TABLE SELECT * FROM tbl;

# snapshot: drop both macros
statement ok
DROP MACRO add_one;

statement ok
DROP MACRO TABLE get_all;

# macro metadata should still exist (referenced by historical snapshots)
query I
SELECT COUNT(*) FROM metadata.ducklake_macro;
----
2

query I
SELECT COUNT(*) FROM metadata.ducklake_macro_impl;
----
2

query I
SELECT COUNT(*) FROM metadata.ducklake_macro_parameters;
----
1

# expire all old snapshots
statement ok
CALL ducklake_expire_snapshots('ducklake', older_than => NOW());

# after expiration, all macro metadata should be cleaned up
query I
SELECT COUNT(*) FROM metadata.ducklake_macro;
----
0

query I
SELECT COUNT(*) FROM metadata.ducklake_macro_impl;
----
0

query I
SELECT COUNT(*) FROM metadata.ducklake_macro_parameters;
----
0
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

[...]

PR #985, opened 2026-04-10
Fix macros cleanup on snapshot expiration

dentiny:
Closes https://github.com/duckdb/ducklake/issues/984

The root cause and fix should be straightforward: when we delete snapshots, we should consider macros related tables.


pdet:
Thanks!

Issue #988, opened 2026-04-10
BUG: setting comment on a table/column expires all other tags


dentiny:
### What happens?

See my attached SQL repro, should be straightforward.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake);
memory D CREATE TABLE ducklake.tbl (i INT)l
         
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake);
memory D CREATE TABLE ducklake.tbl (i INT);
memory D INSERT INTO __ducklake_metadata_ducklake.ducklake_tag VALUES (1, 2, NULL, 'owner', 'name');
memory D COMMENT ON TABLE ducklake.tbl IS 'second comment';
memory D SELECT key, value, end_snapshot FROM __ducklake_metadata_ducklake.ducklake_tag;
┌─────────┬────────────────┬──────────────┐
│   key   │     value      │ end_snapshot │
│ varchar │    varchar     │    int64     │
├─────────┼────────────────┼──────────────┤
│ owner   │ name           │            2 │
│ comment │ second comment │         NULL │
└─────────┴────────────────┴──────────────┘
```
The <owner, name> tag should have end snapshot NULL.

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

336591eafea08a7297752c6e1135b04f35b31dee

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #989, opened 2026-04-10
Fix set tag SQL

dentiny:
Closes https://github.com/duckdb/ducklake/issues/988

The bug is when we overwrite snapshot for the new tag written, the WHERE clause only checks table id and column id, instead of new key as well, thus all key-value pairs in the same table / column are overwritten.
https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_metadata_manager.cpp#L3824-L3831
https://github.com/duckdb/ducklake/blob/336591eafea08a7297752c6e1135b04f35b31dee/src/storage/ducklake_metadata_manager.cpp#L3783-L3790

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -3799,7 +3799,7 @@ VALUES %s
 UPDATE {METADATA_CATALOG}.ducklake_tag
 SET end_snapshot = {SNAPSHOT_ID}
 FROM overwritten_tags
-WHERE object_id=tid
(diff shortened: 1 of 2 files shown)


dentiny:
Hmm I think the failed CI test has nothing to do with my change
```sh
1. test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54
Wrong result in query! (test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54)!
SELECT COUNT(*) FROM ducklake.t0 WHERE a < 5;
2 <> 0
Expected result:
0

Actual result:
2
```

PR #990, opened 2026-04-10
[minor] Delete dead declaration

dentiny:
The declaration is not used or defined anywhere.
```sh
hjiang@hjiangs-MacBook-Pro ducklake % git grep "FlushSettingChanges"
src/include/storage/ducklake_transaction.hpp:   void FlushSettingChanges();
```
Question: I think we should add `-Wunused-variable -Wunused-function` compilation option


pdet:
Thanks!

pdet:
I think adding `-Wunused-variable -Wunused-function` as  compilation flags make sense!

PR #996, opened 2026-04-11
Add query result helpers and guard query call sites

txbm:
A number of previous bugs as well as current/latent ones are all caused by inconsistent or ad-hoc handling of the `QueryResult` object. Specifically, the pattern of checking for and handling possible execution errors before attempting to access or index the `result`, which will crash if the `QueryResult` is carrying an execution error.

This PR defines a minimal set of functional utilities to run queries with a defined error fallback message and ensure that each result object is error-checked-and-handled before the result is returned to the caller. This ensures that callsites can rely on either receiving a valid result object that can be safely further referenced or they will get the expected error as explicitly defined by the caller (and therefore also handled by the caller).

Consistent application of the utility functions will prevent crashes caused by unchecked query errors and reduces boilerplate. They follow a simple pattern of `x_or_y`, making them easy to compose and extend over time as needed.

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -168,7 +168,8 @@ private:
 };
 
 void DuckLakeFileProcessor::ReadParquetFullMetadata(const string &glob, vector<DuckLakeDataFile> &written_files) {
-	auto result = transaction.Query(StringUtil::Format(R"(
+	auto result = result_or_throw(
+	    transaction.Query(StringUtil::Format(R"(
 SELECT 
     list_transform(parquet_file_metadata, lambda x: struct_pack(
         file_name := x.file_name,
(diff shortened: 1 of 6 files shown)


pdet:
I like the idea of having a helper function to avoid missing error checking, but I think the lambda functions are unnecessary.  Also some functions are defined but never used? (e.g., `query_rows_or_throw`).

Maybe something like, would be sufficient?

```cpp
inline unique_ptr<QueryResult> result_or_throw(unique_ptr<QueryResult> result, const string &err) {                                                                                                               
      if (result->HasError()) {                                                                                                                                                                                     
          result->GetErrorObject().Throw(err);
      }                                                                                                                                                                                                             
      return result;
  }
```

txbm:
> I like the idea of having a helper function to avoid missing error checking, but I think the lambda functions are unnecessary. Also some functions are defined but never used? (e.g., `query_rows_or_throw`).
> 
> Maybe something like, would be sufficient?
> 
> ```c++
> inline unique_ptr<QueryResult> result_or_throw(unique_ptr<QueryResult> result, const string &err) {                                                                                                               
>       if (result->HasError()) {                                                                                                                                                                                     
>           result->GetErrorObject().Throw(err);
>       }                                                                                                                                                                                                             
>       return result;
>   }
> ```

Updated to lambda-free impl and dropped extra helper. Rebased against `main`

PR #997, opened 2026-04-11
non-native, nested and bit types serdes fixes to prevent Postgres catalog from crashing on inlined data

txbm:
The Postgres catalog needs different serialization/deserialization from/to SQL for non-native, nested or `BIT` types.

Currently it crashes when trying to read inlined data rows containing any such types.

This adds specific conversions and for non-native, nested and bit types.

This was tested end-to-end via:
1. catalog bootstrap
2. inline data writes to the Postgres catalog
3. inline data flush out to lake storage

Payloads contained a series of complex and nested types on input, flushed parquet files confirmed to contain exact schema post-flush.


pdet:
Hi @txbm thanks for the PR, could you also add the tests that trigger the issue?

txbm:
> Hi @txbm thanks for the PR, could you also add the tests that trigger the issue?

Added reproducing test cases and rebased against `main`

Issue #740, opened 2026-02-03
Adding column and renaming inside txn fails when inlining


Alex-Monahan:
### What happens?

I found an odd bug with alter table statements within a transaction when doing some testing on the next phase of sorting. This is a fairly artificial example (not a lot of scenarios where it would be needed). 

If you are using inlining, adding a column and renaming it within the same transaction will cause the table to not be readable after commit (select * from table will fail). 

### To Reproduce

```sql
# name: test/sql/sorted_table/data_inlining_flush_sorted_transaction_expression_column_changes.test
# description: Test SET SORTED BY with expression when columns are dropped and re-added in same transaction
# group: [sorted_table]

require ducklake

require parquet

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

test-env DATA_PATH __TEST_DIR__


statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/ducklake_inlining_flush_expression_column_changes', DATA_INLINING_ROW_LIMIT 10)

# Create table with VARCHAR columns suitable for concatenation expression
statement ok
CREATE TABLE ducklake.expression_column_test (unique_id INTEGER, sort_key_1 VARCHAR, sort_key_2 VARCHAR);

# Insert data (will be inlined initially)
statement ok
INSERT INTO ducklake.expression_column_test (unique_id, sort_key_1, sort_key_2)
VALUES
    (0, 'a', 'val0'),
    (1, 'c', 'val1'),
    (2, 'b', 'val2'),
    (3, 'a', 'val3'),
    (4, 'c', 'val4');

statement ok
BEGIN

statement ok
ALTER TABLE ducklake.expression_column_test ADD COLUMN sort_key_2_new VARCHAR;

statement ok
ALTER TABLE ducklake.expression_column_test RENAME COLUMN sort_key_2_new TO sort_key_3;

statement ok
CALL ducklake_flush_inlined_data('ducklake', table_name => 'expression_column_test');

statement ok
COMMIT

# This is currently erroring out with:
# Binder Error: Failed to read inlined data from DuckLake: Referenced column "sort_key_2_new" not found in FROM clause!
# Candidate bindings: "sort_key_2", "sort_key_1", "sort_key_3"

[...]

PR #998, opened 2026-04-11
Fix column rename in same transaction

dentiny:
Closes https://github.com/duckdb/ducklake/issues/740

Key transaction SQL pasted here
```sql
BEGIN
ALTER TABLE ducklake.expression_column_test ADD COLUMN sort_key_2_new VARCHAR;
ALTER TABLE ducklake.expression_column_test RENAME COLUMN sort_key_2_new TO sort_key_3;
COMMIT
```

**Root cause**
When we rename column after adding the column in one single transaction, DuckLake handles them in this way:
- process add column: add `sort_key_2_new`
- process rename column: drop `sort_key_2_new` and add `sort_key_3`
- As a result, `WriteDropColumns` tries to remove `sort_key_2_new`, `WriteNewColumns` tries to add `sort_key_2_new` and `sort_key_3`
- There're always two obvious issues now: `sort_key_2_new` shouldn't appear in columns to drop and add

But why it doesn't surface during when transaction commits?
Checking how DuckLake drops column at transaction commit, it assumes all columns already committed, and we never check check affected rows
https://github.com/duckdb/ducklake/blob/96f8eb86926add08715b150888a413537b815f31/src/storage/ducklake_metadata_manager.cpp#L2252-L2259

**Followup questions**
Why does it surface at post-transaction query instead of commit?
After transaction commits, we don't check the consistency of data and schema, but did so at query.

This PR tries to solve the issue by recording in-transaction columns, so they're processed in-transaction without touching committed schemas.
This PR shamelessly steals the SQL statements attached in the test file, also add a comprehensive SQL tests for NULL/RENAME/DEFAULT operations in single transaction.

How do we handle multiple column rename operation in one transactions?
Only the last update is processed: https://github.com/duckdb/ducklake/blob/6c69da796c95976eefd6e86856a18769afd0c5ac/src/storage/ducklake_transaction.cpp#L1638-L1644

**Discussion**
To surface the bug earlier, I think one thing we could do is to check affected rows for each sql: https://github.com/duckdb/ducklake/discussions/995


test/sql/alter/multi_alter_same_column_transaction.test
```diff
@@ -0,0 +1,93 @@
...
+    (4, 'c', 'val4');
+
+statement ok
+BEGIN
+
+statement ok
+ALTER TABLE ducklake.expression_column_test ADD COLUMN rename_column_1 VARCHAR;
+
+statement ok
+ALTER TABLE ducklake.expression_column_test RENAME COLUMN rename_column_1 TO rename_column_2;
```
pdet, line 75: What if we Add + Rename +Drop?

pdet: Also, add two columns and remane both?

pdet (commented):
Hi @dentiny, thanks for the PR, just two comments on more testing! :-)

test/sql/alter/multi_alter_same_column_transaction.test
dentiny: That's a good point! Test cases added and fixes made.

pdet:
Thanks!

PR #1002, opened 2026-04-13
Remove stale comment for CTAS IF NOT EXISTS.

simi:
I just randomly found this comment in codebase while working on other patch. and got curious. 

It seems handled on DuckDB level already at 
https://github.com/duckdb/duckdb/blob/fc553552af581fbc84851da6f96fa315d4133729/src/catalog/catalog_entry/duck_schema_entry.cpp#L117-L122. I do offer to exchange the comment removal for unit test. Per my understanding, when table exists, DuckLake's `PlanCreateTableAs` is not reached at all.


pdet:
Thanks!

PR #1013, opened 2026-04-15
Fix 'executed' for snapshot expiration

dentiny:
`executed` is declared but never assigned to true, this PR fixes it.
It's hard to create a regression test since the side effect is to delete snapshots for multiple times which is not a functionality issue.


pdet:
Can we retarget it to `v1.5`?

dentiny:
> Can we retarget it to `v1.5`?

Done

Issue #1018, opened 2026-04-15
Unbounded memory growth when performing many small inserts into wide sparse tables


zhming0:
### What happens?

When performing many small INSERT batches into a wide table (e.g. 300 columns), DuckLake exhibits unbounded linear memory growth that ignores `SET memory_limit`. Each INSERT creates a new parquet file and associated metadata. The RSS growth is proportional to `num_inserts × num_columns`, suggesting per-file, per-column overhead held in unmanaged memory (outside DuckDB's buffer manager).

For a 300-column table with 20,000 insert batches of 100 rows each:
- RSS grows from ~95 MB to ~1,209 MB
- `memory_limit` set to 200 MB has no effect — no errors, no spilling
- Growth is perfectly linear (~57 MB per 2,000 batches) with no plateau

A narrow 10-column table with the same 20,000 batches only grows ~116 MB, confirming the column-count multiplier.

We have not confirmed which specific internal data structure is responsible for the RSS growth, but the strong correlation with `inserts × columns` points to something tracking per-file, per-column state in memory.

NOTE: I originally discover the memory leak symptom as part of a much larger system on JVM, I managed to narrow it down to insert and compaction, the compaction bit was raised by someone else in https://github.com/duckdb/ducklake/issues/929.

### To Reproduce

```python
import duckdb, os, resource, time, shutil

DB_PATH = "/tmp/ducklake_repro.duckdb"
DATA_DIR = "/tmp/ducklake_repro_data"
NUM_COLUMNS = 300
NUM_INSERTS = 20000
ROWS_PER_INSERT = 100
COLUMNS_PER_ROW = 5

for path in [DB_PATH, DATA_DIR]:
    if os.path.exists(path):
        shutil.rmtree(path) if os.path.isdir(path) else os.remove(path)

os.makedirs(DATA_DIR, exist_ok=True)

conn = duckdb.connect()
conn.execute("LOAD ducklake;")
conn.execute(f"ATTACH '{DB_PATH}' AS lake (TYPE DUCKLAKE, DATA_PATH '{DATA_DIR}');")
conn.execute("SET memory_limit='200MB';")
conn.execute("SET ducklake_default_data_inlining_row_limit=0;")

col_defs = ", ".join([f"col_{i} INTEGER" for i in range(NUM_COLUMNS)])
conn.execute(f"CREATE TABLE lake.wide_table ({col_defs});")

def get_mem_mb():
[...]

Issue #852, opened 2026-03-19
Memory leak with Catalog snapshots stored per schema version for connection lifetime


kposborne:
### What happens?

Ducklake stores the full catalog during `LoadSchemaForSnapshot` keyed by the `schema_version`, but doesn't evict from this map. For large catalogs this can consume a lot of memory if there are a lot of schema changes
https://github.com/duckdb/ducklake/blob/62503eff89876cc6403799e56e44c31fbe4b17ed/src/storage/ducklake_catalog.cpp#L195-L199

### To Reproduce

1. install uv: https://docs.astral.sh/uv/getting-started/installation/
2. Run reproducer script like: `uv run repro.py` - it first seeds a large catalog, then creates and drops tables repeatedly in 2 different ways, in one way it uses the same connection throughout, the other closes the connection periodically. You will see that when it keeps the same connection the memory usage continues to climb much greater than when it is reset
3. (optional) reproduce the profiles shared here by passing the `--memray` flag. This will be much slower but can identify the allocation locations. I didn't figure out how to get symbols working on MacOS but you can at least tell the primary culprit is this `LoadSchemaForSnapshot` function

repro.py:
```
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.12"
# dependencies = ["duckdb>=1.4.0", "memray"]
# ///
"""Minimal DuckLake schema-churn memory repro."""

from __future__ import annotations

import argparse
import csv
import ctypes
import ctypes.util
import gc
import json
import resource
import subprocess
import sys
import tempfile
import time
from contextlib import contextmanager
from pathlib import Path

import duckdb

CATALOG = "dl"
META = f"__ducklake_metadata_{CATALOG}"
MEMRAY_PROFILE_COUNT = 2


def say(line: str) -> None:
    print(line, flush=True)


def brief(row: dict[str, int | float | str]) -> str:
    return f"RSS {row['rss_mb']} MiB | snapshots {row['snapshot_count']} | schema_version {row['schema_version']}"


def rss_mb() -> float:
    if sys.platform == "darwin":
        libc = ctypes.CDLL(ctypes.util.find_library("System"))

        class TV(ctypes.Structure):
            _fields_ = [("s", ctypes.c_int), ("us", ctypes.c_int)]

        class Info(ctypes.Structure):
            _fields_ = [
                ("virtual_size", ctypes.c_uint64),
                ("resident_size", ctypes.c_uint64),
[...]

PR #1020, opened 2026-04-16
Fix unbounded memory (partially)

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1018

The issue when we do insertion and flush data files we store the stats in-memory and never evicts
https://github.com/duckdb/ducklake/blob/75223af99f0b1e1081fb7ca19f7faac0c9a03970/src/include/storage/ducklake_catalog.hpp#L214-L215

The workflow looks like (drawn by opus):
```sh
INSERT INTO ducklake.t VALUES (...)
  → auto-commit
    → FlushChanges()
      → GetNewDataFiles()
        → ducklake_catalog.GetTableStats(*this, table_id)
          → GetStatsForSnapshot(transaction, snapshot)
            → cache miss → LoadStatsForSnapshot → insert into cache
```

In this PR I store the table stats into object cache, which I used to implemented as a LRU cache and managed by buffer manager (similar to what I did for parquet metadata).

There's one caveat:
- There're another cache for schema: https://github.com/duckdb/ducklake/blob/75223af99f0b1e1081fb7ca19f7faac0c9a03970/src/include/storage/ducklake_catalog.hpp#L212-L213 I didn't add it into object cache because
  + In terms of memory consumption it's less of a problem compared with table stats
  + `LookupSchema` returns a raw pointer (instead of shared pointer): https://github.com/duckdb/ducklake/blob/75223af99f0b1e1081fb7ca19f7faac0c9a03970/src/include/storage/ducklake_catalog.hpp#L89-L90, which I don't have a good way to ensure memory safety


pdet:
Hi @dentiny, I think this is the right direction, and we probably have other unbounded growing maps that should be treated the same way. You identified `schemas`, `name_maps` is probably another. I have to do another pass to try to identify other unbound growths. I think we should fix all of them. (See https://github.com/duckdb/ducklake/issues/852 for someone running into the schema issue)

> I don't have a good way to add a good regression test, since when memory limit set low checkpoint will fail.. :(

I'm not sure what you mean here. In general, we should have long-running tests with a low enough memory limit to trigger eviction. What error are you seeing here?

dentiny:
> You identified schemas, name_maps is probably another. I have to do another pass to try to identify other unbound growths. I think we should fix all of them. (See https://github.com/duckdb/ducklake/issues/852 for someone running into the schema issue)

Yeah definitely agree.

For `schemes`, I need to check if I have to update DuckDB core, due to the invalid memory access issue (returning `optional_ptr` doesn't ensure cache entry still valid).

I think we could take a progressive way, that's why I name the PR as "partially". WDYT?

pdet:
If the test takes too long, it might be a good idea to add it as a `.test_slow`. And making sure we are not executing all over our CI (especially the one built on debug)

dentiny:
> If the test takes too long, it might be a good idea to add it as a .test_slow.

Thanks for the heads up! Sadly you're right -- I usually do reldebug build and it takes only 4 sec, good enough IMO; but debug build is another story.
Side topic, our debug build CI is surprisingly long (~2h), might be worth parallel test.

dentiny:
CI passes but I found a flaky test
```sh
1. test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54
Wrong result in query! (test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54)!
SELECT COUNT(*) FROM ducklake.t1 WHERE a < 5;
2 <> 0
Expected result:
0

Actual result:
2
```

pdet:
> CI passes but I found a flaky test
> 
> ```shell
> 1. test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54
> ================================================================================
> Wrong result in query! (test/sql/deletion_inlining/test_deletion_inlining_concurrent_retry.test:54)!
> ================================================================================
> SELECT COUNT(*) FROM ducklake.t1 WHERE a < 5;
> ================================================================================
> Mismatch on row 1, column count_star()(index 1)
> 2 <> 0
> ================================================================================
> Expected result:
> ================================================================================
> 0
> 
> ================================================================================
> Actual result:
> ================================================================================
> 2
> ```

Interesting, in which CI?

dentiny:
https://github.com/duckdb/ducklake/actions/runs/24509026602/job/71634898604
- The CI error is from my second last commit [31589a6](https://github.com/duckdb/ducklake/pull/1020/commits/31589a6556a9ec0f7c6aa87b567579d540995888)
- I tried to run multiple times but didn't get it reproduced
- I didn't do any (related) code change after the CI failure, nor rebase, so pretty confident it's a flaky test

test/sql/insert/insert_wide_table_low_memory.test_slow
```diff
@@ -0,0 +1,118 @@
+# name: test/sql/insert/insert_wide_table_low_memory.test
```
pdet, line 1: nit: `test_slow` but we can fix it in the next pr.

pdet:
Thanks!

test/sql/insert/insert_wide_table_low_memory.test_slow
dentiny: Ah yeah, I will followup with other object cache migrations.

PR #1028, opened 2026-04-17
Disable hive partitioning auto-detection in ParquetFileScanner

### LINKED ISSUE #1027: ScanDeleteFile fails when DuckLake data path contains key=value segments (hive auto-detect)
### What happens?

ParquetFileScanner binds parquet_scan without setting hive_partitioning, so DuckDB's default auto mode kicks in and synthesises an extra column for every key=value segment in the file path. DuckLakeDeleteFilter::ScanDeleteFile then sees 4 return types (file_path, pos, _ducklake_internal_snapshot_id, <hive col>) and throws:

```
Invalid Input Error: Invalid schema contained in the delete file
  <path>/ducklake-…-delete.parquet - expected file_name/position/[snapshot_id or row]
```
Pre-cc45d378 (2-column delete files) the same inflation silently produced a 3-column schema that passed the check and was used as if the hive column were a row/snapshot_id column.

### To Reproduce

```
-- mkdir -p '/tmp/ducklake-hive-repro/site_id=42'
ATTACH 'ducklake:/tmp/ducklake-hive-repro/site_id=42/lake.ducklake' AS lake;

CREATE TABLE lake.t AS SELECT range AS id FROM range(1000);

-- First DELETE produces a 2-column delete file (file_path, pos).
DELETE FROM lake.t WHERE id % 7 = 0;

-- Second DELETE produces a delete file with 3 columns on disk:
-- (file_path, pos, _ducklake_internal_snapshot_id).
-- ParquetFileScanner binds parquet_scan with default hive_partitioning='auto',
-- so the 'site_id=42' segment in DATA_PATH is parsed as a hive partition and
-- appended as a 4th synthetic column. DuckLakeDeleteFilter::ScanDeleteFile
-- then sees 4 columns and rejects the schema:
DELETE FROM lake.t WHERE id % 5 = 0;

-- Throws:
--   Invalid Input Error: Invalid schema contained in the delete file
--     <path>/...-delete.parquet - expected file_name/position/[snapshot_id or row]
SELECT count(*) FROM (SELECT * FROM lake.t);

-- Remove 'site_id=42' from the DATA_PATH and the error goes away.
```
Remove site_id=42 from the path and the error goes away.

Data file reads go through the same scanner and are also hive-inflated; currently masked by name-based projection but a latent issue if a real column name collides with a hive key.

### OS:

macOS 15.4, arm64

### DuckDB Version:

1.5.2

### DuckLake Version:

1.0

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Ilia Ablamonov

### Affiliation:

Rick.ai

- [x] Yes, I have

- [x] Yes, I have

Flamefork:
ParquetFileScanner binds parquet_scan without setting hive_partitioning, so DuckDB's default 'auto' mode synthesizes a column for every key=value segment along the file path. For ducklake-managed files those segments are incidental and the extra column breaks consumers that rely on an exact schema (DuckLakeDeleteFilter::ScanDeleteFile, ScanDataFileRowIds).

Fixes #1027.

### DIFF
--- a/src/common/parquet_file_scanner.cpp
+++ b/src/common/parquet_file_scanner.cpp
@@ -25,6 +25,9 @@ ParquetFileScanner::ParquetFileScanner(ClientContext &context, const DuckLakeFil
 	vector<LogicalType> input_types;
 	vector<string> input_names;
 
+	// ducklake-managed paths may contain incidental key=value segments
+	named_params["hive_partitioning"] = Value::BOOLEAN(false);
+
 	if (!file.encryption_key.empty()) {
 		child_list_t<Value> encryption_values;
 		encryption_values.emplace_back("footer_key_value", Value::BLOB_RAW(file.encryption_key));
--- a/test/sql/issues/issue_1027_hive_path_scan.test
+++ b/test/sql/issues/issue_1027_hive_path_scan.test
@@ -1,0 +1,30 @@
+# name: test/sql/issues/issue_1027_hive_path_scan.test
+# description: scanning a ducklake lake whose DATA_PATH contains key=value segments must not hive-auto-detect (#1027)
+# group: [issues]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+# DATA_PATH contains an incidental key=value segment: ParquetFileScanner
+# must not treat it as hive partitioning and append a synthetic column
+# to the delete-file schema.
+test-env DATA_PATH __TEST_DIR__/issue_1027/site_id=42
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}')
+
+statement ok
+CREATE TABLE ducklake.t AS SELECT i AS id FROM range(1000) t(i)
+
+statement ok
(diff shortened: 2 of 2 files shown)


src/common/parquet_file_scanner.cpp
```diff
@@ -25,6 +25,9 @@ ParquetFileScanner::ParquetFileScanner(ClientContext &context, const DuckLakeFil
 	vector<LogicalType> input_types;
 	vector<string> input_names;
 
+	// ducklake-managed paths may contain incidental key=value segments
+	named_params["hive_partitioning"] = Value::BOOLEAN(false);
```
pdet, line 29: The test added passes without this change

pdet (commented):
Hi @Flamefork, thanks for the PR.

It seems the test you added doesn't need your change to pass, could you confirm that?

src/common/parquet_file_scanner.cpp
Flamefork: Sorry, didn’t test this properly after debugging and fixing it in our production. It actually needs a second DELETE to fail: that one creates a 3-column delete file, and Hive then adds a fourth column from the path.

Fixed the test and pushed.

pdet:
Thanks! LGTM!

PR #1029, opened 2026-04-17
Embed end_snapshot in flushed Parquet files

qsliu2017:
Currently, `ducklake_flush_inlined_data` writes all inlined rows to Parquet, then re-queries the same inlined table to find deleted rows, computes their positions via `ROW_NUMBER()`, and writes separate position-based delete files. The end_snapshot information was already available during the initial scan -- the re-query and delete files exist only because the write path didn't carry end_snapshot through.

This PR adds `_ducklake_internal_end_snapshot_id` as a third virtual column in flushed Parquet files. Scans apply a visibility filter (IS NULL OR > current_snapshot) in FinalizeChunk.

Result for a flush that includes deletions: one data file instead of one data file + one delete file, no per-file re-query, simpler Finalize.

### DIFF
--- a/src/functions/ducklake_flush_inlined_data.cpp
+++ b/src/functions/ducklake_flush_inlined_data.cpp
@@ -31,21 +31,6 @@
 
 namespace duckdb {
 
-static void AttachDeleteFilesToWrittenFiles(vector<DuckLakeDeleteFile> &delete_files,
-                                            vector<DuckLakeDataFile> &written_files) {
-	unordered_map<string, reference<DuckLakeDataFile>> file_map;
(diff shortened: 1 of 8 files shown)


pdet:
Hi @qsliu2017 thanks for the PR. This is not the direction we want to go, I think the changes here are problematic for a few reasons:
1. It makes interop with iceberg/delta harder
2. It creates a double-source of truth
3. The optimization for the flushing is not that critical, as flushing is a maintenance op.
4. It's quite a big spec change.

For spec changes, in the future, is always better to communicate with the team before making the PR.

PR #1033, opened 2026-04-18
Fix unbounded memory consumption on schema cache entries

dentiny:
Closes https://github.com/duckdb/ducklake/issues/852

Followup PR for https://github.com/duckdb/ducklake/pull/1020
The idea to limit unbounded memory consumption on schema entries is the same: use the LRU object cache.
One thing different is: a few APIs returns optional pointer, which requires `DuckLakeCatalogSet` to be valid on return.

To workaround, I stored a shared pointer for catalog set in transaction context, so it's kept valid until transaction completes.

An alternative I've thought about is: update [`LookupSchema` API](https://github.com/duckdb/duckdb/blob/e64b98f66712b9674897bb27f7f417e5e88fb4be/src/include/duckdb/catalog/catalog.hpp#L221-L223) in DuckDB core to return shared pointer
But there're two big downsides:
- It basically changes all storage extensions (i.e., ducklake, iceberg, delta, etc)
- For all other storage extensions, I don't see any benefit for changing the interface

I checked locally on the regression test (with dummy logging), LRU cache entries goes get evicted during the test run.

Followup items (which I will do in next PRs):
- Check for other unbounded cache entries
- On detach, all existing cache entries should be evicted from object cache

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -232,11 +232,8 @@ public:
 
 	static unique_ptr<DuckLakeStats> ConstructStatsMap(vector<DuckLakeGlobalStatsInfo> &global_stats,
 	                                                   DuckLakeCatalogSet &schema);
-	//! Return the schema cache entry for the requested snapshot.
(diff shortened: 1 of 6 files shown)


pdet:
Thanks, LGTM!

PR #1034, opened 2026-04-18
fix: compaction loops

hello-world-bfree:
## Problems

At the moment running `MERGE_ADJACENT_TABLES` cycles through some files unnecessarily and not count single files against `max_files`. This fixes each issue, reducing some unnecessary work.

There was also an annoying build issue that hardcoded the include path for `roaring`. The change to `CMakeLists.txt` now works no matter where in the file system `roaring` is installed.


pdet:
Thanks, LGTM!

PR #1035, opened 2026-04-18
[fixes broken main build] Use FlatVector::SetNull for murmur3

txbm:
The DuckDB `ValidityMask` [change here](https://github.com/duckdb/duckdb/commit/585c52efc84215a274992f672fd5d2cdcf512616) breaks the DuckLake build on recent `main` builds. This fixes it.


pdet:
Thanks!

Issue #1036, opened 2026-04-19
BUG: transaction commit succeed with conflict table macros creation


dentiny:
### What happens?

See attached python repro script.

### To Reproduce

```py
import duckdb
db = duckdb.connect('test_macro.duckdb')
db.execute("INSTALL ducklake; LOAD ducklake")
db.execute("ATTACH '/tmp/test_macro.ducklake' AS dl (TYPE ducklake);")
db.execute("CREATE OR REPLACE TABLE dl.t1 (id INT)")
# Two cursors = two independent transactions on the same DB
c1 = db.cursor()
c2 = db.cursor()
# Conn1: create table macro, don't commit yet
c1.execute("BEGIN")
c1.execute("CREATE MACRO dl.my_macro(x) AS TABLE SELECT x AS col")
# Conn2: create same table macro and commit first
c2.execute("BEGIN")
c2.execute("CREATE MACRO dl.my_macro(x) AS TABLE SELECT x * 2 AS col")
c2.execute("COMMIT")
print("Conn2: committed OK")
# Conn1: commit — should conflict but may not
try:
    c1.execute("COMMIT")
    print("BUG: Conn1 also committed — duplicate table macro!")
except Exception as e:
    print(f"OK: Conn1 correctly rejected: {e}")
db.close()
```
With the latest release, both transaction commits.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

python

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1037, opened 2026-04-19
Add missing table macros conflict in transaction

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1036


pdet:
LGTM, can we add also a test that triggers the case? Also, what about scalar macros?

dentiny:
> can we add also a test that triggers the case? 

Ahh I forgot to push my latest commit...

> Also, what about scalar macros?

Scalar macro is checked one line below: https://github.com/duckdb/ducklake/blob/61d67a88c43b14d3cee2fdcc6b402cd7eec12a22/src/storage/ducklake_transaction.cpp#L1227

pdet:
Nice, I only forgot to add one of them 😂

PR #1039, opened 2026-04-20
Populate stats for more types

dentiny:
Closes https://github.com/duckdb/ducklake/discussions/1038

Currently we only populate stats (min/max/null) for a few types, so for uncovered types we have to go through one round trip of table scan, which involves IO, metadata query and relevant CPU work.

A few types are not included intentionally due to our parquet writer implementation:
- UHUGEINT doesn't emit stats: https://github.com/duckdb/duckdb/blob/e64b98f66712b9674897bb27f7f417e5e88fb4be/extension/parquet/include/writer/parquet_write_operators.hpp#L351-L353
- HUGEINT doesn't emit stats: https://github.com/duckdb/duckdb/blob/e64b98f66712b9674897bb27f7f417e5e88fb4be/extension/parquet/include/writer/parquet_write_operators.hpp#L335-L337
- TIME_TZ does emit stats, but with timezone information dropped: https://github.com/duckdb/duckdb/blob/e64b98f66712b9674897bb27f7f417e5e88fb4be/extension/parquet/include/writer/parquet_write_operators.hpp#L317-L322

### DIFF
--- a/src/storage/ducklake_stats.cpp
+++ b/src/storage/ducklake_stats.cpp
@@ -250,6 +250,8 @@ unique_ptr<BaseStatistics> DuckLakeColumnStats::CreateStringStats() const {
 
 unique_ptr<BaseStatistics> DuckLakeColumnStats::ToStats() const {
 	switch (type.id()) {
+	case LogicalTypeId::BOOLEAN:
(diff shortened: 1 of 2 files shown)


pdet:
Thanks!

Issue #512, opened 2025-10-13
Adding partition to unused table causes transaction conflicts


hlinander:
### What happens?

When trying to set partitioning on a newly created empty table the transaction fails with conflict when there are concurrent writers into other tables.
```
CREATE TABLE lake.test (...);
ALTER TABLE lake.test SET PARTITIONED BY (...);
```
```
TransactionContext Error: Failed to commit: Failed to commit DuckLake transaction.
Exceeded the maximum retry count of 10 set by the ducklake_max_retry_count setting.
. Consider increasing the value with: e.g., "SET ducklake_max_retry_count = 100;"
Failed to write new snapshot to DuckLake: Duplicate key "snapshot_id: 1122" violates primary key constraint.
```
See the provided test code for reproducing using Python threads.

### To Reproduce

https://gist.github.com/hlinander/3346b39a399f8989968dbef599d4fb1d

```
import duckdb
import pandas as pd
import threading
import time
import os
import shutil
import random
import string

NUM_WRITER_THREADS = 100
WRITE_DURATION_SECONDS = 5

stop_event = threading.Event()

def setup_environment():
    if os.path.exists("./ducklake_generic_test"):
        shutil.rmtree("./ducklake_generic_test")
    os.makedirs("./ducklake_generic_test")

    con = duckdb.connect()
    con.sql("INSTALL ducklake;")
    con.sql("LOAD ducklake;")
    con.sql(f"ATTACH 'ducklake:./ducklake_generic_test/metadata.ducklake' AS generic_lake (DATA_PATH './ducklake_generic_test');")
    con.sql(f"USE generic_lake;")

    print(f"Creating table (not partitioned yet)...")
    con.sql("""
        CREATE TABLE main_table (
            col1 INTEGER,
            col2 DECIMAL(10, 2),
            partition_key DATE
        );
    """)

    print("Creating separate writer table")
    con.sql("""
        CREATE TABLE writer_table (
            ts BIGINT,
            payload VARCHAR
        );
    """)
    con.close()

def writer_task(lake_connection: duckdb.DuckDBPyConnection, thread_id: int):
    cursor = lake_connection.cursor()
    print(f"[Writer {thread_id}] Started for table 'generic_lake.writer_table")
    while not stop_event.is_set():
        try:
            df = pd.DataFrame({
                'ts': [int(time.time() * 1000)],
                'payload': [''.join(random.choices(string.ascii_letters, k=10))]
            })
[...]

Issue #243, opened 2025-06-23
Concurrent writes can fail on first write to table


kyzyl:
I noticed a slightly obscure bug/behavior.  As of 0.2, doing a concurrent write to a freshly created catalog fails with:

```
ERROR:  duplicate key value violates unique constraint "ducklake_data_file_pkey"
DETAIL:  Key (data_file_id)=(0) already exists.
CONTEXT:  COPY ducklake_data_file, line 1
```
Note that this is a different error than you get for normal write conflicts ("failed to serialize"). The error seems to be indicating that the unique constraint on the file id is being violated, implying that the conflicting transactions are making it through (even though they shouldn't) but are hitting a further primary key conflict with the one transaction that did make it through. If there are any rows whatsoever in `ducklake_data_file` when the concurrent write occurs, it will still fail but with the regular transaction conflict error, not the PK error. 

This only occurs under the following conditions:
1. Do concurrent writes to an empty catalog (i.e. the state ducklake creates after initial ATTACH)
2. The writes must conflict and fail (e.g. `ducklake_max_retry_count` is set to 0)

So it's not a very serious bug, given that you can only see it once, when the catalog is empty, if the first write is concurrent, and that concurrent write fails with a conflict. So even if you had a conflicting, concurrent write as your first write, you still wouldn't see it unless you disabled retrying or the transactions conflicted for some other reason. 

Anyhow, maybe this is somehow expected behavior? I'm not sure how it works under the hood, but it seems a bit odd.

Mytherin:
Thanks for the report!

This seems like expected behavior to me. Data file ids are sequentially assigned, so concurrent writers can encounter primary key constraints there. The retry will then re-assign file identifiers with subsequent file ids.

kyzyl:
Yeah the mechanism makes sense. I guess the question is why do we only get a PK conflict on `data_file_id` when doing the first write to `ducklake_data_file`? Most concurrent writes will be attempting to use one of those sequentially assigned IDs, so why does the first batch of concurrent writes fail with the PK error, meanwhile all other conflicts from then on fail with a serialization error? Naively I would expect that if I do two identical concurrent writes, one with zero rows in `ducklake_data_file` and one with one row, the conflicts would fail with the same errors both times.

It just smelled like an edge case to me, but maybe not. Automated concurrency tests like mine are likely the only ones to see this anyhow.

PR #1044, opened 2026-04-21
fix: clashing ids, use race-free ids w/ postgres sequences

hello-world-bfree:
## Problem

Currently the snapshot, catalog, and file ids are created by reading the max id and adding 1 in C++ across two Postgres round trips. Concurrent writers often read the same max id, leading to both picking same plus-1 id, raising a uniqueness error. The retry loop eventually gets the data inserted, but the errors spam the logs, We're seeing 1MB+/hr. 

## Solution

Use Postgres `sequence`s and `nextval()`. It's atomic and race-free.

---

> [!NOTE]
> ***UPDATE***: Welp, I learned a lot really fast! I soon found out that the primary key violations are actually what helped keep files ordered and identifiers properly allocated. #243 and #512  were really helpful in understanding exactly what was going on. 

With postgres as the ducklake catalog, there are some unique conditions that the other catalogs do not experience as single writers, namely races. `snapshot_id`, `catalog_id`, `file_id`, and `schema_version` need unique values. Currently we find these with `max(id) ++` which gets complicated by multiple ducklake writers. These collisions/dupes can silently overwrite each other's stats or lead to stale caches that never self-correct.

To fix these uniquely postgres issues we use [sequences](https://www.postgresql.org/docs/17/sql-createsequence.html). We're guaranteed uniqueness even under heavy concurrent writes. This does lead to some edge cases to handle though.

Worker A can get assigned 23 in the sequence and worker B 24 shortly after, but worker B can finish and get committed first. That breaks a cache that assumes all previous ids in the sequence have been handled. To get around it, we add a short lock that's picked up by each writer immediately before id assignment and let go of upon commit. a 30 second timeout is provided to prevent any possibility of becoming stuck.

We also change up the cache key to be the `snapshot_id` (now always unique at any moment) rather than the `schema_version` since two different `snapshot_id`s can share a `schema_version`. The problem being they could still see two different tables despite the same `schema_version`. 

Before coordinated queueing, it was possible for one worker to overwrite the stats of another. A version number was added to the stats that'll be checked, triggering fresh numbers if not matched ensuring stats always converge towards accurate.

### DIFF
--- a/src/include/metadata_manager/postgres_metadata_manager.hpp
+++ b/src/include/metadata_manager/postgres_metadata_manager.hpp
@@ -34,11 +34,19 @@ public:
 
 	unique_ptr<QueryResult> Query(DuckLakeSnapshot snapshot, string &query) override;
 
+	idx_t AllocateNextSnapshotId(idx_t current_snapshot_id) override;
+	idx_t AllocateNextCatalogId(idx_t current_next_catalog_id) override;
+	idx_t AllocateNextFileId(idx_t current_next_file_id) override;
+
+	void EnsureIdSequences();
+
 protected:
 	string GetLatestSnapshotQuery() const override;
 
 private:
 	unique_ptr<QueryResult> ExecuteQuery(DuckLakeSnapshot snapshot, string &query, string command);
+
+	idx_t FetchScalarSequenceValue(const string &seq_name);
 };
 
 } // namespace duckdb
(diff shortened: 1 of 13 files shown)


pdet:
Hi, looks like there are many main -> v1.5 commits?

pdet:
Hi @hello-world-bfree, thanks for the PR.

I think your solution is no longer detecting the conflicts at all, as you're no longer running conflict resolution. I think the proper way of speeding up conflict resolution in postgres is by using stored procedures, thus eliminating round-trip costs, but that is a complex feature. I'm closing your PR for now.

PR #1045, opened 2026-04-21
Merge V1.5 -> Main

pdet:


src/storage/ducklake_merge_into.cpp
```diff
@@ -166,10 +166,20 @@ static void FinalizeCopyToInsert(Pipeline &pipeline, Event &event, ClientContext
...
                                                OperatorSinkFinalizeInput &input) const {
-	OperatorSinkFinalizeInput copy_finalize {*copy.sink_state, input.interrupt_state};
-	auto finalize_result = copy.Finalize(pipeline, event, context, copy_finalize);
-	if (finalize_result == SinkFinalizeType::BLOCKED) {
-		return SinkFinalizeType::BLOCKED;
+	auto &copy_to_file = copy.Cast<PhysicalCopyToFile>();
+	if (copy_to_file.partition_output) {
+		// Partitioned writes are normally driven by pipeline events scheduled from Finalize;
+		// that pattern doesn't fit here because we need the written files before we can sink into
+		// the metadata insert. Drive the partitioned finalize synchronously instead.
```
txbm, line 173: Just following along since you mentioned you're trying and sync with main which I have been trying to do as well -- I’m new to this codebase and just trying to learn so just asking questions out of curiousity.

I ran into this same ordering issue and thought it could stay [async by registering a flush callback](https://github.com/txbm/ducklake/blob/duckdb-main-alignment/.github/patches/extensions/ducklake/0007-partitioned-copy-flush-hook.patch) so copy→insert [runs after](https://github.com/txbm/ducklake/blob/duckdb-main-alignment/src/storage/ducklake_merge_into.cpp) the final `PartitionedCopyFinalizeEvent`, which preserves parallel flush.

Curious if you see a lifecycle/correctness issue with the callback approach, or if making finalize synchronous is required/preferred for some another reason I am totally missing?

src/storage/ducklake_multi_file_reader.cpp
```diff
@@ -678,6 +682,23 @@ void DuckLakeMultiFileReader::GatherDeletionScanSnapshots(BaseFileReader &reader
 	}
 }
 
+// Materialize a vector into a fresh flat vector of the given count, preserving values. Used to
+// normalize nested outputs from remap_struct, where dropping every source child of a struct inside
+// a list/map produces a Reference'd constant default whose underlying size does not match the
+// logical list length — downstream consumers (e.g. create_sort_key) then index past the buffer.
```
txbm, line 688: I tried to fix this by having DuckDB’s [remap_struct re‑set list child size after remap](https://github.com/txbm/ducklake/blob/duckdb-main-alignment/.github/patches/extensions/ducklake/0004-remap-list-child-size.patch) (to avoid per‑chunk materialization) - again just trying to learn: is copying the entire vector a tradeoff for something I am totally missing? Cases where the remap_struct fix doesn’t cover it?

txbm (commented):
Hi just following along since you referenced this PR in my attempts to stabilize `main` and I am trying to learn about this codebase

PR #1049, opened 2026-04-21
feat: Merge adjacent files across schema versions when only columns were added

### LINKED ISSUE #536: merge_adjacent_files not compacting a table with lots of single row files
### What happens?

We have a situation where some of our DuckLake tables in our application do not appear to get compacted when calling `ducklake_merge_adjacent_files`. 

One of the tables that we have been testing with contains about 2838 rows; looking through the `ducklake_data_files` table and filtering by the `table_id` of that table, we see that there are 2555 data files linked to the table, most of them containing only 1 row, with the notable exception of the initial file, which contined 157 rows. There are no entries in `ducklake_deleted_files` associated with the table, and none of the existing files contain any lightweight snapshot entries in the `partial_file_info` column. We also do not have a custom `target_file_size` set. Yet, for some reason the table just refuses to get compacted down, and as a result our queries running against this table end up experiencing significant slowdown. 

I have attempted to reproduce the issue but have not been able to so far. I have created tables with approximately the same number of rows as well as distribution amongst the files, but while I have sometimes seen a table compact only "partially" (in the sense that a handful of single-row files survive the compaction), I have not been able to reproduce the issue where no files at all are getting compacted. 

I have attached the rows of the `ducklake_data_files` table that pertain to the table that we investigated.

[ducklake_data_file_tb_13.csv](https://github.com/user-attachments/files/23241713/ducklake_data_file_tb_13.csv)

### To Reproduce

Unfortunately I do not have a fully reproducible example at this point, but I am still trying to get one and will update here if I succeed. 

### OS:

macOS 15.5

### DuckDB Version:

1.4.1

### DuckLake Version:

f134ad8

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Oliver Hsu

### Affiliation:

Ascend.io

- [ ] Yes, I have

- [x] Yes, I have

zhming0:
This solves #1048.
Fixes #536.

## Summary

`ducklake_merge_adjacent_files` now merges files across a `schema_version` boundary automatically when it is safe - no option required. A file written under an older schema is rewritten under the latest schema only when every field it carries still exists unchanged (same field id, name and type), i.e. the schema was unchanged or only had columns added (`ADD COLUMN`).

Fixes the limitation where `ADD COLUMN` left small files stranded on each side of a schema seam forever.

## Safety consideration

- Only column-add-only changes merge; `RENAME` / `DROP` / `ALTER TYPE` keep their files isolated, as before.
- Each file is checked against the latest schema (not pairwise), so this is robust across many schema versions.
- Time travel is preserved: old rows read `NULL`/default for added columns, and columns still can't be referenced before they existed.

## Tests

- `merge_adjacent_cross_schema.test` - ADD COLUMN merges, column defaults, time travel after cleanup, re-compaction safety.
- `merge_adjacent_cross_schema_isolation.test` - `RENAME` / `DROP` / `ALTER TYPE` and differing partition specs stay isolated.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -196,7 +196,9 @@ struct DuckLakeCompactionCandidates {
 };
 
 struct DuckLakeCompactionGroup {
-	idx_t schema_version;
+	//! Unset when cross-schema merging is allowed, so files from different
+	//! schema_versions fall into the same bucket.
+	optional_idx schema_version;
 	optional_idx partition_id;
 	vector<string> partition_values;
 };
@@ -204,7 +206,9 @@ struct DuckLakeCompactionGroup {
 struct DuckLakeCompactionGroupHash {
 	uint64_t operator()(const DuckLakeCompactionGroup &group) const {
 		uint64_t hash = 0;
-		hash ^= std::hash<idx_t>()(group.schema_version);
+		if (group.schema_version.IsValid()) {
+			hash ^= std::hash<idx_t>()(group.schema_version.GetIndex());
+		}
 		if (group.partition_id.IsValid()) {
 			hash ^= std::hash<idx_t>()(group.partition_id.GetIndex());
 		}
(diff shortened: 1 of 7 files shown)


pdet:
Hi @zhming0, thanks for the PR.

I believe that, indeed, `ADD COLUMN` is not a blocker for file merging, but we also should not gate that through an option, user's should not have to fiddle with that. 

Let's maybe focus on having only `ADD COLUMN` not impacting compaction, and add tests to show case that time travel works fine over it, and also that the other `ALTER` statements do not allow for cross schema compaction. And let's nuke the option.

zhming0:
@pdet No prob, I've reworked the PR. Please kindly give it another look 🙏🏿

dentiny:
cc @dentiny 
I will take a look recently

dentiny:
Hey @zhming0 CI seems to fail, do you mind taking a look? Thanks!

zhming0:
@dentiny the CI failure does not seem to be related to my changes. The same failure seems to happen in main branch too

zhming0:
@dentiny could you re-trigger the GHA workflow for me, please? Hopefully we can see a green build this time.

pdet:
Hi @zhming0, CI seems to be failing, could you have a look at that?

zhming0:
@pdet I fixed the failing test. The other failing test looks like a flaky test. Do you mind triggering the workflow again please?

pdet:
I think it looks good. One minor nit, can we reduce the comments  to one line (max two if absolutely necessary), and also only have them if they are necessary to explain something that is not clear from the code/test itself

pdet:
Thanks!

Issue #1055, opened 2026-04-22
BUG: column stats incorrect for columns with default value


dentiny:
### What happens?

See attached SQL, should be straightforward.

### To Reproduce

```sql
memory D  ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl (a INT);
memory D BEGIN;
memory D INSERT INTO dL.tbl VALUES (1), (2), (3);
memory D ALTER TABLE dl.tbl ADD COLUMN b INT DEFAULT 42;
memory D INSERT INTO dl.tbl VALUES (4, 200);
memory D COMMIT;
memory D SELECT * FROM __ducklake_metadata_dl.ducklake_table_column_stats;
┌──────────┬───────────┬───────────────┬──────────────┬───────────┬───────────┬─────────────┐
│ table_id │ column_id │ contains_null │ contains_nan │ min_value │ max_value │ extra_stats │
│  int64   │   int64   │    boolean    │   boolean    │  varchar  │  varchar  │   varchar   │
├──────────┼───────────┼───────────────┼──────────────┼───────────┼───────────┼─────────────┤
│        1 │         1 │ false         │ NULL         │ 1         │ 4         │ NULL        │
│        1 │         2 │ false         │ NULL         │ NULL      │ NULL      │ NULL        │
└──────────┴───────────┴───────────────┴──────────────┴───────────┴───────────┴─────────────┘
```
The second column doesn't have min and max value assigned properly.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1056, opened 2026-04-22
Fix ducklake stats for columns with default

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1055

The debugging process is straightforward, we just need to trace down the function call chain for adding a column.

Drawn by opus 4.6
```
DuckLakeTableEntry::AlterTable(AddColumnInfo)
  └─ transaction.AddColumnToLocalInlinedData(...)
       └─ LocalTableChanges::AddColumnToLocalInlinedData
            ├─ rebuilds data collection with new column (default or NULL)
            └─ creates DuckLakeColumnStats for column b:
                 ├─ null_count = 0
                 ├─ has_null_count = true 
                 ├─ any_valid = true
                 ├─ has_min = false        ← BUG: should be true
                 └─ has_max = false        ← BUG: should be true
```

### DIFF
--- a/test/sql/alter/add_column_default_stats.test
+++ b/test/sql/alter/add_column_default_stats.test
@@ -1,0 +1,103 @@
+# name: test/sql/alter/add_column_default_stats.test
+# description: verify min/max/null stats are correct after ADD COLUMN with DEFAULT on inlined data
+# group: [default]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/add_column_default_stats_files', METADATA_CATALOG 'ducklake_meta', DATA_INLINING_ROW_LIMIT 100)
+
+# ============================================
+# Case 1: INSERT + ADD COLUMN with DEFAULT inside explicit transaction
+# The added column's min/max should reflect the default value.
+# ============================================
+
+statement ok
+CREATE TABLE ducklake.t1 (a INT)
+
+statement ok
+BEGIN
+
+statement ok
+INSERT INTO ducklake.t1 VALUES (1), (2), (3)
+
+statement ok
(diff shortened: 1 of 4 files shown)


test/sql/alter/add_column_default_stats.test
```diff
@@ -0,0 +1,103 @@
...
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/add_column_default_stats_files', METADATA_CATALOG 'ducklake_meta', DATA_INLINING_ROW_LIMIT 100)
+
+# ============================================
```
pdet, line 16: can we generalize this test for multiple types?  maybe it's possible to get coverage with `FROM test_all_types()`?

dentiny: Sure, I add a new sql test with all column types.
Meanwhile, I think the current unit test is very clear on what it's trying to verify, so keep it here.

pdet:
Thanks, I asked the clanker to try to break it, he found the following issue:

```sql
statement ok
CREATE TABLE ducklake.t_del (a INT)

statement ok
BEGIN

statement ok
INSERT INTO ducklake.t_del VALUES (1), (2), (3)

statement ok
DELETE FROM ducklake.t_del

statement ok
ALTER TABLE ducklake.t_del ADD COLUMN b INT DEFAULT 999

statement ok
INSERT INTO ducklake.t_del VALUES (10, 5)

statement ok
COMMIT

query II
SELECT * FROM ducklake.t_del ORDER BY a
----
10	5

# Only 5 is present in column b — min and max must both be 5.
query II
SELECT s.min_value, s.max_value
FROM ducklake_meta.ducklake_table_column_stats s
JOIN ducklake_meta.ducklake_table t USING (table_id)
JOIN ducklake_meta.ducklake_column c USING (table_id, column_id)
WHERE t.table_name = 't_del' AND t.end_snapshot IS NULL
  AND c.end_snapshot IS NULL AND c.column_name = 'b'
----
5	5
```

dentiny:
> Thanks, I asked the clanker to try to break it, he found the following issue

Cool! That's indeed an issue, fixed

pdet:
Thanks!

PR #1061, opened 2026-04-22
Add query result helpers and guard call sites

txbm:
Re-opening from https://github.com/duckdb/ducklake/pull/996 - had to rename fork branches

This is now passing all CI checks

### DIFF
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -12,9 +12,12 @@
 #include "common/ducklake_data_file.hpp"
 #include "common/ducklake_snapshot.hpp"
 #include "duckdb/common/case_insensitive_map.hpp"
+#include "duckdb/common/exception.hpp"
 #include "duckdb/common/types/value_map.hpp"
 #include "duckdb/main/connection.hpp"
+#include "duckdb/main/query_result.hpp"
 #include "duckdb/transaction/transaction.hpp"
+#include <utility>
 #include "storage/ducklake_catalog_set.hpp"
 #include "storage/ducklake_inlined_data.hpp"
 #include "storage/ducklake_metadata_manager.hpp"
@@ -40,6 +43,36 @@ struct DuckLakePath;
 struct DuckLakeCommitState;
 class DuckLakeFieldId;
 class LocalTableChangeIterationHelper;
+
+inline unique_ptr<QueryResult> result_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	if (result->HasError()) {
+		result->GetErrorObject().Throw(err);
+	}
+	return result;
+}
+
+inline void execute_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	(void)result_or_throw(std::move(result), err);
+}
+
+template <class T>
+T query_scalar_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	auto checked_result = result_or_throw(std::move(result), err);
+	auto chunk = checked_result->Fetch();
+	if (!chunk || chunk->size() == 0) {
+		throw InvalidInputException("Expected scalar result");
+	}
+	return chunk->GetValue(0, 0).template GetValue<T>();
+}
+
+template <class T>
+T query_scalar_or_zero(unique_ptr<QueryResult> result, const string &err) {
+	auto checked_result = result_or_throw(std::move(result), err);
+	for (auto &row : *checked_result) {
+		return row.template GetValue<T>(0);
+	}
+	return T {};
+}
 
 struct FlushedInlinedTableInfo {
 	DuckLakeInlinedTableInfo inlined_table;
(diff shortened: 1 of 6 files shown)


txbm:
Hey @pdet - confirming this is rebased and still passes all checks. Also I did make all the edits you previously requested. Please let me know if there is anything else I can do to improve for you!

src/include/storage/ducklake_transaction.hpp
```diff
@@ -12,9 +12,12 @@
 #include "common/ducklake_data_file.hpp"
 #include "common/ducklake_snapshot.hpp"
 #include "duckdb/common/case_insensitive_map.hpp"
+#include "duckdb/common/exception.hpp"
 #include "duckdb/common/types/value_map.hpp"
 #include "duckdb/main/connection.hpp"
+#include "duckdb/main/query_result.hpp"
 #include "duckdb/transaction/transaction.hpp"
+#include <utility>
```
pdet, line 20: Do we need this?

src/storage/ducklake_initializer.cpp
```diff
@@ -61,15 +61,13 @@ string DuckLakeInitializer::GetAttachOptions() {
...
-		                catalog.MetadataPath() + "\"");
-	}
+	execute_or_throw(
+	    transaction.Query("ATTACH OR REPLACE {METADATA_PATH} AS {METADATA_CATALOG_NAME_IDENTIFIER}" +
+	                      GetAttachOptions()),
+	    "Failed to attach DuckLake MetaData \"" + catalog.MetadataDatabaseName() + "\" at path + \"" +
+	        catalog.MetadataPath() + "\"");
 	// explicitly load all secrets - work-around to secret initialization bug
-	transaction.Query("FROM duckdb_secrets()");
+	execute_or_throw(transaction.Query("FROM duckdb_secrets()"), "Failed to load DuckDB secrets");
```
pdet, line 70: I'm not sure if that's what we want to do, as errors being discarded before would allow for initialization still

src/include/storage/ducklake_transaction.hpp
```diff
@@ -41,6 +44,36 @@ struct DuckLakeCommitState;
...
+}
+
+template <class T>
+T query_scalar_or_zero(unique_ptr<QueryResult> result, const string &err) {
+	auto checked_result = result_or_throw(std::move(result), err);
+	for (auto &row : *checked_result) {
+		return row.template GetValue<T>(0);
+	}
+	return T {};
+}
```
pdet, line 75: They don't follow name convention for methods

src/include/storage/ducklake_transaction.hpp
```diff
@@ -41,6 +44,36 @@ struct DuckLakeCommitState;
 class DuckLakeFieldId;
 class LocalTableChangeIterationHelper;
 
+inline unique_ptr<QueryResult> result_or_throw(unique_ptr<QueryResult> result, const string &err) {
```
pdet, line 47: For these functions should we do  `const char *err` instead and only do the string heap allocation if we have to throw?

src/include/storage/ducklake_transaction.hpp
```diff
@@ -41,6 +44,36 @@ struct DuckLakeCommitState;
...
+
+template <class T>
+T query_scalar_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	auto checked_result = result_or_throw(std::move(result), err);
+	auto chunk = checked_result->Fetch();
+	if (!chunk || chunk->size() == 0) {
+		throw InvalidInputException("Expected scalar result");
+	}
+	return chunk->GetValue(0, 0).template GetValue<T>();
+}
```
pdet, line 66: nit: 

```cpp
        auto checked_result = result_or_throw(std::move(result), err);
        for (auto &row : *checked_result) {
                return row.template GetValue<T>(0);
        }
        throw InvalidInputException("Expected scalar result");
```

src/include/storage/ducklake_transaction.hpp
```diff
@@ -41,6 +44,36 @@ struct DuckLakeCommitState;
...
 
+inline unique_ptr<QueryResult> result_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	if (result->HasError()) {
+		result->GetErrorObject().Throw(err);
+	}
+	return result;
+}
+
+inline void execute_or_throw(unique_ptr<QueryResult> result, const string &err) {
+	(void)result_or_throw(std::move(result), err);
```
pdet, line 55: We need the cast?

pdet: actually do we even need this function? we can just ignore the result of the og function at call no?

src/include/storage/ducklake_transaction.hpp
pdet: maybe we can also drop them to one function? e.g.,

```cpp
  template <class T>
  T query_scalar(unique_ptr<QueryResult> result, const string &err, T default_value = T {}) {
        auto checked_result = result_or_throw(std::move(result), err);
        for (auto &row : *checked_result) {
                return row.template GetValue<T>(0);
        }
        return default_value;
  }
```

pdet (commented):
Thanks for the PR, I have a few comments

txbm:
Thank you so much for the comments! Will address ASAP

pdet:

PR #1063, opened 2026-04-23
Throw exception for invalid snapshot expiration request

dentiny:
Closes https://github.com/duckdb/ducklake/discussions/1062

Hi team, I feel confused that no error message or exception thrown when I expire snapshot with invalid params, it just silently passes.
In this PR, I directly throw InvalidInputException when invalid requests detected. Let me know if you think it makes sense!


pdet:
Thanks!

Issue #1068, opened 2026-04-25
BUG: rename view leads to tag loss


dentiny:
### What happens?

See attached SQL statements for repro, should be straightforward.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D USE dl;
dl D CREATE TABLE t(i INT);
dl D CREATE VIEW v AS SELECT * FROM t;
dl D COMMENT ON VIEW v IS 'view comment';
dl D SELECT view_name, comment from duckdb_views() WHERE view_name = 'v';
┌───────────┬──────────────┐
│ view_name │   comment    │
│  varchar  │   varchar    │
├───────────┼──────────────┤
│ v         │ view comment │
└───────────┴──────────────┘
dl D ALTER VIEW v RENAME TO v2;
dl D SELECT view_name, comment from duckdb_views() WHERE view_name = 'v2';
┌───────────┬─────────┐
│ view_name │ comment │
│  varchar  │ varchar │
├───────────┼─────────┤
│ v2        │ NULL    │
└───────────┴─────────┘
```
^ the `comment` column should return `view comment` instead of NULL.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1069, opened 2026-04-25
Fix view rename but comment dropped

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1068

The issue is
- When we commit a transaction for a renamed view, it's added to `dropped_views` directly
https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_transaction.cpp#L3097-L3099
- As a comparison (which works fine for table), it's added to `renamed_tables` https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_transaction.cpp#L3063-L3065

The fix here is simple: just mimic what we do for tables to add a `renamed_views` field.

### DIFF
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -275,6 +275,9 @@ public:
 	const set<TableIndex> &GetRenamedTables() {
 		return renamed_tables;
 	}
+	const set<TableIndex> &GetRenamedViews() {
+		return renamed_views;
+	}
 	const case_insensitive_map_t<unique_ptr<DuckLakeCatalogSet>> &GetNewTables() {
 		return new_tables;
 	}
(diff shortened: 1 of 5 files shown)


src/include/storage/ducklake_transaction.hpp
```diff
@@ -275,6 +275,9 @@ class DuckLakeTransaction : public Transaction, public enable_shared_from_this<D
 	const set<TableIndex> &GetRenamedTables() {
 		return renamed_tables;
 	}
+	const set<TableIndex> &GetRenamedViews() {
+		return renamed_views;
+	}
```
pdet: Is this unused?

pdet (commented):
Thanks, just. a minor nit, otherwise LGTM!

src/include/storage/ducklake_transaction.hpp
dentiny: Good catch, removed; I have mis-understanding for the option
```sh
-Wunused-function
Warn whenever a static function is declared but not defined or a non-inline static function is unused. This warning is enabled by -Wall.
```
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-function

pdet:
Thanks!

Issue #1070, opened 2026-04-25
BUG: dropping a table/view after altering inside of a transaction doesn't work


dentiny:
### What happens?

See my attached SQL, should be straightforward.

### To Reproduce

Example-1:
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D USE dl;
dl D USE dl;
dl D CREATE TABLE t (i INT);
dl D INSERT INTO t VALUES (1), (2);
dl D BEGIN;
dl D COMMENT ON TABLE t IS 'foo';
dl D DROP TABLE t;
dl D COMMIT;
dl D SELECT count(*) FROM t;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
```

Example-2:
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D USE dl;
dl D CREATE TABLE t (i INT);
dl D INSERT INTO t VALUES (1), (2);
dl D BEGIN;
dl D ALTER TABLE t RENAME TO t2;
dl D DROP TABLE t2;
dl D COMMIT;
dl D SELECT count(*) FROM t;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1071, opened 2026-04-25
Fix drop table/view after change in txn

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1070

The issue is
- For [`DuckLakeTableEntry::IsTransactionLocal`](https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/include/storage/ducklake_table_entry.hpp#L49-L51), it's returning "whether the table was changed inside the transaction", instead of "whether the table is created brand-new in the transaction"
- The confusing part is, [`SchemaIndex`](https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/include/common/index.hpp#L52-L55) has a function with the same name, but indicates whether it's created inside of the transaction
- Now looking back to our current implementation, if a table was changed inside of a transaction, it won't be added to `dropped_tables` https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_transaction.cpp#L2900-L2913

The fix for this PR is simple:
- Correct handle both semantics
- Apply the same change to view as well

Followup items:
- We need better naming / semantics for `IsTransactionLocal`, but I would prefer and willing to to followup with another PR; this PR should be self-contained enough

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -2935,13 +2935,12 @@ void DuckLakeTransaction::DropSchema(DuckLakeSchemaEntry &schema) {
 
 void DuckLakeTransaction::DropTable(DuckLakeTableEntry &table) {
 	catalog_version = ducklake_catalog.GetNewUncommittedCatalogVersion();
+	auto table_id = table.GetTableId();
 	if (table.IsTransactionLocal()) {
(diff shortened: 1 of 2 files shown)


pdet:
Thanks! LGTM!

PR #1075, opened 2026-04-26
Fix deletion filter initialization with snapshot cleared

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1074

Debugging process:
- First, I confirm data file, deletion file, and inline deletion looks ok
- So the incorrect query result must come from how to filter deleted rows, namely `DuckLakeDeleteFilter`
- I add logging to `deleted_rows` and `snapshot_ids` and find the culprit: when we combine deletion files and inline deletions, [snapshot ids are cleared](https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_delete_filter.cpp#L252), which means certain rows are always considered as deleted

### DIFF
--- a/src/storage/ducklake_delete_filter.cpp
+++ b/src/storage/ducklake_delete_filter.cpp
@@ -245,13 +245,44 @@ void DuckLakeDeleteFilter::Initialize(ClientContext &context, const DuckLakeFile
 
 void DuckLakeDeleteFilter::Initialize(const DuckLakeInlinedDataDeletes &inlined_deletes) {
 	D_ASSERT(std::is_sorted(delete_data->deleted_rows.begin(), delete_data->deleted_rows.end()));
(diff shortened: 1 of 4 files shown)


pdet:
Is the problem here that when we have a deletion file with embedded snapshot ids, the `ScanDeleteFile` is not properly executing the filter?

dentiny:
> Is the problem here that when we have a deletion file with embedded snapshot ids, the ScanDeleteFile is not properly executing the filter?

No. ScanDeleteFile is correct, the issue is we [clear the snapshot ids](https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_delete_filter.cpp#L252), making deleted rows undistinguishable when (in which version) are they deleted.

Issue #1080, opened 2026-04-28
BUG: build sort SQL doesn't consider quoted columns


dentiny:
### What happens?

See my attached example, should be straightforward.

### To Reproduce

```sql
# name: test/sql/sorted_table/flush_sorted_rename_quoted.test
# description: flush sorted inlined data with deletes when column name requires SQL quoting
# group: [sorted_table]

require ducklake

require parquet

test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db

test-env DATA_PATH __TEST_DIR__

statement ok
ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS dl (DATA_PATH '${DATA_PATH}/rename_with_quote', DATA_INLINING_ROW_LIMIT 100)

statement ok
CREATE TABLE dl.t("My Col" INTEGER)

statement ok
INSERT INTO dl.t VALUES (3), (1), (2)

statement ok
DELETE FROM dl.t WHERE "My Col" = 1

statement ok
ALTER TABLE dl.t SET SORTED BY ("My Col" DESC)

statement ok
CALL ducklake_flush_inlined_data('dl', table_name => 't')

query I
SELECT "My Col" FROM dl.t ORDER BY "My Col" DESC
----
3
2
```
The test fails for the current implementation, with error message
```sh
1. test/sql/sorted_table/flush_sorted_rename_quoted.test:31
================================================================================
Wrong result in query! (test/sql/sorted_table/flush_sorted_rename_quoted.test:31)!
================================================================================
SELECT "My Col" FROM dl.t ORDER BY "My Col" DESC;
================================================================================
Mismatch on row 2, column My Col(index 1)
1 <> 2
================================================================================
Expected result:
================================================================================
3
2

================================================================================
Actual result:
================================================================================
3
1
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1081, opened 2026-04-28
Fix sort SQL with quoted columns

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1080

I add debugging message on building sort SQL, and found the issue is [current implementation](https://github.com/duckdb/ducklake/blob/202c966dd0f6dc2241bbe657f6ee5cad25207c75/src/storage/ducklake_sort_data.cpp#L19) doesn't consider quoted columns.
- `current_columns.GetColumn(PhysicalIndex(i)).Name()` -> `My Col` (no quotes around)
- `field.expression` -> `"My Col"` (with quotes around)
- so column name comparison fails

### DIFF
--- a/src/storage/ducklake_sort_data.cpp
+++ b/src/storage/ducklake_sort_data.cpp
@@ -1,6 +1,8 @@
 #include "storage/ducklake_sort_data.hpp"
 
 #include "duckdb/parser/column_list.hpp"
+#include "duckdb/parser/parser.hpp"
+#include "duckdb/parser/expression/columnref_expression.hpp"
 #include "duckdb/common/string_util.hpp"
 #include "duckdb/common/exception_format_value.hpp"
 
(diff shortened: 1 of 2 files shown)


src/storage/ducklake_sort_data.cpp
```diff
@@ -16,7 +16,8 @@ string DuckLakeSort::BuildSortOrderSQL(const DuckLakeSort &sort_data, const Colu
 		// Check if expression matches a column name in the current table, then map to inlined table's name by index
 		string mapped_col;
 		for (idx_t i = 0; i < current_columns.PhysicalColumnCount(); i++) {
-			if (current_columns.GetColumn(PhysicalIndex(i)).Name() == field.expression) {
+			if (KeywordHelper::WriteOptionallyQuoted(current_columns.GetColumn(PhysicalIndex(i)).Name()) ==
```
pdet: Shouldn't we do `StringUtil::CIEquals`? If so, could we also add a test checking that.

dentiny: Sure -- the original code doesn't have case insensitivity check so I kept it

pdet: Prolly an oversight, I think this should work:

```sql
memory D CREATE TABLE t_2("MyCol" INTEGER);
memory D insert into t_2 values(10);
memory D select mycol from t_2;
```

PR #1083, opened 2026-04-28
Upgrade Ducklake to latest DuckDB Main

pdet:
Things to keep in mind:
1. Remove `mode skip` when https://github.com/duckdb/duckdb/pull/22346 gets in:
* test/sql/alter/struct_in_list_evolution.test
* test/sql/alter/struct_evolution_list_alter.test

2. Remove `SET pg_experimental_filter_pushdown=false;` from postgres.json, when postgres starts accepting filter expressions, as those can now be pushed down. (cc @staticlibs)


staticlibs:
@pdet 

Thanks for the heads up! Will look into the expression filters on Postgres side.

PR #1093, opened 2026-04-29
fix: retryable sqlite busy lock

txbm:
The SQLite busy lock is considered retryable as far as I can tell.

But the test suite fails intermittently because the current code does not treat it as such. Making it retryable causes the tests to reliably reach a consistent state via retry on busy locks

### DIFF
--- a/src/include/metadata_manager/sqlite_metadata_manager.hpp
+++ b/src/include/metadata_manager/sqlite_metadata_manager.hpp
@@ -27,6 +27,9 @@ public:
 	}
 
 	string GetColumnTypeInternal(const LogicalType &type) override;
+
+protected:
+	bool RetryOnMessage(const string &lower_message) const override;
 };
 
 } // namespace duckdb
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "duckdb/common/common.hpp"
+#include "duckdb/common/error_data.hpp"
 #include "duckdb/common/unordered_set.hpp"
 #include "duckdb/common/case_insensitive_map.hpp"
 #include "duckdb/common/optional_idx.hpp"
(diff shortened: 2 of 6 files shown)


pdet:
Thanks for the PR, I imagine this will allow us to run:

```   
 {
      "reason": "Failed to execute query COMMIT: database is locked",
      "paths": [
        "test/sql/compaction/compaction_delete_conflict.test",
        "test/sql/concurrent/concurrent_insert_conflict.test",
        "test/sql/concurrent/concurrent_insert_data_inlining.test",
        "test/sql/rewrite_data_files/test_rewrite_concurrency.test",
        "test/sql/rewrite_data_files/test_rewrite_transaction_conflict.test",
        "test/sql/settings/max_retry_count.test",
        "test/sql/snapshot_info/ducklake_current_commit.test",
        "test/sql/snapshot_info/ducklake_last_commit.test",
        "test/sql/stats/global_stats_transactions.test",
        "test/sql/transaction/concurrent_table_creation.test",
        "test/sql/transaction/transaction_conflict_cleanup.test",
        "test/sql/transaction/transaction_conflicts.test",
        "test/sql/transaction/transaction_conflicts_delete.test",
        "test/sql/transaction/transaction_conflicts_view.test",
        "test/sql/concurrent/file_level_conflict.test",
        "test/sql/transaction/transaction_conflict_inlining.test"
      ]
    }
```

If so, can we remove those from the test configuration?

Also, since this is catalog specific, maybe the catalog can have a Retry() function with specific rules for each catalog? I can imagine that becoming more frequent in the future.

txbm:
@pdet Such intuition! Indeed that allowed us to unskip all. What a win.

I've also generalized the `Retry` interface as you suggested. I think that's a great idea.

edit; I'm looking into the Linux CI failures. I think there might be some nuance required to fully align SQLlite retry semantics with the Ducklake transaction manager but I'll take a stab at figuring it out

pdet:
Hi @txbm, I think this looks like it's in the right direction, I think you might need to look into performing these retries in the snapshot-read / conflict-check path, then it should pass CI and be good to go.

txbm:
> Hi @txbm, I think this looks like it's in the right direction, I think you might need to look into performing these retries in the snapshot-read / conflict-check path, then it should pass CI and be good to go.

Oh thanks for the pointer! I was wondering that actually and just got bogged down testing it but that's encouraging I'll pick this back up and report back!

pdet:
Closing this due to lack of activity

Issue #525, opened 2025-10-20
`merge_adjacent_files` INTERNAL Error: DucklakeCompaction - expected a single output file


J-Meyers:
### What happens?

When calling ducklake_merge_adjacent_files on a single table (that was rather large and had quite a few files) it eventually crashed with this error:
```
----> 1 con.sql(f"CALL ducklake_merge_adjacent_files('ducklake', {name}, schema => '{schema_name}')")

InternalException: INTERNAL Error: DuckLakeCompaction - expected a single output file

Stack Trace:

/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb9Exception6ToJSONENS_13ExceptionTypeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x56) [0x7f67e6cf6666]
/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb9ExceptionC2ENS_13ExceptionTypeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x14) [0x7f67e6cf6714]
/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb17InternalExceptionC1ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x11) [0x7f67e6cf9811]
/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(+0xbc674b) [0x7f67e66de74b]
/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb18PipelineFinishTask11ExecuteTaskENS_17TaskExecutionModeE+0x22c) [0x7f67e70b85ac]
/home/shopuser/environment/post_processing/lib/python3.11/site-packages/_duckdb.cpython-311-x86_64-linux-gnu.so(_ZN6duckdb12ExecutorTask7ExecuteENS_17TaskExecutionModeE+0xd6) [0x7f67e70a8bd6]
[...]

PR #1095, opened 2026-04-29
Fix merge_adjacent_files with empty source files

Flamefork:
Allow compaction to finish without an output file when the source files have zero live rows. This lets merge_adjacent_files compact zero-row parquet files without tripping the single-output-file assertion while still rejecting missing output for non-empty input.

Fixes https://github.com/duckdb/ducklake/issues/525


pdet:
Thanks! LGTM!

PR #1099, opened 2026-04-30
Fix glob leak - DuckDB Main CI

pdet:


pdet:
@staticlibs I think that duckdb-main is broken on the postgres scanner tests, I'm disabling the postgres tests for now on DuckLake.

staticlibs:
@pdet 

Thanks for heads up! I have not touched the `main` branch with Postgres yet, there are a few things pending to get into 1.5.3, going to look at the `main` once 1.5.3 is out. I filed duckdb/duckdb-postgres#460 to track these tests.

Issue #1098, opened 2026-04-30
BUG: macro creation doesn't handle special character (single quote)


dentiny:
### What happens?

See my attached repro script, should be straightforward to understand.

### To Reproduce

```sql
-- Works for DuckDB
memory D CREATE MACRO "it's a macro"() AS 42; 
memory D SELECT "it's a macro"();
┌──────────────────┐
│ "it's a macro"() │
│      int32       │
├──────────────────┤
│               42 │
└──────────────────┘

-- Fails for DuckLake
memory D ATTACH 'ducklake:demo.db' AS dl (DATA_PATH 'demo_data', DATA_INLINING_ROW_LIMIT 100);
memory D USE dl;
dl D CREATE MACRO "it's a macro"() AS 42; 
TransactionContext Error:
Failed to commit: Failed to commit DuckLake transaction.
Failed to flush changes into DuckLake: syntax error at or near "s"

LINE 2: ... "__ducklake_metadata_dl"."main".ducklake_macro values(0,1,'it's a macro',1, NULL);
                                                                          ^
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1100, opened 2026-04-30
Fix quote processing for macros

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1098

The issue with current implementation is, take macro creation as an example, it takes the macro name and simply prepending and appending a single quote.
In detail, `it's_a_macro` becomes `'it''s_a_macro'` which is unparsable.

The fix is easy -- DuckDB already has util function `SQLString` to handle all cases.
I used opus 4.7 to scan the codebase and replace all self-implemented quote-handling to the util function.


pdet:
Thanks!

PR #1103, opened 2026-05-01
CREATE and CTAS SORTED BY and PARTITIONED BY

Alex-Monahan:
This PR adds support for the syntax from https://github.com/duckdb/duckdb/pull/20431 and https://github.com/duckdb/duckdb/pull/20728 :
```sql
CREATE TABLE woot 
  PARTITIONED BY (i) 
  SORTED BY (j)
  AS 
  FROM RANGE(9001) t1(i) 
  SELECT i, i+1 as j 
  ;
```
plus the non CTAS `CREATE TABLE` equivalent.

One open issue is that the syntax added to DuckDB in the above PRs does not support adding sort modifiers (`ASC` / `DESC`) to the sort, so I'll open up a bug for that here shortly if there isn't one already (added a FIXME in the code once that is unblocked). 

The key value of this PR is to make dbt work smoother with DuckLake. By default, everything in dbt is a CTAS. So this allows partitioning and sorting in a much smoother way in dbt (without a workaround). 

This was built with Claude, but I did make a number of adjustments to its initial solution to try and refactor it (and I trimmed down the ridiculously long comments...). Let me know if there is anything you would like to see changed!

CC @nicku33

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -47,16 +47,15 @@ vector<OrderByNode> DuckLakeCompactor::ParseSortOrders(const DuckLakeSort &sort_
 }
 
 //! Binds ORDER BY expressions directly using ExpressionBinder.
-vector<BoundOrderByNode> DuckLakeCompactor::BindSortOrders(Binder &binder, DuckLakeTableEntry &table,
-                                                           TableIndex table_index,
+vector<BoundOrderByNode> DuckLakeCompactor::BindSortOrders(Binder &binder, const ColumnList &columns,
+                                                           const string &table_name, TableIndex table_index,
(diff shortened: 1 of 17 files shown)


Alex-Monahan:
This is the related DuckDB PR to unblock `ASC`/`DESC` on the `SORTED BY` clause for `CREATE` and `CTAS`: https://github.com/duckdb/duckdb/issues/22411

Alex-Monahan:
Happy to target 1.5 if it would be helpful! [Edit - I made another PR for that in case it is helpful - see below!]

Alex-Monahan:
Marking this as draft - I think 1.5 is the better branch to target for now? (See https://github.com/duckdb/ducklake/pull/1107 )

pdet:
I don't think this is a spec change, so I'm closing it in favor of https://github.com/duckdb/ducklake/pull/1107

PR #1106, opened 2026-05-01
Drop view id from renamed

dentiny:
Followup PR for https://github.com/duckdb/ducklake/pull/1071 and https://github.com/duckdb/ducklake/pull/1069

Mimic what we do for tables
- When a view (which exists before transaction) gets dropped, we remove it from both dropped items and renamed items 
https://github.com/duckdb/ducklake/blob/90da74d5df014008d4ada27146c128a1c36beec6/src/storage/ducklake_transaction.cpp#L2956

I don't find it easy to reproduce with SQL tests, so as a regression test I added debug assertion.

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -2975,7 +2975,7 @@ void DuckLakeTransaction::DropView(DuckLakeViewEntry &view) {
 
 	// The view exists before the transaction, drop the view anyway.
 	if (!view_id.IsTransactionLocal()) {
-		// TODO(hjiang): sync with https://github.com/duckdb/ducklake/pull/1069 to remove renamed views.
+		renamed_views.erase(view_id);
 		dropped_views.insert(view_id);
 	}
 }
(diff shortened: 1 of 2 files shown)


dentiny:
The failed minio test doesn't seem related to my change.

src/storage/ducklake_transaction.cpp
```diff
@@ -2309,6 +2312,23 @@ string DuckLakeTransaction::CommitChanges(DuckLakeCommitState &commit_state,
...
+		D_ASSERT(overlap.empty());
+	}
+	{
+		set<TableIndex> overlap;
+		std::set_intersection(dropped_views.begin(), dropped_views.end(), renamed_views.begin(),
+		                      renamed_views.end(), std::inserter(overlap, overlap.begin()));
+		D_ASSERT(overlap.empty());
+	}
+#endif
+
```
pdet: Can't we assert this during the drop/rename?

dentiny: Yes, we could assert right after operations, but my concern is perf.
Our debug CI already takes >2h, I'm worried more frequently assertion makes wide table debug run worse :(

pdet: Ah, I think we should skip the debug CI and just have it run with a dispatch before release, or as a nightly, that CI rarely fails by itself.

dentiny: Moved the debug assertion around to when renamed/dropped happens.

PR #1107, opened 2026-05-01
CREATE and CTAS SORTED BY and PARTITIONED BY

Alex-Monahan:
I rebased https://github.com/duckdb/ducklake/pull/1103 onto DuckDB 1.5 in case that is better for the moment! Feel free to merge one or the other, or let me know what I can change!

This PR adds support for the syntax from https://github.com/duckdb/duckdb/pull/20431 and https://github.com/duckdb/duckdb/pull/20728 :
```sql
CREATE TABLE woot 
  PARTITIONED BY (i) 
  SORTED BY (j)
  AS 
  FROM RANGE(9001) t1(i) 
  SELECT i, i+1 as j 
  ;
```
plus the non CTAS CREATE TABLE equivalent.

One open issue is that the syntax added to DuckDB in the above PRs does not support adding sort modifiers (ASC / DESC) to the sort, so I'll open up a bug for that here shortly if there isn't one already (added a FIXME in the code once that is unblocked).

The key value of this PR is to make dbt work smoother with DuckLake. By default, everything in dbt is a CTAS. So this allows partitioning and sorting in a much smoother way in dbt (without a workaround).

This was built with Claude, but I did make a number of adjustments to its initial solution to try and refactor it (and I trimmed down the ridiculously long comments...). Let me know if there is anything you would like to see changed!

### DIFF
--- a/src/storage/ducklake_insert.cpp
+++ b/src/storage/ducklake_insert.cpp
@@ -35,20 +35,20 @@
 namespace duckdb {
 
 DuckLakeInsert::DuckLakeInsert(PhysicalPlan &physical_plan, const vector<LogicalType> &types, DuckLakeTableEntry &table,
-                               optional_idx partition_id, string encryption_key_p)
+                               optional_idx partition_id_p, string encryption_key_p)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, types, 1), table(&table), schema(nullptr),
-      partition_id(partition_id), encryption_key(std::move(encryption_key_p)) {
+      partition_id(partition_id_p), encryption_key(std::move(encryption_key_p)) {
 }
 
 DuckLakeInsert::DuckLakeInsert(PhysicalPlan &physical_plan, const vector<LogicalType> &types,
                                SchemaCatalogEntry &schema, unique_ptr<BoundCreateTableInfo> info, string table_uuid_p,
                                string table_data_path_p, unique_ptr<DuckLakePartition> ctas_partition_data_p,
-                               unique_ptr<DuckLakeSort> ctas_sort_data_p, optional_idx partition_id,
+                               unique_ptr<DuckLakeSort> ctas_sort_data_p, optional_idx partition_id_p,
                                string encryption_key_p)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, types, 1), table(nullptr), schema(&schema),
       info(std::move(info)), table_uuid(std::move(table_uuid_p)), table_data_path(std::move(table_data_path_p)),
       ctas_partition_data(std::move(ctas_partition_data_p)), ctas_sort_data(std::move(ctas_sort_data_p)),
-      partition_id(partition_id), encryption_key(std::move(encryption_key_p)) {
+      partition_id(partition_id_p), encryption_key(std::move(encryption_key_p)) {
 }
 
 //===--------------------------------------------------------------------===//
@@ -866,7 +866,7 @@ PhysicalOperator &DuckLakeCatalog::PlanCreateTableAs(ClientContext &context, Phy
 
 	reference<PhysicalOperator> root = plan;
 
-	// CTAS has no table_id yet, so honor only the schema-level sort_on_insert (default true).
+	// No table_id yet, so the table-level override can't apply; schema/global/default still do.
 	bool sort_on_insert =
 	    GetConfigOption<string>("sort_on_insert", duck_schema.GetSchemaId(), TableIndex(), "true") == "true";
 	if (sort_data && sort_on_insert) {
--- a/src/storage/ducklake_insert.cpp
+++ b/src/storage/ducklake_insert.cpp
@@ -42,10 +42,13 @@ DuckLakeInsert::DuckLakeInsert(PhysicalPlan &physical_plan, const vector<Logical
 
 DuckLakeInsert::DuckLakeInsert(PhysicalPlan &physical_plan, const vector<LogicalType> &types,
                                SchemaCatalogEntry &schema, unique_ptr<BoundCreateTableInfo> info, string table_uuid_p,
-                               string table_data_path_p, string encryption_key_p)
+                               string table_data_path_p, unique_ptr<DuckLakePartition> ctas_partition_data_p,
+                               unique_ptr<DuckLakeSort> ctas_sort_data_p, optional_idx partition_id,
+                               string encryption_key_p)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, types, 1), table(nullptr), schema(&schema),
       info(std::move(info)), table_uuid(std::move(table_uuid_p)), table_data_path(std::move(table_data_path_p)),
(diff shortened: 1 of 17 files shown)


guillesd:
I think reviewing this PR is necessary before 1125 (all of these commits are also there). @dentiny will you have a look? i can help if needed. For starts @Alex-Monahan, I think it would be good to merge 1.5 since this has been up for a while!

Alex-Monahan:
> I think reviewing this PR is necessary before 1125 (all of these commits are also there). @dentiny will you have a look? i can help if needed. For starts @Alex-Monahan, I think it would be good to merge 1.5 since this has been up for a while!

I have merged with 1.5! I will do the same with the other PR #1125 as well.

pdet:

Alex-Monahan:
Howdy @pdet!  I finally retargeted! I'll do the follow up PR (the `WITH` options) next.

dentiny:
Hi Alex, I'm aware of the PR and sorry for the delay. I'm slowly review (and catching up) since it's a big one, thanks for the patience!

dentiny:
Trying to play with the CTAS with partition :thinking_face:
```sql
# name: test/sql/create_table_inline_partition_sort/create_table_inline_partition_evolution.test
# description: Partition specs referenced by CTAS data files survive same-transaction partition evolution
# group: [create_table_inline_partition_sort]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

# Force CTAS rows into Parquet so their CREATE-time partition_id is recorded in ducklake_data_file.
statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/inline_partition_evolution', METADATA_CATALOG 'ducklake_meta', DATA_INLINING_ROW_LIMIT 0)

statement ok
USE ducklake

# CTAS writes files using the initial (a) partition spec. The ALTER creates a new (b) spec in the
# same transaction. Commit must preserve both specs because existing files still reference the first,
# while the second becomes active for future writes.
statement ok
BEGIN

statement ok
CREATE TABLE partition_evolution PARTITIONED BY (a) AS
SELECT 1 AS a, 101 AS b

statement ok
ALTER TABLE partition_evolution SET PARTITIONED BY (b)

statement ok
COMMIT

# Every partition_id referenced by a committed data file must resolve to a persisted partition spec.
# CTAS must have produced a data file with a partition spec.
query I
SELECT count(*)
FROM ducklake_meta.ducklake_data_file
----
1

# Every partition_id referenced by a data file must resolve to a persisted partition spec.
query I
SELECT count(*)
FROM ducklake_meta.ducklake_data_file df
WHERE NOT EXISTS (
    SELECT 1
    FROM ducklake_meta.ducklake_partition_info pi
    WHERE pi.partition_id = df.partition_id
)
----
0
```
gets error message
```sh
1. test/sql/create_table_inline_partition_sort/create_table_inline_partition_evolution.test:45
Wrong result in query! (test/sql/create_table_inline_partition_sort/create_table_inline_partition_evolution.test:45)!
SELECT count(*)
FROM ducklake_meta.ducklake_data_file df
WHERE NOT EXISTS (
    SELECT 1
    FROM ducklake_meta.ducklake_partition_info pi
    WHERE pi.partition_id = df.partition_id
);
1 <> 0
Expected result:
0

Actual result:
1
```

Alex-Monahan:
Ah, I didn't try editing the partition in the same transaction. I will have a look there. Thank you!

Alex-Monahan:
From my Clauding so far, it seems like this is a pre-existing issue with a flow of:
1.  `BEGIN`
1. `SET PARTITIONED BY (a)`
1. `INSERT`
1. `SET PARTITIONED BY (b)`
1. `COMMIT`

I'm going to see if it is possible to squash both bugs!

dentiny:
> I'm going to see if it is possible to squash both bugs!

I would appreciate if we could make it a separate PR, this one's already a giant one :)

Alex-Monahan:
So, I have found that this was something pre-existing. I'll have a PR up with a way to fix that shortly. Would you be able to review the remainder of this PR in the meantime? Is this one good to merge with this known gap?

src/storage/ducklake_insert.cpp
```diff
@@ -851,25 +853,55 @@ PhysicalOperator &DuckLakeCatalog::PlanCreateTableAs(ClientContext &context, Phy
...
+
+	// No table_id yet, so the table-level override can't apply; schema/global/default still do.
+	bool sort_on_insert =
+	    GetConfigOption<string>("sort_on_insert", duck_schema.GetSchemaId(), TableIndex(), "true") == "true";
+	if (sort_data && sort_on_insert) {
+		auto sorted_plan =
+		    PlanInsertSortFromColumns(context, planner, root.get(), columns, create_info.table, sort_data.get());
+		if (sorted_plan) {
+			root = *sorted_plan;
+		}
```
pdet, line 877: Can we have a helper function from PlanInsert and reuse it here?
It seems we are also missing the path from "When sort_on_insert=false but inlining is enabled..." Is that on purpose? If not, can we also add tests for that?

test/sql/create_table_inline_partition_sort/create_table_inline_transactions.test
```diff
@@ -0,0 +1,488 @@
...
+statement ok
+ALTER TABLE chain_winner SET SORTED BY (b DESC)
+
+statement ok
+COMMIT
+
+statement ok
+SET VARIABLE chain_winner_id = (SELECT table_id FROM ducklake_meta.ducklake_table WHERE table_name = 'chain_winner')
+
+# exactly one partition_info row (the last ALTER's) and one sort_info row (the last ALTER's)
```
pdet, line 258: Can we also check that we have no data files pointing to an uncommited partition spec?

pdet (commented):
Thanks for the PR @Alex-Monahan, I just had two comments!

src/storage/ducklake_insert.cpp
Alex-Monahan: I'm having a look at creating a helper function!

On the "When sort_on_insert=false but inlining is enabled..." piece, if `sort_on_insert=false`, I am not sorting during CTAS. I am being a little bit liberal with what I call an insert... I'm considering a `CTAS` to also be an insert. The parameter is a bit more like the opposite of "no sorts except during compaction". It would be better named as "sort_on_insert_or_create". All of that to say, it was a deliberate choice to not do the sorting during the CTAS if `sort_on_insert=false`. It might not be a wise choice though - happy to change it if you would prefer!

I'm not sure exactly how inlining would tie into it, but I can add some testing there. I guess if the total CTAS is smaller than the inlining limit, it would get inlined.

test/sql/create_table_inline_partition_sort/create_table_inline_transactions.test
Alex-Monahan: Looks like there are some cases of that. Investigating further (It's taking a bit more time for me understand that piece...)

pdet:
Closing it in favor of https://github.com/duckdb/ducklake/pull/1440

Issue #1109, opened 2026-05-02
BUG: out-of-range bucket leads to insertion failure


dentiny:
### What happens?

See attached repro sql, the issue is when set partition, the partition bucket could exceed u32::max, but when we it fails when we insert any value.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS ducklake (TYPE ducklake);
memory D CREATE TABLE ducklake.tbl (c INT);
memory D ALTER TABLE ducklake.tbl SET PARTITIONED BY (bucket(2147483648), c);
Invalid Input Error:
Expected bucket(bucket_count, column), but got bucket(2147483648)
memory D ALTER TABLE ducklake.tbl SET PARTITIONED BY (bucket(2147483648, c));
memory D INSERT INTO ducklake.tbl VALUES (2);
INTERNAL Error:
Information loss on integer cast: value 2147483648 outside of target range [-2147483648, 2147483647]
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1110, opened 2026-05-02
Fix bucket out of range

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1109

The root cause is, on row ingestion we apply partition transformation, here we convert partition count to uint32, which leads to INTERNAL exception if our partition bucket exceeds that range: https://github.com/duckdb/ducklake/blob/654147764169a948a99b92d53b1823135ce07ff5/src/storage/ducklake_partition_data.cpp#L123


dentiny:
The failed minio test doesn't seem related to my change.

pdet:
Thanks!

Issue #1111, opened 2026-05-02
BUG: cannot insert row after dropping sorted columns


dentiny:
### What happens?

See attached SQL for reproduction.

I think the current behavior is confusing: users are not able to insert data normally; I guess the reason is when a table drops sorted columns, the sort metadata is still kept, so when new row comes in, DuckLake still tries to sort.

As a comparison, dropping a partitioned column fails, whose behavior is much more clear.
```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl (a INT, b INT);
memory D ALTER TABLE dl.tbl SET PARTITIONED BY (b);
memory D ALTER TABLE dl.tbl DROP COLUMN b;
Catalog Error:
Cannot drop column "b" - the table is partitioned by this column. Reset or change the partitioning on this table in order to drop this column
```

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl (a INT, b INT);
memory D ALTER TABLE dl.tbl SET SORTED BY (b ASC);
memory D INSERT INTO dl.tbl VALUES (1, 10), (2, 20);
memory D ALTER TABLE dl.tbl DROP COLUMN b;
memory D INSERT INTO dl.tbl VALUES (3);
Binder Error:
Columns in the SET SORTED BY statement were not found in the DuckLake table. Unmatched columns were: b
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1112, opened 2026-05-02
Disallow dropping sorted columns

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1111

I think the current behavior for dropping a sorted column is confusing:
- Dropping a sorted column is acceptable with no issue
- But inserting a new row afterwards fails

In this PR, I try to mimic the behavior for dropping a partitioned column to keep behavior consistent and reasonable: users can only drop a column when it's not partitioned or sorted, and print out prompt to drop / reset sort before dropping.


dentiny:
Failed CI seems unrelated

pdet:
LGTM! Thanks!

Issue #1094, opened 2026-04-29
concurrent commit correctness; PK collisions ->> 1MB+ logs/hr.


hello-world-bfree:
## Problem

Currently the snapshot, catalog, and file ids are created by reading the max id and adding 1 in C++ across two Postgres round trips. Concurrent writers often read the same max id, leading to both picking same plus-1 id, raising a uniqueness error. The retry loop eventually gets the data inserted, but the errors spam the logs. We're seeing 1MB+/hr. 

With postgres as the ducklake catalog, there are some unique conditions that the other catalogs do not experience as single writers, namely races. `snapshot_id`, `catalog_id`, `file_id`, and `schema_version` need unique values. Currently we find these with `max(id) ++` which gets complicated by multiple ducklake writers. These collisions/dupes can silently overwrite each other's stats or lead to stale caches that never self-correct.

### To Reproduce

Observe logs for heavy-write workloads; i.e., cdc

### OS:

na

### DuckDB Version:

v1.5.2

### DuckLake Version:

1.0

### DuckDB Client:

Python, CLI

### Full Name:

Brandon Freeman

### Affiliation:

Hallow

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have not tested with any build

### Did you include all relevant data sets for reproducing the issue?

No - Other reason (please specify in the issue body)

### Did you include all code required to reproduce the issue?

- [ ] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [ ] Yes, I have

kelf:
This is an issue for us.

jason-cardinal-vadar:
This also a big issue for us regarding adoption. We are specifically seeing this with our DBT workloads.

jzillmann:
We're hitting this in production on Google Cloud Run (multi-instance) backed by Cloud SQL PostgreSQL 16. Confirming the exact mechanism Brandon described, with one important addition: **the cascade extends well beyond log spam** — for our workload it produces request failures and container OOM-kills, blocking our blue-green deploys.

### Setup

- DuckDB: 1.5.0 (via `duckdb-go-bindings` v0.10500.0)
- DuckLake: installed via `INSTALL ducklake; LOAD ducklake;` from the above
- Catalog: PostgreSQL 16 (Cloud SQL)
- DuckLake retry: `SET GLOBAL ducklake_max_retry_count = 100`
[...]

PR #1113, opened 2026-05-02
fix: concurrent-commit correctness on the postgres metadata backend

hello-world-bfree:
> [!NOTE]
> This is follow-up to #1044 , addresses issue #1094. Original PR closed by @pdet citing concerns that conflict resolution was being bypassed. That feedback surfaced a window where a concurrent committer could land  changes between the check and the lock. This was due to running `CheckForConflicts` before `AcquireCommitLock`. This PR resolves that as well as adds definitive conflict-detection tests and better frames/explains the issue. It's not actually performance related (that'll require the stored proc implementation that @pdet mentioned) but correctness related. Given the level of effort and cost for the stored proc implementation (as I understand it at least) this c++ implementation would stop the bleeding and allow for incremental progress towards the stored proc implementation with minimal cost.

## Problem

`*_id` allocation uses `max(id) + 1` in C++ across two postgres round trips. Concurrent writers race, hit
 `ducklake_snapshot_pkey` violations, retry succeeds but logs fill at **>1 MB/h** per cluster, and correctness bugs are introduced.

Multiple silent-correctness failure modes on the postgres metadata backend under concurrent writers. PK-violation log spam (>1 MB/h per cluster) is the visible symptom; the underlying defects are worse:
- **Cross-action conflicts not detected.** First-attempt commits skip `CheckForConflicts` and rely on `ducklake_snapshot_pkey` collision as the sole conflict signal. INSERT × ALTER, INSERT × DROP, ALTER × ALTER, COMPACT × ALTER, ALTER-view × DROP-view all silently both-commit. Catalog ends up referencing dropped tables, columns added to non-existent tables, etc.                                
- **No active-uniqueness invariants.** Two concurrent `CREATE TABLE` (same name), two concurrent `ADD COLUMN` (same name), or two concurrent DELETEs on the same data file all both-commit. `ducklake_table` / `ducklake_column` / `ducklake_delete_file` end up with duplicate active rows; subsequent resolution is nondeterministic.
- **Stats overwrites silently.** Concurrent UPDATE on `ducklake_table_stats` is last-writer-wins with no version tracking. Planner reads stale cardinalities.
- **Schema cache mis-keyed.** `GetSchemaForSnapshot` was keyed on `schema_version`, but two snapshots can share `schema_version` and see different tables (filtered by `begin_snapshot`). `AT(VERSION => N)` can return wrong column sets.                              - **Default isolation hides committers.** postgres_scanner attaches at `REPEATABLE_READ`. Once a DuckLake tx issues any metadata read, its PG snapshot pins for the rest of the tx, so `CheckForConflicts` sees empty `changes_made` even when conflicts exist.
- **PK-collision retry storms.** Every snapshot-id collision is a hard PG error producing the observed log noise plus retry backoff under load.

## Change

Allocate ids with postgres sequences (`nextval`, `cache 1`). This is race-free so there are no PK violations to log and create unnecessary bloat. Sequences are bootstrapped on attach and `setval()`'d from current `max` so existing catalogs migrate without issue. Decoupling allocation from commit order requires:

- **`pg_advisory_xact_lock`** (30 s timeout) around snapshot allocation + commit. `SetCommittedSnapshotId` becomes monotonic-max.   - **Partial unique indexes** on active rows of `ducklake_schema(schema_name)`, `ducklake_table(schema_id, table_name)`, `ducklake_view(schema_id, view_name)`, `ducklake_delete_file(data_file_id)`. `WHERE end_snapshot IS NULL`. Replaces PK-violation as the conflict signal.
- **`CheckForConflicts` runs after the lock on every attempt.** Authoritative under the commit critical section. `can_retry` gated so logical conflicts don't retry; transient lock errors do.  
- **Stats CAS** via new `stats_version` column on `ducklake_table_stats` plus post-commit `SELECT` to catch silent zero-row UPDATEs.
- **Schema cache** keys on `snapshot_id` instead of `schema_version`.                                                          
- **`isolation_level 'read committed'`** for PG metadata (overridable). Applies for both `metadata_type == "postgres"` and `"postgres_scanner"` — the original change missed the latter, which is what `DBPathAndType::ExtractExtensionPrefix` actually returns for the `postgres:` prefix.
- **Conflict-check matrix completion.** Five `ConflictCheck` calls added covering `altered_tables ↔ tables_merge_adjacent`, `altered_tables ↔ tables_rewrite_delete`, `altered_views ↔ dropped_views` (both directions). Pre-existing matrix omissions, surfaced while building the test suite.

## Tests

`test/sql/concurrent/concurrent_pg_conflict_detection.test` (new); 17 deterministic scenarios + iso-level regression guard + smoke. Each scenario uses named connections + explicit `BEGIN`/`COMMIT` and asserts on the regex-matched error from the labelled `ConflictCheck` line, so passing means the path actually fired.

| Mechanism | Scenarios |
|-----------|-----------|
| Partial unique index | A (CREATE × CREATE), D (DELETE × DELETE same file) |
| `CheckForConflicts` post-lock | B (INSERT × ALTER), C (INSERT × DROP), F (ALTER × ALTER), G (DROP × ALTER), H (DROP × DROP), I (ALTER × DELETE), J (DROP × DELETE), L (ALTER view × ALTER view), M (DROP view × DROP view), N (COMPACT × DELETE), O (DELETE × COMPACT), P (COMPACT × COMPACT), Q (COMPACT × ALTER), R (ALTER × COMPACT), S (ALTER view × DROP view), T (DROP view × ALTER view) |         | Stats CAS post-verify | E (`stats_version` monotone advance) |
| Negative control | K (INSERT × INSERT — both commit, row count = 2) |
| Iso-level regression guard | reads `current_setting('transaction_isolation')` from internal metadata catalog, asserts `read committed` (catches reintroduction of the `postgres_scanner` branch typo class) |
| Smoke | 8-way `concurrentloop` CREATE same name converges to one active row |

Compact × INSERT remains by-design no-conflict, asserted by the existing `test/sql/compaction/compaction_delete_conflict.test:119-138`.

`[concurrent]` suite: 470 assertions / 8 files, all green. `[compaction]` suite green on modified paths (`compaction_delete_conflict.test`, 50 assertions).

### DIFF
--- a/src/include/metadata_manager/ducklake_metadata_manager_v1_1.hpp
+++ b/src/include/metadata_manager/ducklake_metadata_manager_v1_1.hpp
@@ -20,6 +20,9 @@ public:
 
 	string GetCreateTableStatements() override;
 	string GetVersionString() override;
+	bool SupportsStatsVersionCAS() const override {
+		return true;
+	}
 };
 
 } // namespace duckdb
--- a/src/include/metadata_manager/postgres_metadata_manager.hpp
+++ b/src/include/metadata_manager/postgres_metadata_manager.hpp
@@ -37,11 +37,26 @@ public:
 
 	unique_ptr<QueryResult> Query(DuckLakeSnapshot snapshot, string &query) override;
 
+	idx_t AllocateNextSnapshotId(idx_t current_snapshot_id) override;
+	idx_t AllocateNextCatalogId(idx_t current_next_catalog_id) override;
+	idx_t AllocateNextFileId(idx_t current_next_file_id) override;
+	idx_t AllocateNextSchemaVersion(idx_t current_schema_version) override;
+	void AcquireCommitLock() override;
+
+	void EnsureIdSequences();
+
 protected:
 	string GetLatestSnapshotQuery() const override;
 
 private:
 	unique_ptr<QueryResult> ExecuteQuery(DuckLakeSnapshot snapshot, string &query, string command);
+
+	idx_t FetchScalarSequenceValue(const string &seq_name);
+
+	// classid half is hashtext(schema) so multiple DuckLake catalogs on one
+	// pg instance do not share the key.
+	static constexpr int32_t DUCKLAKE_COMMIT_ADVISORY_SUBKEY = 0x44754C4B;
+	optional_idx commit_lock_classid;
 };
 
 } // namespace duckdb
--- a/src/include/metadata_manager/postgres_metadata_manager.hpp
+++ b/src/include/metadata_manager/postgres_metadata_manager.hpp
@@ -52,10 +52,14 @@ private:
 	unique_ptr<QueryResult> ExecuteQuery(DuckLakeSnapshot snapshot, string &query, string command);
 
 	idx_t FetchScalarSequenceValue(const string &seq_name);
+	idx_t EnsureCatalogClassid();
+	bool BootstrapObjectsPresent();
 
 	// classid half is hashtext(schema) so multiple DuckLake catalogs on one
 	// pg instance do not share the key.
 	static constexpr int32_t DUCKLAKE_COMMIT_ADVISORY_SUBKEY = 0x44754C4B;
+	// Distinct subkey for bootstrap lock so it does not serialise with commits.
+	static constexpr int32_t DUCKLAKE_BOOTSTRAP_ADVISORY_SUBKEY = 0x42535452;
 	optional_idx commit_lock_classid;
 };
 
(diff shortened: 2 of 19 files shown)


pdet:
Hi @hello-world-bfree, thanks for the PR and for putting in the work. I spent some time reviewing the PR  and i think your chances introduces a few correctness issues, including potential PK collisions during the file ID sequence bootstrap and a change in the retry classification logic that no longer matches DuckDB's MVCC conflict handling,.

Given the complexity of the problem and the diff size, I think it's easier if this were to be broken to a series of small PRs that I can then review more carefully.

PR #1115, opened 2026-05-03
fix: throw CatalogException for column names exceeding 63 bytes in Postgres inlined data

### LINKED ISSUE #1089: Postgres inlined tables with column names >=64 characters in length cause crash on flush
### What happens?

When inlining a table that has column names that are longer than 63 characters ([potentially 63 bytes](https://stackoverflow.com/questions/8213127/maximum-characters-in-labels-table-names-columns-etc)), the inline initially works seemingly without issues - but attempting to flush inlined data causes this error:
```
_duckdb.InternalException: INTERNAL Error: Attempted to access index 0 within vector of size 0
This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
```

I'm not sure if this is confined to Postgres flavour ducklakes - I did test it with a duckdb catalog and did not see an error.

I think probably there is an element of this that can't be fixed due to intrinsic postgres column name limit (see above link), but ideally duckdb would throw an error when inlining the data in the first place (or when creating/amending the table with the too long column name?), rather than failing on the flush.

### To Reproduce

```python 
ALIAS = "my_ducklake"
DATA_PATH = "catalog.ducklake.files"
DATA_INLINING_ROW_LIMIT = 4 # so 3 row table below will be inlined


def get_ducklake_connection() -> duckdb.DuckDBPyConnection:
    conn = duckdb.connect()
    conn.execute("INSTALL ducklake")
    conn.execute("LOAD ducklake")
    conn.execute(f"""
        CREATE OR REPLACE SECRET postgres_secret (
            TYPE postgres,
            HOST '{POSTGRES_HOST}',
            PORT {POSTGRES_PORT},
            DBNAME '{POSTGRES_DBNAME}',
            USER '{POSTGRES_USER}',
            PASSWORD '{POSTGRES_PASSWORD}'
        )
    """)
    conn.execute(f"""
        ATTACH '' AS {ALIAS} (
            TYPE ducklake,
            DATA_PATH '{DATA_PATH}',
            DATA_INLINING_ROW_LIMIT {DATA_INLINING_ROW_LIMIT},
            METADATA_PATH '',
            METADATA_PARAMETERS MAP {{'TYPE': 'postgres', 'SECRET': 'postgres_secret'}}
        )
    """)
    conn.execute(f"USE {ALIAS}")
    return conn


def flush_inlined_tables(conn: duckdb.DuckDBPyConnection) -> None:
    conn.execute(f"""
        CALL ducklake_flush_inlined_data('{ALIAS}')
    """)


csv_name = "test_csv.csv"
column_name = "_" * 64
csv_data = f"""{column_name}
1
2
3"""

with open(csv_name, "w") as f:
    f.write(csv_data)


conn = get_ducklake_connection()
print(f"creating table...")
table_name = (
    f"test_table_{uuid.uuid4().hex[:8]}"  # unique name so you can re-run the test
)

conn.execute(f"""
    CREATE TABLE {table_name} AS SELECT * FROM read_csv_auto('{csv_name}')
""")

print("Flushing inlined tables...")
flush_inlined_tables(conn)
```

### OS:

Windows 11 x64

### DuckDB Version:

1.5.2

### DuckLake Version:

415a9eb

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Adam Johnston

### Affiliation:

University of Edinburgh

Yes

- [x] Yes, I have

- [x] Yes, I have

EdanStasiuk:
## Summary

Fixes #1089. Postgres silently truncates identifiers longer than 63 bytes, which caused an internal assertion failure (`Attempted to access index 0 within vector of size 0`) when flushing inlined tables with column names >= 64 bytes. Instead of crashing at flush time, we now throw a clear `CatalogException` in the `DuckLakeTableEntry` constructor, which covers table creation, column add, and column rename.


EdanStasiuk:
I hardcoded the 63-byte limit in the `DuckLakeTableEntry` constructor. Ideally this would use `metadata_manager.MaxIdentifierLength()` dynamically (which already returns 63 for the Postgres backend via `PostgresMetadataManager`)

https://github.com/duckdb/ducklake/blob/654147764169a948a99b92d53b1823135ce07ff5/src/include/metadata_manager/postgres_metadata_manager.hpp#L28-L30

 but the constructor doesn't have access to a transaction or `ClientContext` to get the metadata manager. Happy to refactor if the maintainers prefer a dynamic approach.

pdet:
Hi @EdanStasiuk, could you move your solution to v1.5-andium? Indeed the size should not be hardcoded.

PR #1124, opened 2026-05-05
fix: use ANSI CAST instead of PostgreSQL ::boolean cast in stats update

rustyconover:
## Summary
- Replaces PostgreSQL-style `::boolean` casts with ANSI `CAST(... AS boolean)` in the column-stats `UPDATE` query in `ducklake_metadata_manager.cpp`.
- Improves portability of the metadata SQL across catalog backends that don't support the `::` cast syntax.

## Test plan
- [ ] Existing test suite passes against DuckDB/SQLite metadata catalogs
- [ ] Verify column stats updates still produce identical results


pdet:
Thanks

PR #1126, opened 2026-05-06
Add VgiMetadataManager for VGI / Durable Object backend

rustyconover:
@pdet, how do you want to start to handle pluggable metadata managers?  Right now this is the start of one but it doesn't seem like you've gotten many others. I'll probably have a few more changes soon, as this is just Durable Objects but really this could be any API.

Routes DuckLake metadata commits through the VGI worker's ducklake_do_execute(sql) function so the entire commit batch ships as one HTTP POST to the DO. Reads flow through the existing VGI catalog readers. Bootstrap is owned by the VGI ATTACH (data_path option) — InitializeDuckLake() throws if tables are missing.

### DIFF
--- a/src/include/metadata_manager/vgi_metadata_manager.hpp
+++ b/src/include/metadata_manager/vgi_metadata_manager.hpp
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// metadata_manager/vgi_metadata_manager.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "storage/ducklake_metadata_manager.hpp"
+
+namespace duckdb {
+
+class VgiMetadataManager : public DuckLakeMetadataManager {
+public:
+	explicit VgiMetadataManager(DuckLakeTransaction &transaction);
+
+	static unique_ptr<DuckLakeMetadataManager> Create(DuckLakeTransaction &transaction) {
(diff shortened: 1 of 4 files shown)


pdet:
Hi Rusty, thanks for the PR. I definetly don't want extra catalogs directly registered in the DuckLake extension. These should live as separate extensions. The way I'm envisioning this is to have a registration query in DuckLake that would allow these catalogs to be registered through a third-party extension, but that is not currently a priority for me.

For example, it could look something like:

```sql
 CALL ducklake_register_metadata_manager(
'vgi',
    '{
      "base": "sqlite",
      "types": {                                                                                                                                                                                                                                                                                                          
        "natively_unsupported": [
          "STRUCT","MAP","LIST","FLOAT","DOUBLE","TIMESTAMP_TZ","VARIANT"                                                                                                                                                                                                                                                 
        ],                                                                                                                                                                                                                                                                                                                
        "inlining_unsupported": ["VARIANT"],                                                                                                                                                                                                                                                                              
        "column_type_overrides": {                                                                                                                                                                                                                                                                                        
          "FLOAT": "VARCHAR", "DOUBLE": "VARCHAR", "VARIANT": "VARCHAR"
        }                                                                                                                                                                                                                                                                                                                 
      },
...
);
```

Then that could live in an extension that depends on DuckLake. I think that should be doable since the catalog changes should never be heavily intertwined with the cpp code. That is also something for the extension version that will come out with DuckDB 2.0, not 1.5.3.

Issue #1127, opened 2026-05-06
BUG: alter column type to wider type is not allowed


dentiny:
### What happens?

See attached SQL for example.

- `UINTEGER` type is of `uint32_t` physical type: https://github.com/duckdb/duckdb/blob/b3ed0d5191465b8e10fc1dc74fe1880e26e4ca45/src/include/duckdb/common/types/value.hpp#L401-L403
- `BIGINT` type of of `int64_t` physical type: https://github.com/duckdb/duckdb/blob/b3ed0d5191465b8e10fc1dc74fe1880e26e4ca45/src/include/duckdb/common/types/value.hpp#L385-L387
- so the conversion should be allowed

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.tbl(x UINTEGER);
memory D ALTER TABLE dl.tbl ALTER COLUMN x TYPE BIGINT;
Catalog Error:
Cannot change type of column x from UINTEGER to BIGINT - only widening type promotions are allowed
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1128, opened 2026-05-06
Fix type promotion for UINTEGER

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1127

According to the physical/logical type mapping
- `UINTEGER` type is of `uint32_t` physical type: https://github.com/duckdb/duckdb/blob/b3ed0d5191465b8e10fc1dc74fe1880e26e4ca45/src/include/duckdb/common/types/value.hpp#L401-L403
- `BIGINT` type of of `int64_t` physical type: https://github.com/duckdb/duckdb/blob/b3ed0d5191465b8e10fc1dc74fe1880e26e4ca45/src/include/duckdb/common/types/value.hpp#L385-L387
- So we `UINTEGER` columns should be able to cast to `BIGINT`

Update: use DuckDB type promotion check function to unlock all possible type conversions.

### DIFF
--- a/src/storage/ducklake_table_entry.cpp
+++ b/src/storage/ducklake_table_entry.cpp
@@ -850,7 +850,7 @@ bool TypePromotionIsAllowed(const LogicalType &source, const LogicalType &target
 	case LogicalTypeId::USMALLINT:
 		return TypePromotionIsAllowedUSmallint(target);
 	case LogicalTypeId::UINTEGER:
-		return target.id() == LogicalTypeId::UBIGINT;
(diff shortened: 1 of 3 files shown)


src/storage/ducklake_table_entry.cpp
pdet: I think we should probably fix this FIXME. This function seems to have a lot of gaps as is. Would you be willing to tackle this?

dentiny: sure! let me check it out

pdet:
Thanks @dentiny, looks great!

Issue #1129, opened 2026-05-06
BUG: failed to rename view with the same name


dentiny:
### What happens?

We cannot rename view with the same name.
```sql
memory D CREATE VIEW dl.MY_VIEW_1 AS SELECT 42;
memory D ALTER VIEW dl.MY_VIEW_1 RENAME TO my_view_1;
Catalog Error:
Could not rename view "MY_VIEW_1" to "my_view_1": another entry with this name already exists!
```

But the same sql works for DuckLake table rename.
```sql
memory D CREATE TABLE dl.MY_TABLE(i INT);
memory D ALTER TABLE dl.MY_TABLE RENAME TO my_table;
```

I also checked DuckDB behavior -- it allows table/view rename to the same name.
```sql
-- View rename works.
memory D CREATE VIEW MY_VIEW AS SELECT 42;
memory D ALTER VIEW MY_VIEW RENAME TO my_view;
-- Table rename works.
memory D CREATE TABLE MY_TABLE(i INT);
memory D ALTER TABLE MY_TABLE RENAME TO my_table;
```

### To Reproduce

repro sql already attached above.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1130, opened 2026-05-06
Fix view rename

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1129

Mimic what we do for table entry renames: https://github.com/duckdb/ducklake/blob/de16277a2eefd90077de2ccba3da7edd2aa565ec/src/storage/ducklake_schema_entry.cpp#L209-L216


pdet:
LGTM!

Issue #1131, opened 2026-05-06
BUG: failed to create table macro and scalar macro with the same name


dentiny:
### What happens?

See repro sql, should be straightforward.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D USE dl;
-- We're able to create scalar/table macro with the same name.
dl D CREATE MACRO foo(x) AS x + 1;
dl D CREATE MACRO foo(x) AS TABLE SELECT x AS a, y AS b;
-- But we're able to do so in the same transaction.
dl D BEGIN;
dl D CREATE MACRO bar(x) AS x + 1;
dl D CREATE MACRO bar(x) AS TABLE SELECT x AS a, y AS b;
Catalog Error:
Macro Function with name "bar" already exists!
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1132, opened 2026-05-06
Fix table/scalar macro with same name creation in same txn

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1131

The root cause of the issue is
- For committed entries, scalar and table macros are stored as separate data members: https://github.com/duckdb/ducklake/blob/de16277a2eefd90077de2ccba3da7edd2aa565ec/src/include/storage/ducklake_schema_entry.hpp#L77-L78
- While inside of a transaction, these two macros are stored in the same field: https://github.com/duckdb/ducklake/blob/de16277a2eefd90077de2ccba3da7edd2aa565ec/src/include/storage/ducklake_transaction.hpp#L354-L355

So the fix in this PR is fairly straightforward, just to split the new macro map into two parts and update all usages.


pdet:
Thanks!

Issue #1141, opened 2026-05-09
BUG: schema evolution leads to missing column


dentiny:
### What happens?

See my repro sql, should be clear.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D CREATE TABLE dl.test(a INT);
memory D BEGIN;
memory D ALTER TABLE dl.test ADD COLUMN b INT;
memory D ALTER TABLE dl.test RENAME COLUMN b TO c;
memory D ALTER TABLE dl.test ALTER c SET DATA TYPE BIGINT;
memory D COMMIT;
memory D DESCRIBE dl.test;
┌────────────────────────┐
│          test          │
│                        │
│ a integer default NULL │
└────────────────────────┘
-- column c is missing!
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1142, opened 2026-05-09
Fix missing column in schema evolution

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1141

This is what happened in the repro SQL, in the reverse-temporal pass
- ADD_COLUMN -> txn_added_fields[2] = 0, result.new_columns = [(b,2,INT)]
- RENAME -> in‑place update at result.new_columns[0] to (c,2,INT)
- CHANGE_COLUMN_TYPE → in HandleChangedFields: new_fields loop sees the new field with id 2, because 2 is already columns_handled_by_later_ops, it gets skipped; dropped_fields loop sees id 2 in txn_added_fields and cancels the ADD.

The root cause is
- For the current implementation, we use [`columns_handled_by_later_ops`](https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/storage/ducklake_transaction.cpp#L1590) to deduplicate drop + add column logic (because for these 4 involved operations, they're implemented as drop + add, only the last one wins)
- What makes things complicated is, it treats (1) already existing columns before transaction; (2) new columns created in the transaction differently
  + ex1: for in-transaction added columns, we directly update column info https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/storage/ducklake_transaction.cpp#L1682
  + ex2: treat differently for two types of columns on drop https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/storage/ducklake_transaction.cpp#L1537

There're two types of fixes:
- Follow the current style and handle new columns and existing columns differently (this is my initial implementation, also what opus suggested, and it works)
- But I think it's very error prune -- you need to remember check column source in every place we delete and drop; so the way I refactor in the PR is to always drop + add
  + It's much easier to reason: when you drop, you have `txn_added_fields` to check where to store drop information
  + No difference when adding columns

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1526,41 +1526,39 @@ struct NewMacroInfo {
 	vector<DuckLakeMacroInfo> new_macros;
 };
 
-void HandleChangedFields(TableIndex table_id, const ColumnChangeInfo &change_info, NewTableInfo &result,
-                         const set<FieldIndex> &columns_handled_by_later_ops,
-                         unordered_map<idx_t, idx_t> &txn_added_fields) {
-	for (auto &new_col_info : change_info.new_fields) {
-		// Skip adding columns that will be handled by a later operation (e.g., SET_DEFAULT after CHANGE_COLUMN_TYPE)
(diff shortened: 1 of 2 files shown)


pdet:
Thanks, LGTM!

PR #1144, opened 2026-05-11
Add ducklake_target_file_size session setting

utay:
This PR adds a `ducklake_target_file_size` session-level setting that overrides the `target_file_size` metadata option for both inserts and compaction. This is useful when concurrent sessions need different file size targets; for example, a background compaction session targeting large files while foreground inserts use a smaller size, without either modifying the shared catalog metadata.

### DIFF
--- a/test/sql/insert/insert_session_target_file_size.test
+++ b/test/sql/insert/insert_session_target_file_size.test
@@ -31,6 +31,11 @@ statement error
 SET ducklake_target_file_size = 'not-a-size'
 ----
 
+# 0 should fail at SET time
+statement error
+SET ducklake_target_file_size = '0'
+----
+
 # Session setting overrides all metadata scopes - should produce multiple files
 statement ok
 SET ducklake_target_file_size = '10KB'
--- a/test/sql/insert/insert_session_target_file_size.test
+++ b/test/sql/insert/insert_session_target_file_size.test
@@ -1,0 +1,54 @@
+# name: test/sql/insert/insert_session_target_file_size.test
+# description: test that ducklake_target_file_size session setting overrides metadata target_file_size
+# group: [insert]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION {TEST_DIR}/{UUID}.db
+
+test-env DATA_PATH {TEST_DIR}
+
+
+statement ok
+ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/ducklake_session_target_file_size')
+
+statement ok
+CREATE TABLE ducklake.test(id INTEGER, s VARCHAR);
+
+# Set a large target_file_size in metadata so everything fits in one file
+statement ok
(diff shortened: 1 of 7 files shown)


test/sql/insert/insert_session_target_file_size.test
```diff
@@ -0,0 +1,54 @@
...
+
+# Session setting should have taken precedence: multiple files expected
+query I
+SELECT COUNT(*) > 1 FROM glob('{DATA_PATH}/ducklake_session_target_file_size/main/test/*.parquet')
+----
+true
+
+# Reset the session setting - metadata value (512MB) should now govern: everything in one file
+statement ok
+RESET ducklake_target_file_size
```
pdet, line 56: Can you add a test where you set ducklake_target_file_size after reset?

test/sql/insert/insert_session_target_file_size.test
```diff
@@ -0,0 +1,54 @@
+# name: test/sql/insert/insert_session_target_file_size.test
```
pdet, line 1: Can you add a test that verifies a borked string fails at set?

e.g., 'not-a-size'

pdet: Can you also cover compaction?

pdet: Can you also validate this option is over other scopes? (e.g., per-schema, per-table)

pdet: what happens if we set to '0'?

pdet (commented):
Hi @utay, thanks for the PR, I have some extra tests requests!

test/sql/insert/insert_session_target_file_size.test
utay: I wanted to test this and realized: after `RESET`, (global) `SET` doesn't work apparently:
```
memory D set ducklake_max_retry_count = 5;
memory D select current_setting('ducklake_max_retry_count');
┌─────────────────────────────────────────────┐
│ current_setting('ducklake_max_retry_count') │
│                   uint64                    │
├─────────────────────────────────────────────┤
│                                           5 │
└─────────────────────────────────────────────┘
memory D reset ducklake_max_retry_count;
memory D select current_setting('ducklake_max_retry_count');
┌─────────────────────────────────────────────┐
│ current_setting('ducklake_max_retry_count') │
│                   uint64                    │
├─────────────────────────────────────────────┤
│                                          10 │
└─────────────────────────────────────────────┘
memory D set ducklake_max_retry_count = 5;
memory D select current_setting('ducklake_max_retry_count');
┌─────────────────────────────────────────────┐
│ current_setting('ducklake_max_retry_count') │
│                   uint64                    │
├─────────────────────────────────────────────┤
│                                          10 │
└─────────────────────────────────────────────┘
```

What do you think is the right behavior?

test/sql/insert/insert_session_target_file_size.test
utay: I added/updated tests with your cases!

pdet:
Thanks!

PR #1147, opened 2026-05-12
In Postgres, use index in ducklake_file_column_stats if that exists.

pdet:
If an index exists in `ducklake_file_column_stats`, e.g.,

```sql
CREATE INDEX IF NOT EXISTS idx_fcs_table_col ON public.ducklake_file_column_stats(table_id, column_id)
```

DuckLake will actually issue a `postgres_query` to take advantage of the index for filters.

Note that the index must currently be manually created by the user. We might revisit the postgres spec on this for DuckLake 1.1.

Performance:

```
  ┌──────────┬────────────┬─────────────────┬───────────────────┬─────────┐
  │  Scale   │ Stats rows │ no_index median │ with_index median │ Speedup │
  ├──────────┼────────────┼─────────────────┼───────────────────┼─────────┤
  │ baseline │         6M │          102 ms │             11 ms │    9.3× │
  ├──────────┼────────────┼─────────────────┼───────────────────┼─────────┤
  │ 10×      │        60M │          2.24 s │             11 ms │    203× │
  ├──────────┼────────────┼─────────────────┼───────────────────┼─────────┤
  │ 100×     │       600M │         30.05 s │             16 ms │   1878× │
  └──────────┴────────────┴─────────────────┴───────────────────┴─────────┘
```


src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -127,6 +127,50 @@ string PostgresMetadataManager::GetLatestSnapshotQuery() const {
...
+		// If the above query errors, we just default to the standard duckdb query later on
+		return false;
+	}
+	auto chunk = result->Fetch();
+	return chunk && chunk->size() > 0;
+}
+
+string PostgresMetadataManager::GenerateFileColumnStatsCTEBody(const CTERequirement &req, TableIndex table_id) {
+	if (!transaction.GetCatalog().HasFileColumnStatsIndex()) {
+		// If we dont have an index we do a duckdb scan
```
Mytherin: Could you expand on why? This seems to add a lot of complexity and I'm not sold on why this is required - I'm quite sure `postgres_query` would always be faster.

pdet: Fair point, was mostly to keep the old behavior, I'll benchmark that and if that's no performance hiccup simplify for `postgres_query`.

PR #1152, opened 2026-05-13
SQL Server catalog backend (mssql metadata manager)

tomanizer:
## Summary
Adds DuckLake catalog support for SQL Server via the existing `ATTACH ... (TYPE sqlserver)` / `mssql` path: `SQLServerMetadataManager`, CMake + patches to build `mssql` when `ENABLE_MSSQL` is set, settings normalization, test config + `ducklake_settings` SQL test.

## Notes for reviewers
- Relies on [mssql-extension](https://github.com/hugr-lab/mssql-extension) with local API-compat patches under `.github/patches/extensions/mssql/`.
- CI may need a follow-up to run SQL Server tests in addition to existing matrix.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/config/extensions/mssql.cmake
+++ b/.github/config/extensions/mssql.cmake
@@ -1,0 +1,12 @@
+if (NOT ${WASM_ENABLED})
+    set(MSSQL_PATCH_DIR "${CMAKE_SOURCE_DIR}/.github/patches/extensions/mssql")
+    file(MAKE_DIRECTORY "${MSSQL_PATCH_DIR}")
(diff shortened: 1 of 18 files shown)


tomanizer:
Closing: intended to iterate on fork first via origin PR instead of upstream for now.

VGSML:
That's amazing! Could you prepare also the PR to the https://github.com/hugr-lab/mssql-extension?
I think it would be better to add attach option to use attached db as the duck lake catalog, so we can change internal mssql extension behavior.

Issue #1155, opened 2026-05-13
Inlined STRUCT columns with non-natively-supported field types fail to read back on Postgres


neel-leendev:
## Description

  When a DuckLake table using a **Postgres metadata backend** has a `STRUCT` column
  containing fields of types not natively supported by that backend (e.g. `TIMESTAMPTZ`/`TIMESTAMP`/`DATE`/`TIME` on Postgres), inserting a small number of rows
  triggers data inlining. The inlined data cannot be read back.

  ## Error
  Conversion Error: invalid timestamp field format: "2026-05-13 04:56:37+00::VARCHAR",
  expected format is (YYYY-MM-DD HH:MM:SS[.US][±HH[:MM[:SS]]| ZONE])

  On v1.5 the stored annotation is `CAST('value' AS VARCHAR)` instead of `'value'::VARCHAR`
  but the error is identical.
```
Traceback (most recent call last):
  File "/Users/neel/workspaces/ducklake-repro/repro.py", line 38, in <module>
    con.execute("CHECKPOINT lake")
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
_duckdb.InternalException: INTERNAL Error: Failed to perform CHECKPOINT; in DuckLake:  Attempted to access index 0 within vector of size 0

Stack Trace:

0        _ZN6duckdb9ExceptionC2ENS_13ExceptionTypeERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEE + 52
1        _ZN6duckdb17InternalExceptionC1ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 20
2        _ZN6duckdb17InternalExceptionC1IJRKyS3_EEERKNSt3__112basic_stringIcNS4_11char_traitsIcEENS4_9allocatorIcEEEEDpOT_ + 136
3        _ZN6duckdb6vectorINSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEELb1ENS5_IS7_EEEixEm + 136
4        _ZN6duckdb23PostgresMetadataManager20TransformInlinedDataERNS_11QueryResultERKNS_6vectorINS_11LogicalTypeELb1ENSt3__19allocatorIS4_EEEE + 84
[...]

PR #1157, opened 2026-05-13
fix: don't annotate non-native type values with ::TYPE inside STRUCT while Inlining

neel-leendev:
When a STRUCT column is inlined on Postgres/SQLite, field values of non-natively supported types (TIMESTAMPTZ, DATE, etc.) were serialized as CAST('value' AS VARCHAR). This annotation survives verbatim in the stored struct string and breaks the struct-from-string parser on read-back, causing an internal crash in TransformInlinedData.

Fix: emit a plain quoted string when !use_native_type since the storage column is already VARCHAR — the annotation is redundant and harmful in this context.

Related to https://github.com/duckdb/ducklake/issues/1155

### DIFF
--- a/src/common/ducklake_util.cpp
+++ b/src/common/ducklake_util.cpp
@@ -147,9 +147,19 @@ string ToSQLString(DuckLakeMetadataManager &metadata_manager, const Value &value
 		// `'value'::type` operator: SQLite's parser rejects `::` outright,
 		// which breaks SQLite-backed metadata backends that ship these
 		// inlined-INSERT batches directly to SQLite.
+		if (!use_native_type) {
+			// When the backend stores this type as VARCHAR, emit a plain quoted
+			// string with no type annotation. The CAST(... AS VARCHAR) suffix
+			// survives verbatim inside STRUCT string literals and breaks the
+			// struct-from-string parser on read-back.
+			return EscapeVarcharForSQL(value.ToString());
+		}
 		return StringUtil::Format("CAST('%s' AS %s)", value.ToString(), value_type);
 	case LogicalTypeId::INTERVAL: {
 		auto interval = IntervalValue::Get(value);
+		if (!use_native_type) {
+			return EscapeVarcharForSQL(value.ToString());
+		}
 		return StringUtil::Format("CAST('%d months %d days %lld microseconds' AS %s)", interval.months, interval.days,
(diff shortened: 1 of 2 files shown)


test/sql/data_inlining/data_inlining_struct_non_native_types.test
```diff
@@ -0,0 +1,39 @@
+# name: test/sql/data_inlining/data_inlining_struct_non_native_types.test
+# description: STRUCT columns with non-natively-supported field types (e.g. TIMESTAMPTZ on Postgres)
```
pdet, line 2: Can we parameterize the test to the other affected types?

i.e.,

```
	case LogicalTypeId::UUID:
	case LogicalTypeId::DATE:
	case LogicalTypeId::TIME:
	case LogicalTypeId::TIME_NS:
	case LogicalTypeId::TIMESTAMP:
	case LogicalTypeId::TIME_TZ:
	case LogicalTypeId::TIMESTAMP_TZ:
	case LogicalTypeId::TIMESTAMP_SEC:
	case LogicalTypeId::TIMESTAMP_MS:
	case LogicalTypeId::TIMESTAMP_NS:
	case LogicalTypeId::BLOB:
	case LogicalTypeId::GEOMETRY:
```

src/common/ducklake_util.cpp
```diff
@@ -147,9 +147,19 @@ string ToSQLString(DuckLakeMetadataManager &metadata_manager, const Value &value
...
+			// string with no type annotation. The CAST(... AS VARCHAR) suffix
+			// survives verbatim inside STRUCT string literals and breaks the
+			// struct-from-string parser on read-back.
+			return EscapeVarcharForSQL(value.ToString());
+		}
 		return StringUtil::Format("CAST('%s' AS %s)", value.ToString(), value_type);
 	case LogicalTypeId::INTERVAL: {
 		auto interval = IntervalValue::Get(value);
+		if (!use_native_type) {
+			return EscapeVarcharForSQL(value.ToString());
```
pdet, line 161: Is the format of `value.ToString()` the same as '%d months %d days %lld microseconds'?

pdet (commented):
Hi @neel-leendev, thanks for the PR, is looking good, just had a few comments!

pdet:

pdet:

PR #1159, opened 2026-05-14
Fix ducklake option segfault

dentiny:
Related to https://github.com/duckdb/ducklake/discussions/1156#discussioncomment-16916661

The root cause is 
- `ducklake_option_array` is declared to have [21 entries](https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/functions/ducklake_options.cpp#L15), but it's only populated with 20 entries
- so when we attempt to get a non-existent option key [here](https://github.com/duckdb/ducklake/blob/c80b9155c8ff260778b6a5bc726e50d65edc84dc/src/functions/ducklake_options.cpp#L90-L97), the non-initialized key-value which default initialized to NULL leads to invalid memory access

In this PR, I use C-style raw array to replace C++ array
- I'm aware of `std::array` is usually recommended because it errors for out-of-bound access, but considering the number of options here, and the fact that it's only used for iteration, I think raw array is still acceptable and preferred here
- C++20 supports template argument deduction so we could write `duckdb::array ducklake_options = {...}`

### DIFF
--- a/src/functions/ducklake_options.cpp
+++ b/src/functions/ducklake_options.cpp
@@ -2,7 +2,6 @@
 #include "storage/ducklake_transaction.hpp"
 #include "storage/ducklake_catalog.hpp"
 #include "storage/ducklake_metadata_manager.hpp"
-#include "duckdb/common/array.hpp"
 #include "duckdb/catalog/catalog_entry/table_catalog_entry.hpp"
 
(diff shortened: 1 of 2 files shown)


pdet:
Thanks!

Issue #1162, opened 2026-05-15
Quack metadata manager remote-executes parquet_full_metadata during ducklake_add_data_files


tobilg:
## Summary

When using DuckLake with a Quack-backed metadata catalog, `ducklake_add_data_files(...)` appears to send the `parquet_full_metadata(...)` inspection query to the remote Quack catalog backend.

This differs from the Postgres metadata catalog path, where the Parquet inspection query is executed locally in the DuckDB process, and only metadata writes are sent to Postgres.

This means a remote Quack metadata service must also have access to the data files and object-store credentials, even though the DuckDB client already has those credentials.

### Why this matters

For serverless Quack metadata backends, this creates an unexpected requirement:

- The DuckDB client has the `TYPE r2` / `TYPE s3` secret needed to read/write DuckLake data files.
- But `ducklake_add_data_files(...)` sends `parquet_full_metadata('r2://...')` through Quack.
- Therefore the remote Quack backend also needs R2/S3 access, or must reimplement Parquet metadata extraction.

This is surprising because the metadata catalog backend should not necessarily need object-store credentials. For example, Postgres does not need to read Parquet files.

### Code path observed

In `ducklake_add_data_files.cpp`, `DuckLakeFileProcessor::ReadParquetFullMetadata(...)` builds a query containing:

```sql
FROM parquet_full_metadata(...)
```

and calls:

```cpp
transaction.Query(...)
```

For the base metadata manager path, `DuckLakeMetadataManager::Query(...)` eventually executes the query locally through `transaction.ExecuteRaw(...)`.

For Postgres, `PostgresMetadataManager::Query(...)` falls back to the base implementation, so `parquet_full_metadata(...)` runs locally in DuckDB. Postgres only overrides `Execute(...)` for metadata writes via `postgres_execute(...)`.

For Quack, `QuackMetadataManager::Query(...)` overrides this behavior and wraps the query in:

```sql
CALL system.main.quack_query_by_name(...)
```

As a result, `parquet_full_metadata(...)` is executed by the remote Quack endpoint instead of the local DuckDB client.

### Expected behavior

`ducklake_add_data_files(...)` should inspect Parquet files in the DuckDB client process, where the relevant filesystem extensions and secrets already exist.

The Quack metadata catalog should only receive the resulting metadata writes/reads that truly belong to the metadata catalog.

### Actual behavior

[...]

PR #1164, opened 2026-05-15
Run ReadParquetFullMetadata in the process always

pdet:
Fix: https://github.com/duckdb/ducklake/issues/1162


tobilg:
Thank you @pdet, that was fast! I assume I can test this tomorrow with the nightly build, right?

pdet:
Uf, not really, the setup for nightly build for extensions is a bit clumsy, and because we are preparing for the v1.5.3 release next week there are no nightlies being built for v1.5.2 anymore :(

tobilg:
I see, thanks for the heads-up!

Issue #999, opened 2026-04-12
BUG: flush sorted inline data leads to incorrect data


dentiny:
### What happens?

See attached SQL for example.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS dl (TYPE ducklake);
memory D USE dl;
dl D CREATE TABLE t(id INTEGER, val INTEGER);
dl D INSERT INTO t VALUES (2, 0), (1, 0), (3, 0);
dl D UPDATE t SET val = val + 1 WHERE id = 2;
dl D UPDATE t SET val = val + 1 WHERE id = 2;
dl D UPDATE t SET val = val + 1 WHERE id = 2;
dl D ALTER TABLE t SET SORTED BY ((id + 0) ASC NULLS LAST);
dl D CALL ducklake_flush_inlined_data('dl', table_name => 't');
┌─────────────┬────────────┬──────────────┐
│ schema_name │ table_name │ rows_flushed │
│   varchar   │  varchar   │    int128    │
├─────────────┼────────────┼──────────────┤
│ main        │ t          │            6 │
└─────────────┴────────────┴──────────────┘
dl D SELECT id, val FROM t ORDER BY id;
┌───────┬───────┐
│  id   │  val  │
│ int32 │ int32 │
├───────┼───────┤
│     2 │     2 │
│     2 │     3 │
│     3 │     0 │
└───────┴───────┘
```
The correct result should be
```sql
┌───────┬───────┐
│  id   │  val  │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     0 │
│     2 │     3 │
│     3 │     0 │
└───────┴───────┘
```

### OS:

macos

### DuckDB Version:

v1.5.1

### DuckLake Version:

6c69da796c95976eefd6e86856a18769afd0c5ac

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

pdet:
@Alex-Monahan maybe you could have a look on this issue?

dentiny:
[...]

PR #1166, opened 2026-05-15
Fix incorrect data when flushing inlined deletes with sorting

Alex-Monahan:
This PR fixes https://github.com/duckdb/ducklake/issues/999 . 

However, there is a previously silently failing condition that this now causes to error, so it is a FIXME for a follow up PR. It will fail if a single transaction contains:
1. A `CREATE MACRO` for a DuckLake macro 
2. A`SET SORTED BY` that uses that macro
3. An inline flush will fail with an error message that the function is missing. 

This is because the connection to the metadata DB does not have the user's DuckLake transaction context where that macro exists. 

I wanted to split it apart since the PR was growing in scope and I think I need some help on how to cleanly handle within-transaction DuckLake macros in this case. This PR at least removes the silent corruption issue and throws an error in the very specific case above instead of a silent failure.


pdet:
Thanks!

Issue #1136, opened 2026-05-08
`ducklake_merge_adjacent_files` leaks RSS when interleaved with INSERTs


zhming0:
### What happens?

When `INSERT` and `ducklake_merge_adjacent_files` alternate (the normal compaction
workload), RSS grows per cycle while `duckdb_memory()` barely moves. With the reproducer script, over 100 cycles on a table partitioned by `(year, month, day, service)`:

| cycle | RSS | duckdb_memory() |
|---:|---:|---:|
| 1 | 79 MB | 13 MB |
| 10 | 113 MB | 29 MB |
| 30 | 156 MB | 40 MB |
| 50 | 169 MB | 45 MB |
| 70 | 185 MB | 44 MB |
| 100 | **201 MB** | **50 MB** |

This issue is more obvious on production:

<img width="948" height="325" alt="Image" src="https://github.com/user-attachments/assets/c66923fd-8a3a-4813-9d4c-3d93f82e2bba" />

Note: previously I have reported #1018 and it seems to be largely fixed, now inserts along won't trigger memory leak. 

### To Reproduce

My clanker helped made this reproducer. 

```python
#!/usr/bin/env python3
"""Usage: python3 merge_leak_repro.py [--cycles 30] [--duckdb build/release/duckdb]
       [--ext build/release/extension/ducklake/ducklake.duckdb_extension]

Prints duckdb_memory() after each cycle. On macOS monitor real physical
memory with: while sleep 1; do vmmap -summary $PID | grep 'Physical footprint'; done
"""
import subprocess, os, shutil, argparse, sys, time
from datetime import datetime, timedelta

args = argparse.ArgumentParser()
args.add_argument("--cycles", type=int, default=30)
args.add_argument("--duckdb", default="build/release/duckdb")
args.add_argument("--ext", default="build/release/extension/ducklake/ducklake.duckdb_extension")
args = args.parse_args()

DB, DATA = "/tmp/merge_leak.duckdb", "/tmp/merge_leak_data"
shutil.rmtree(DB, ignore_errors=True)
shutil.rmtree(DATA, ignore_errors=True)
os.makedirs(DATA)

p = subprocess.Popen([args.duckdb, "-unsigned"], stdin=subprocess.PIPE,
                     stdout=sys.stdout, stderr=sys.stderr, text=True)
sql = lambda s: (p.stdin.write(s + ";\n"), p.stdin.flush())
[...]

PR #1167, opened 2026-05-16
Cache per-table stats

dentiny:
Hi team, recently I see more and more issue report on memleak / excessive memory consumption so I'd love to take a look, analysis see: https://github.com/duckdb/ducklake/issues/1136#issuecomment-4414008087

In the current implementation, we use [`DuckLakeStats`](https://github.com/duckdb/ducklake/blob/d897bc5a35f887c9087403bc20efd92d99272e69/src/include/storage/ducklake_stats.hpp#L66-L78) as the cache entry, which contains all the tables and all their columns, thus not scalable.

In this PR, I update the cache granularity to use table as the cache entry, which mimics DuckDB core [`ParquetMetadataCache`](https://github.com/duckdb/duckdb/blob/c7a9c8fe3bcd10a923037b8e03ab7936515e47c4/extension/parquet/include/parquet_file_metadata_cache.hpp).

Two considerations:
- I thought about using table column as the cache unit, but we have a few functions to get table stats, thought it'd be an overkill
- This PR targets at v1.5 branch, because I found all recent PRs (whether bug or feature) all land against v1.5 instead of main (maybe we should merge v1.5 commits to main branch?)

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -30,17 +30,17 @@ struct DuckLakeConfigOption;
 struct DeleteFileMap;
 class LogicalGet;
 
-//! Cache entry for DuckLake table statistics
-struct DuckLakeStatsCacheEntry : public ObjectCacheEntry {
+//! Per-table stats cache entry, keyed by <next_file_id, table_id>.
+struct DuckLakeTableStatsCacheEntry : public ObjectCacheEntry {
 	static constexpr idx_t ESTIMATED_BYTES_PER_COLUMN_STATS = 256;
 
-	explicit DuckLakeStatsCacheEntry(unique_ptr<DuckLakeStats> stats_p) : stats(std::move(*stats_p)) {
+	explicit DuckLakeTableStatsCacheEntry(DuckLakeTableStats stats_p) : stats(std::move(stats_p)) {
 	}
 
-	DuckLakeStats stats;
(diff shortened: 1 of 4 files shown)


pdet:
What are the performance considerations of this change?
From a quick look, it seems to me we will increase the RTs to the catalog to tables in the same snapshot. Maybe we should LRU per snapshot cache instead?

dentiny:
> What are the performance considerations of this change?
From a quick look, it seems to me we will increase the RTs to the catalog to tables in the same snapshot.

I don't expect performance hurt, since per-table stats cache still lives in LRU cache (this PR just change the entry granularity). Catalog manager is only involved when stats cache entries evicted.

pdet:
I think that will will have perf implications for queries that have multi-table access, this is more of a trade-off, more round-trips for multi-table queries versus smaller round-trips. Ideally we would fetch all tables required for a given query together. But that's more difficult. 

I also don't think the issue you linked is fixed by this change, right? Since the issue is related to a single-table.

In any case, I can see this being better than the current implementation in many real-world scenarios, while joins and such are somewhat rare and expensive anyway, so I'll merge it.

dentiny:
> I think that will will have perf implications for queries that have multi-table access

Yes -- the new implementation makes single table stats easier to be evicted under memory pressure.

dentiny:
> I also don't think the issue you linked is fixed by this change, right

Yes, I'm working on a fix for external file cache memleak.

Issue #1135, opened 2026-05-08
BUG: default value cannot be queries


dentiny:
### What happens?

When new columns are added with default value, it fails to be queries in some cases, see repro sql for details.

### To Reproduce

```sql
memory D ATTACH ':memory:' AS other (TYPE ducklake);
memory D CREATE TABLE other.tbl (a INT);
memory D INSERT INTO other.tbl SELECT i FROM range(2000) t(i);
memory D ALTER TABLE other.tbl ADD COLUMN b INT DEFAULT 42;
memory D SELECT count(*) FROM other.tbl WHERE b = 42;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
-- should be 2000!
```

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckLake Version:

v1.0.0

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1170, opened 2026-05-18
Fix mis-pruned data files

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1135

The bug is: when we get data files, current implementation only select those with columns which matches the predicate (i.e., col_b <= 42), but miss all data files without the requested column.

Data file query before this fix:
```sql
data.data_file_id IN (
	SELECT data_file_id FROM col_2_stats
	WHERE (value_count IS NULL OR value_count > 0)
	  AND (min_value IS NULL OR max_value IS NULL OR (min_value <= '42' AND max_value >= '42'))
  )
```
After the fix
```sql
(
	data.data_file_id NOT IN (SELECT data_file_id FROM col_2_stats)
	OR
	data.data_file_id IN (
	  SELECT data_file_id FROM col_2_stats
	  WHERE (value_count IS NULL OR value_count > 0)
		AND (min_value IS NULL OR max_value IS NULL OR (min_value <= '42' AND max_value >= '42'))
	)
  )
```

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1306,16 +1306,23 @@ FilterSQLResult DuckLakeMetadataManager::ConvertFilterPushdownToSQL(const Filter
 		if (!conditions.empty()) {
 			conditions += " AND ";
 		}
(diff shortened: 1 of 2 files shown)


pdet:
Thanks!

PR #1171, opened 2026-05-18
Optimize partition-only deletes

jdctinuiti:
## Summary

This PR adds a metadata-only DELETE path for DuckLake tables when the DELETE can be proven to remove whole data files, and fixes table-statistics accounting when whole files are dropped.

## Metadata-only partition DELETE

For DELETE predicates over identity partition columns, DuckLake can retire matching data files by updating metadata instead of scanning files and writing row-level delete files.

This path is used only when all candidate files are safe to drop:

- the predicate is reducible to accepted constant values for identity partition columns (`=` / `IN`)
- every candidate file has recorded partition values matching the accepted set
- no candidate file has active row-level delete files
- no inlined data or active inlined deletion table is involved
- non-identity partition transforms, mixed non-partition predicates, NULL partition predicates, and legacy files without partition metadata fall back to the existing row-level delete path

Full-table DELETE on partitioned tables can also use this path when the scan selects whole active data files.

## Stats fix

Whole-file drops now decrement `ducklake_table_stats.record_count` and `file_size_bytes`.

Before this PR, full-file deletes retired data files but left global table stats unchanged, so repeated delete/reinsert cycles inflated:

- optimizer cardinality estimates via `GetCardinality`
- `duckdb_tables` size reporting

`next_row_id` remains monotonic. Column min/max stats are left unchanged because exact recomputation would require scanning remaining files.

The stats cache is also invalidated after delete-only commits that drop files, since `next_file_id` does not change for those commits.

## Tests

Adds coverage for:

- metadata-only partition DELETE with `=` and `IN`
- intersecting partition filters
- fallback for mixed predicates, NULL predicates, transformed partitions, legacy files, existing deletes, and inlined data
- atomic delete + `ducklake_add_data_files`
- concurrent conflicting metadata deletes
- table-stats decrement and stats-cache invalidation

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -148,6 +148,7 @@ public:
 	                             optional_ptr<PhysicalOperator> plan) override;
 	PhysicalOperator &PlanCreateTableAs(ClientContext &context, PhysicalPlanGenerator &planner, LogicalCreateTable &op,
 	                                    PhysicalOperator &plan) override;
+	PhysicalOperator &PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner, LogicalDelete &op) override;
 	PhysicalOperator &PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner, LogicalDelete &op,
 	                             PhysicalOperator &plan) override;
 	PhysicalOperator &PlanUpdate(ClientContext &context, PhysicalPlanGenerator &planner, LogicalUpdate &op,
--- a/src/include/storage/ducklake_delete.hpp
+++ b/src/include/storage/ducklake_delete.hpp
@@ -160,7 +160,8 @@ public:
 	static PhysicalOperator &PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner,
 	                                    DuckLakeTableEntry &table, PhysicalOperator &child_plan,
 	                                    vector<idx_t> row_id_indexes, string encryption_key,
-	                                    bool allow_duplicates = true);
+	                                    bool allow_duplicates = true,
+	                                    unordered_map<idx_t, vector<Value>> metadata_delete_partition_values = {});
 
 public:
 	// Sink interface
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -421,6 +421,7 @@ struct DuckLakeFileListExtendedEntry {
 	idx_t delete_count = 0;
 	MappingIndex mapping_id;
 	DuckLakeDataType data_type = DuckLakeDataType::DATA_FILE;
+	vector<DuckLakeFilePartitionInfo> partition_values;
 };
 
 struct DuckLakeCompactionBaseFileData {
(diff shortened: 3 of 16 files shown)


kinghuang:
Nice, I've been looking for something like the efficiency of attaching and dropping table partitions in PostgreSQL.

jdctinuiti:
Rebased this stack onto current origin/main and cleaned the branch history so the PRs carry only their feature changes. I intentionally left unrelated DuckDB-main compatibility fixes out of this PR stack; those should land separately. The only compatibility cleanup folded into #1171 is for expression accessors in the metadata-delete planner code introduced by this PR.

pdet:
@jdctinuiti can you resolve the conflicts?

cc @Alex-Monahan can you have a look/review at this PR when you have the time?

jdctinuiti:
> @jdctinuiti can you resolve the conflicts?
> 
> cc @Alex-Monahan can you have a look/review at this PR when you have the time?

@pdet Conflicts resolved here as well as in the rest of the stack  (#1172/#1182/#1220)

pdet:
Hi @jdctinuiti, thanks for the PRs, I believe this PR and the subsequent list add a lot of complexity, making them hard to review. There are also a few new bugs I encountered from the code, will put some examples at the end of the message.
I think the way to go forward with this is to break this PR down into multiple PRs, for example:
1. Decrement stats when a delete empties a file
2. Keep externally-added files on disk when dropped transaction-locally
3. Surface current-spec partition values per file
 4. Metadata delete for a single partition column (equality / IN)
 ...

This will allow us to properly review them.  If you're up for it, please open them one at a time and let each merge before the next, rather than stacking, that's what will let us review each properly.

I'll close this PR and the stacked ones for now so we can restart from a clean base.

Examples (these silently drop rows the predicate should keep,  they all return the wrong count and pass on main):

```sql
-- 1) column = column empties the whole table
CREATE TABLE t1 (a INT, b INT, v INT);
ALTER TABLE t1 SET PARTITIONED BY (a, b);
INSERT INTO t1 VALUES (1,1,10),(1,2,20),(2,2,30),(2,3,40),(3,3,50);
DELETE FROM t1 WHERE a = b;      -- returns 5, should be 3; (1,2) and (2,3) lost

-- 2) cross-column OR on a multi-column partition over-deletes
CREATE TABLE mc (id INT, a VARCHAR, b VARCHAR, v INT);
ALTER TABLE mc SET PARTITIONED BY (a, b);
INSERT INTO mc VALUES (1,'x','p',10),(2,'x','q',20),(3,'y','p',30),(4,'y','q',40);
DELETE FROM mc WHERE a = 'x' OR b = 'p';   -- returns 4, should be 3; id=4 lost

-- 3) after repartitioning, a colliding partition_key_index drops old-spec files
CREATE TABLE rp (a VARCHAR, b VARCHAR, v INT);
ALTER TABLE rp SET PARTITIONED BY (a);
INSERT INTO rp VALUES ('x','aaa',1),('x','zzz',2);
ALTER TABLE rp SET PARTITIONED BY (b);
INSERT INTO rp VALUES ('q','x',3);
DELETE FROM rp WHERE b = 'x';    -- returns 3, should be 1; the two old-spec rows lost
```

jdctinuiti:
Hey @pdet thanks for the review! That sounds great, will get on it!

PR #1174, opened 2026-05-20
Test S3 settings propagation for ducklake_add_data_files

jdctinuiti:
Main now propagates caller settings to DuckLake internal connections via CopyExtensionSettings, which covers the S3 case this PR originally implemented, so I dropped the redundant CopyFileAccessSettings helper. What remains is a test that confirms ducklake_add_data_files inherits S3 settings configured with SET rather than CREATE SECRET, a path the existing checkpoint test does not cover.


jdctinuiti:
@pdet Rebased on main and resolved the conflict. Since CopyExtensionSettings (965dc6cb) already propagates the caller's S3 settings to internal connections, this PR is now just a test for the ducklake_add_data_files path, which the checkpoint test does not cover. Happy to close it instead if you would rather not keep the extra test.

pdet:
Thanks!

PR #1176, opened 2026-05-20
Add bucket-transform-aware partition pruning

guillesd:
Equality and IN-list predicates on `bucket()`-partitioned columns now prune files via `ducklake_file_partition_value` at scan-planning time. Bucket hashes are computed in C++ so the catalog SQL stays backend-agnostic. Range predicates fall through to the existing zone-map path.

Basically makes lookup queries useful with buckets. Zone-maps are not really effective for hashed transforms.

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -345,6 +345,11 @@ private:
 	string GetDeleteFileSelectList(const string &prefix);
 	FilterPushdownQueryComponents GenerateFilterPushdownComponents(const FilterPushdownInfo &filter_info,
 	                                                               TableIndex table_id);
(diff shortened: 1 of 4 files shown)


guillesd:
We could alternatively make `BuildBucketPartitionPruningClause` part of `GenerateFilterPushdownComponents` and centralize everything there. I'd just need to change the signature of the function to get `DuckLakeTableEntry` to have the partition info. Let me know what you think @pdet

pdet:
Thanks, LGTM, can you just fix the failing CI?

guillesd:
@pdet done! Basically all partitioning tests are skipped in Quack in the first place because of a known limitation. So I just added the new test to the skip list only for Quack!

pdet:
Thanks!

PR #1178, opened 2026-05-21
Do not use table metadata to determine emptyness

NiclasHaderer:
To determine the emptyness of a inlined table, ducklake currently introspects the estimated_size of duckdb tables. However this check is unreliable, since `estimated_cardinality` is not transactional. Additionally, this leads to wrong results in MotherDuck because `estimated_cardinality` can be stale at times


pdet:
Thanks!

PR #1180, opened 2026-05-21
Fix stats merging for variant columns with incompatible shredded types

NiclasHaderer:
When different files shred the same variant field to different types (e.g., one file shreds as `INT`, another as `VARCHAR`), the stats merge would previously attempt to compare min/max values across incompatible types, and leave the `shredded_type` in an inconsistent state.

This fixes two things:
- When types differ, invalidate min/max instead of attempting cross-type comparison
- When a shredded field has incompatible types across files, drop it from the table-level shredded stats entirely. This matches how DuckDB core handles it. Fields with matching types across all files are preserved with correctly merged stats


pdet:
Thanks, LGTM!

PR #1181, opened 2026-05-21
Derive DuckLakeCompactor hive partition prefix from partition spec + values

jdctinuiti:
`ducklake_merge_adjacent_files` reconstructed the hive partition prefix by string-manipulating the first file's path, which silently corrupted writes when the source file path didn't share `table.DataPath()` (e.g. files added via `ducklake_add_data_files` with absolute paths) — the absolute prefix was reused as a relative write target and the compacted file ended up nested at a junk S3 key. Derive the prefix from the partition spec and values instead, so the compactor always writes a relative `col=value/` under `table.DataPath()` regardless of where source files live.

### DIFF
--- a/src/storage/ducklake_partition_data.cpp
+++ b/src/storage/ducklake_partition_data.cpp
@@ -1,5 +1,7 @@
 #include "storage/ducklake_partition_data.hpp"
 #include "common/ducklake_murmur3.hpp"
+#include "storage/ducklake_table_entry.hpp"
+#include "duckdb/common/hive_partitioning.hpp"
 #include "duckdb/common/types/value.hpp"
 #include "duckdb/function/function_binder.hpp"
 #include "duckdb/planner/expression.hpp"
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_partition_data.cpp
```diff
@@ -84,6 +86,44 @@ string DuckLakePartitionUtils::BuildPartitionFilter(const vector<string> &partit
...
+	if (partition_data->fields.size() != partition_values.size()) {
+		throw InternalException("DuckLake partition value count does not match partition spec");
+	}
+	string result;
+	case_insensitive_set_t used_names;
+	for (auto &field : partition_data->fields) {
+		if (field.partition_key_index >= partition_values.size()) {
+			throw InternalException("DuckLake partition key index is out of range");
+		}
+		auto &column = table.GetColumnByFieldId(field.field_id);
```
guillesd: Just a nit, but to keep consistency, we do this in ducklake_insert:
```cpp
// src/storage/ducklake_insert.cpp
auto field_id = copy_input.field_data->GetByFieldIndex(field.field_id);
return DuckLakePartitionUtils::GetPartitionKeyName(field.transform.type, field_id->Name(), names);
```
So maybe for consistency we should do the same here

test/sql/compaction/merge_adjacent_external_hive_paths.test
```diff
@@ -0,0 +1,111 @@
+# name: test/sql/compaction/merge_adjacent_external_hive_paths.test
```
guillesd, line 1: On the testing side I'm missing a couple of things, in order of priority:
1. A test that uses a transform (year/month ..) and then does compaction
2.  A single test with a "weird" string partition that checks if `Escape/ToString ever drifted`. The reason for this is that I believe `physical_copy_to_file.cpp` has a slightly different path building than `BuildHivePartitionPath`. Maybe you can also check this divergence.
3. Maybe the add_files scenario but using a partition transform (year() or month()). This is just fuzzing at this point, but doesn't hurt.

src/storage/ducklake_partition_data.cpp
```diff
@@ -84,6 +86,44 @@ string DuckLakePartitionUtils::BuildPartitionFilter(const vector<string> &partit
...
+		auto partition_key_name = GetPartitionKeyName(field.transform.type, column.GetName(), used_names);
+		used_names.insert(partition_key_name);
+
+		auto &partition_value = partition_values[field.partition_key_index];
+		string partition_value_str;
```
guillesd, line 115: Maybe check if this diverges from https://github.com/duckdb/duckdb/blob/v1.5-variegata/src/execution/operator/persistent/physical_copy_to_file.cpp#L121

guillesd (changes requested):
Just some minor things @jdctinuiti !

src/storage/ducklake_partition_data.cpp
jdctinuiti: Hello @guillesd thank you for the review :DD

This should now be consistent!

test/sql/compaction/merge_adjacent_external_hive_paths.test
jdctinuiti: Added now!

src/storage/ducklake_partition_data.cpp
jdctinuiti: I checked this against physical_copy_to_file.cpp, the path construction is aligned: escaped key, =, null sentinel for NULL, otherwise escaped Value::ToString(). The new weird-string test compares the DuckDB-written hive segment with the DuckLake compaction output.

guillesd:
Hi @jdctinuiti if you could merge 1.5 into your branch, that would fix a pending CI issue on windows. Once that's done this is ready to merge!

jdctinuiti:
@guillesd done, thank you! I have other (some admittedly larger) PRs open split between 1.5-variegata and main, will those also get reviewed?

guillesd:
@jdctinuiti I'll have a look at some point this week to see if there's something there. Since you are a first-time contributor we need to start small and build from there. This is a good example

guillesd:
Hey @jdctinuiti can you take a look at the test failure?

jdctinuiti:
> Hey @jdctinuiti can you take a look at the test failure?

Should be good now had to edit the regex for Windows

jdctinuiti:
@guillesd retargeted to main per @pdet 's suggestion

jdctinuiti:
Any updates @guillesd ?

guillesd:
I've flagged it as ready for @pdet ! thanks!

pdet:
Thanks!

PR #1172, opened 2026-05-18
Add ducklake_drop_data_files procedure

jdctinuiti:
Stacked on #1171. Adds `ducklake_drop_data_files`, a procedure for retiring data files by typed partition values (e.g. `partition_values => struct_pack(year := 2024, month := 1)`). Supports `dry_run`, identity/transformed partition keys, transaction-local files, and concurrent conflicts. Please review only the top commit until #1171 is merged.

PR #1182, opened 2026-05-21
Support partition filters for compaction

jdctinuiti:
Stacked on #1220. Adds a table-specific `partition_values` named parameter to `ducklake_merge_adjacent_files` and `ducklake_rewrite_data_files` that filters compaction candidates by partition values, reusing the `DuckLakePartitionFilter` helper introduced in #1172. Only registered for the table-specific overloads, since catalog-wide compaction would make a single filter ambiguous across tables. Metadata-query pruning inside `GetFilesForCompaction` is left for a follow-up performance PR.


jdctinuiti:
Rebuilt on top of the updated #1220.

pdet:
See https://github.com/duckdb/ducklake/pull/1171

PR #1188, opened 2026-05-26
Improve schema cache entry estimated size

dentiny:
Followup PR for https://github.com/duckdb/ducklake/pull/1033

In the last implementation, the memory consumption estimation is kinda coarse, it uses a [fixed size](https://github.com/duckdb/ducklake/blob/d897bc5a35f887c9087403bc20efd92d99272e69/src/storage/ducklake_catalog.cpp#L44-L48) for each catalog entry, which could be far off if people using tags, or wide table. etc.
This PR improves estimation by check the detailed table/view/macro entry.

### DIFF
--- a/src/storage/ducklake_catalog.cpp
+++ b/src/storage/ducklake_catalog.cpp
@@ -33,6 +33,118 @@
 
 namespace duckdb {
 
+namespace {
+
+idx_t EstimateStringMemory(const string &value) {
+	return sizeof(string) + value.length();
+}
+
+idx_t EstimateFieldIdCount(const DuckLakeFieldId &field_id) {
+	idx_t count = 1;
+	for (const auto &child : field_id.Children()) {
+		count += EstimateFieldIdCount(*child);
+	}
+	return count;
+}
+
+idx_t EstimateTagMemory(const CatalogEntry &entry) {
+	idx_t estimate = 0;
+	for (auto &tag : entry.tags) {
+		estimate += EstimateStringMemory(tag.first);
+		estimate += EstimateStringMemory(tag.second);
+	}
+	return estimate;
+}
+
+idx_t EstimateTableEntryMemory(const DuckLakeTableEntry &table) {
+	idx_t estimate = EstimateTagMemory(table);
+	estimate += table.GetColumns().LogicalColumnCount() * sizeof(ColumnDefinition);
+	for (const auto &column : table.GetColumns().Logical()) {
+		estimate += EstimateStringMemory(column.Name());
+	}
+	estimate += table.GetConstraints().size() * sizeof(Constraint);
+	for (const auto &field_id : table.GetFieldData().GetFieldIds()) {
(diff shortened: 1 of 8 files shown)


dentiny:
Hey @guillesd / @Alex-Monahan , I'm wondering if you're willing to take a look when you have some time? Should be straightforward :)

src/storage/ducklake_schema_entry.cpp
```diff
@@ -558,4 +565,20 @@ DuckLakeCatalogSet &DuckLakeSchemaEntry::GetCatalogSet(CatalogType type) {
 	}
 }
 
+const DuckLakeCatalogSet &DuckLakeSchemaEntry::GetCatalogSet(CatalogType type) const {
```
guillesd, line 568: Apparently there is an existing GetCatalogSet that is not a const. Maybe you can reuse that one and use `const_cast` as to not having diverging behaviours?

guillesd: This is a recommendation. Maybe there are some drawbacks to this I am not aware of

src/storage/ducklake_catalog.cpp
```diff
@@ -33,16 +33,126 @@
...
+			estimate += EstimateStringMemory(field.expression);
+			estimate += EstimateStringMemory(field.dialect);
+		}
+	}
+	return estimate;
+}
+
+idx_t EstimateViewEntryMemory(const DuckLakeViewEntry &view) {
+	idx_t estimate = EstimateTagMemory(view);
+	estimate += view.aliases.size() * sizeof(string);
```
guillesd, line 98: I think we are missing the actual view SQL no? i.e.

```c++
estimate += EstimateStringMemory(view.GetQuerySQL());
```

guillesd (changes requested):
Just a couple of things @dentiny! thanks

src/storage/ducklake_schema_entry.cpp
dentiny: > Maybe you can reuse that one and use const_cast as to not having diverging behaviours?

~`const_cast` is discouraged in C++ -- compiler does weird optimization sometimes
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-const-cast.html~

NVM, adding const should be safe here.

src/storage/ducklake_catalog.cpp
dentiny: Yes, in this PR I only considered those which could be countless; let me add the left then.

src/storage/ducklake_schema_entry.cpp
guillesd: Hey @dentiny sorry, actually I reviewed the existing patterns and it seems like `const_cast` is not really used in duckdb core. Better to keep as is

src/storage/ducklake_catalog.cpp
dentiny: Updated.

guillesd (approved):
Thanks!

PR #1191, opened 2026-05-27
Server-side transaction retrial for Quack

pdet:
With this PR, when using duckdb-quack as a catalog we can do transaction retrials directly on the server-side. This is beneficial as we drastically reduce the number of round-trips between client and server, which are a major bottleneck for streaming workloads.
  
 This works as follows: If the quack server has the ducklake extension installed and loaded, we can use the retrial logic existing in the ducklake extension server-side, instead of going through the typical client-server retrial loop. To achieve this, when committing a transaction, we serialize the commit state into a set of staging tables on the server, then call a server-side `ducklake_commit(...)` function that uses the existing commit logic of DuckLake to perform conflict detection and retrials entirely server-side.

## Benchmark
I ran a simple benchmark to measure the benefits of this solution. In our benchmark we simply do single row insertions on the same table over multiple workers (16), I've used an EC2 machine and my own laptop. 

| Catalog DBMS           | Transactions/second |
| ---------------------- | ------------------- |
| Quack (Server Retrial) | 145                 |
| Quack (Client Retrial) | 8                 |
| Postgres               | 2                   |

So with 16 workers, we can see nearly a two orders of magnitude difference between Quack and Postgres.

## Code Refactor

To allow the existing code to be reused, I had to refactor quite a lot of it, with the basic idea being to remove the dependencies of `ATTACH` only structs (e.g., `DuckLakeTransaction`, `DuckLakeCatalog`, ...), this was achieved by creating a dependency-free structure in `DuckLakeTransactionState`, all methods bubble-up these dependencies, and some of the functionality is exposed as callbacks in `DuckLakeCommitContext`.

To allow the existing commit code to be reused server-side, I had to refactor quite a lot of it. The basic idea was to remove the dependency on ATTACH-only types (e.g., `DuckLakeTransaction`, `DuckLakeCatalog`, `DuckLakeMetadataManager`) from the commit logic. This was achieved by extracting a dependency-free `DuckLakeTransactionState` class that owns all the mutable commit state (local file changes, dropped files, flushed inlined tables, schema/table/macro changes, etc.) and the `Commit()` loop itself. Necessary methods that lived in the `DuckLakeTransaction` (e.g., `CommitChanges`, `CheckForConflicts`, `WriteSnapshotChanges`,...) have also been moved into  `DuckLakeTransactionState`.

Some of the functionality that depended in catalog interactions have been wrapped in function calls (e.g., query execution, snapshot fatching, inlined data writes ,...), all defined in the `DuckLakeCommitContext`.

## Commit workflow

The refactor is ~75% of this PR. The remaining is the actual server-side retrial, which works as follows:

We first transform the commit data into temporary staging tables annotated, for example: 

```sql
 ducklake_staged_commit
  CREATE TABLE IF NOT EXISTS ducklake_staged_commit(
      commit_author VARCHAR, 
      commit_message VARCHAR,
      commit_extra_info VARCHAR,
      transaction_snapshot_id BIGINT,
      data_path VARCHAR,
      path_separator VARCHAR,
      transaction_next_catalog_id UBIGINT,
      transaction_next_file_id UBIGINT
  );  
  
  ducklake_staged_data_file
  CREATE TABLE IF NOT EXISTS ducklake_staged_data_file(
      data_file_id BIGINT,
      table_id BIGINT,
      file_order BIGINT,
      path VARCHAR,
      path_is_relative BOOLEAN,
      file_format VARCHAR,
      record_count BIGINT,
      file_size_bytes BIGINT,
      footer_size BIGINT,
      row_id_start BIGINT,
      partition_id BIGINT,
      encryption_key VARCHAR,
      mapping_id UBIGINT,
      partial_max BIGINT,
      begin_snapshot BIGINT,
      compaction_id BIGINT
  );  
 
```
There are 17 table types in total covering data files, column stats, partition values, delete files, inlined data/deletes, compactions, name maps,  etc...

We then batch all CREATE TABLE + INSERT statements together with a single `ducklake_commit `call into one SQL string that is sent to the server in a single round-trip. The call looks like:

```
FROM ducklake_commit(
      '<uuid>',
      '<metadata_schema>',
      <schema_version>,   
      max_retry_count => 10,
      retry_wait_ms => 100,
      retry_backoff => 1.5
  );
```
and it returns three columns: (committed_snapshot_id BIGINT, committed_schema_version BIGINT, had_flushes BOOLEAN). Since this is all batched, we only have one client-server call for the entire commit (or zero extra if the snapshot fetch was also skipped).

The server then hydrates the staging tables into C++ structures (`DuckLakeTransactionState`, `LocalTableChanges`, etc.) that plug directly into the existing commit code. The commit will either fail, with the error propagated back to the client as a query error, or it will succeed, returning the committed snapshot id and schema version, which the client uses to update its local catalog version and committed snapshot cache. If had_flushes is true, the client also cleans up superseded inlined tables and clears the quack cache. Either way, the temporary tables are truncated at the end of the server-side function, for later reuse. Since they are temporary, they will be cleaned-up client side.

## Current Limitations
This PR only supports data retrials, so transactions with DLL operations (e.g., table alter, table creation, schema creation) will still go through the client-server route. 

## Possible Optimizations
We can still optimize this further. A few ideias:
1. Appender fast-path: The `try_append_data_files` callback is a no-op on the server side, so data files are always committed via the SQL batch path rather than the DuckDB Appender optimization.
2. Revisit round-trips in Cold-Paths: If a DDL transaction happens, client caches are cleared, which will increase the number of RT queries to get them up to date.
3. Reduce the number of round-trips during hydration. These should have little impact on network latency, but are still several separate calls to the server.


Alex-Monahan:
Wow, this is super cool! It's like a C++ stored procedure! 
I was able to get more like 75 TPS out of Postgres in some tests. Is it the 16 threads causing a retry storm maybe that brings it down to 2?
I also wonder if it is possible to go faster than 100 TPS, maybe with a faster retry time? I can do some testing - anything you recommend experimenting with?

pdet:
Hi @Alex-Monahan, 

You are absolutely right, is like a C++ stored procedures. I think this PR also open up a bit of the architecture to do that with Postgres as a catalog in the future.

> I was able to get more like 75 TPS out of Postgres in some tests. Is it the 16 threads causing a retry storm maybe that brings it down to 2?

I guess you got that with a single client? But indeed, the multi-client approach is what will cause the retrials, since we have snapshot isolation, each parallel insertion will cause a retrial due to snapshot conflict.

> I also wonder if it is possible to go faster than 100 TPS, maybe with a faster retry time? I can do some testing - anything you recommend experimenting with?

I've experimented with different wait_times, 0 seems to cause a bit of contention, but a low number seems to be generally beneficial.

Alex-Monahan:
My test did try with separate processes (and separate threads as well) and didn't see a big gap between them. All processes were on the same EC2 machine. Still, I couldn't crack 100 though! Exciting stuff - I can't wait to tinker...

Thanks Pedro!

pdet:
> My test did try with separate processes (and separate threads as well) and didn't see a big gap between them. All processes were on the same EC2 machine. Still, I couldn't crack 100 though! Exciting stuff - I can't wait to tinker...
> 
> Thanks Pedro!

Ah, in my case is my laptop -> EC2 machine, so the network cost is much higher. So your retrials are of course much cheaper since you eliminate ze network.

pdet:
> My test did try with separate processes (and separate threads as well) and didn't see a big gap between them. All processes were on the same EC2 machine. Still, I couldn't crack 100 though! Exciting stuff - I can't wait to tinker...
> 
> Thanks Pedro!

Managed to get close to 150 tx/s by using temporary tables and truncating them :-)

Alex-Monahan:
> > My test did try with separate processes (and separate threads as well) and didn't see a big gap between them. All processes were on the same EC2 machine. Still, I couldn't crack 100 though! Exciting stuff - I can't wait to tinker...
> > Thanks Pedro!
> 
> Managed to get close to 150 tx/s by using temporary tables and truncating them :-)

Nice!! Love to see it.

I did my best to be fair on my test - I was using AWS RDS for Postgres, so there was still a network hop in there from EC2 to RDS. I haven't even tried putting things on the same box - I'm sure that would be quite speedy!

dentiny:
I met an issue.

Server side
```sql
memory D FORCE INSTALL quack FROM core_nightly;
memory D LOAD quack;
memory D LOAD httpfs;
memory D LOAD ducklake;
memory D SELECT * FROM quack_serve('quack://localhost:19999/', token := 'ducklake-test-token');
┌──────────────────────────┬────────────────────────┬─────────────────────┐
│        listen_uri        │       listen_url       │     auth_token      │
│         varchar          │        varchar         │       varchar       │
├──────────────────────────┼────────────────────────┼─────────────────────┤
│ quack://localhost:19999/ │ http://localhost:19999 │ ducklake-test-token │
└──────────────────────────┴────────────────────────┴─────────────────────┘
```
client side
```sql
memory D LOAD httpfs;
memory D LOAD quack;
memory D LOAD ducklake;
memory D CREATE OR REPLACE SECRET ducklake_test_quack (
           TYPE quack,
           TOKEN 'ducklake-test-token'
         );
┌─────────┐
│ Success │
│ boolean │
├─────────┤
│ true    │
└─────────┘
memory D ATTACH 'ducklake:quack:localhost:19999' AS ducklake (
           DATA_PATH '/tmp/ducklake_pr1191_inline_lazy',
           METADATA_CATALOG 'dl',
           DATA_INLINING_ROW_LIMIT 0
         );
memory D CREATE TABLE ducklake.t(i INTEGER);
memory D CALL ducklake.set_option('data_inlining_row_limit', 10);
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D INSERT INTO ducklake.t VALUES (42);
memory D SELECT count(*), sum(i) FROM ducklake.t;
┌──────────────┬────────┐
│ count_star() │ sum(i) │
│    int64     │ int128 │
├──────────────┼────────┤
│            0 │   NULL │
└──────────────┴────────┘
```

src/functions/ducklake_commit.cpp
```diff
@@ -0,0 +1,78 @@
...
+                                                   vector<LogicalType> &return_types, vector<string> &names) {
+	for (idx_t i = 0; i < 2; i++) {
+		if (input.inputs[i].IsNull()) {
+			throw BinderException("ducklake_commit arguments cannot be NULL");
+		}
+	}
+	auto result = make_uniq<DuckLakeCommitBindData>();
+	result->metadata_schema_name = StringValue::Get(input.inputs[0]);
+	result->schema_version = input.inputs[1].GetValue<int64_t>();
+	result->retry_config.retry_wait_ms = 0;
```
dentiny: I'm not sure why do we reset retry wait -- we've already set default in the struct.
```cpp
struct DuckLakeRetryConfig {
	idx_t max_retry_count = 10;
	idx_t retry_wait_ms = 100;
	double retry_backoff = 1.5;

	static DuckLakeRetryConfig FromContext(ClientContext &context);
};
```
BTW for retry usually we also cap max retry wait ms (and jitter).

src/storage/ducklake_transaction_changes.cpp
```diff
@@ -112,8 +112,8 @@ ChangeInfo ParseChangeEntry(const string &changes_made, idx_t &pos) {
 	return info;
 }
 
-vector<ChangeInfo> ParseChangesList(const string &changes_made) {
-	vector<ChangeInfo> result;
+vector<ParsedChange> ParseChangesList(const string &changes_made) {
```
dentiny, line 117: most of the functions in this file should be put into anonymous namespace

PR #1192, opened 2026-05-27
Min max pushdown

guillesd:
Simplest part (1st commit) just exposes the min/max values to the optimizer via `DuckLakePartitionRowGroup`. This will allow the optimizer to do a bunch of optimizations (besides answering min/max queries from stats). This does not work in table with deletes.

The second part focuses on updating statistics for tables with deletes whenever a user runs `ducklake_rewrite_data_files`. Basically whenever a table is rewritten (and independently of whether it has inlined data or not), DuckLake recalculates the stats. This generally means that if you do maintenance operations regularly after doing UPDATES/DELETES in your lakehouse, stats will be up to date and therefore the optimizer will be able to do its job better. 

The complexity of the second part lies on adding inlined data stats, which right now is done via `TryMergeInlinedStats`, which queries the inlined tables for min/max values.

### DIFF
--- a/src/storage/ducklake_server_side_commit.cpp
+++ b/src/storage/ducklake_server_side_commit.cpp
@@ -612,10 +612,7 @@ vector<string> DuckLakeServerSideCommit::LookupInlinedTableNames(TableIndex tabl
 	vector<string> names;
 	auto sql =
 	    SubstitutePlaceholders(DuckLakeMetadataManager::GetInlinedTableNamesSql(table_id), transaction_snapshot);
-	auto result = fresh_conn.Query(sql);
-	if (!result || result->HasError()) {
-		return names;
-	}
+	auto result = RunQuery(sql, "lookup inlined table names");
 	for (auto &row : *result) {
 		names.push_back(row.GetValue<string>(0));
 	}
@@ -721,10 +718,7 @@ DuckLakeCommitContext DuckLakeServerSideCommit::BuildContext(idx_t &committed_sn
 		vector<DuckLakeColumnSchemaEntry> schema;
 		auto sql = SubstitutePlaceholders(DuckLakeMetadataManager::GetTableColumnSchemaSql(table_id),
 		                                  transaction_snapshot);
-		auto result = fresh_conn.Query(sql);
-		if (!result || result->HasError()) {
-			return schema;
-		}
+		auto result = RunQuery(sql, "read table column schema");
 		for (auto &row : *result) {
 			schema.push_back({FieldIndex(static_cast<idx_t>(row.GetValue<int64_t>(0))),
 			                  row.GetValue<string>(1), DuckLakeTypes::FromString(row.GetValue<string>(2))});
@@ -747,10 +741,7 @@ DuckLakeCommitContext DuckLakeServerSideCommit::BuildContext(idx_t &committed_sn
 		}
 		auto sql = SubstitutePlaceholders(
 		    DuckLakeMetadataManager::GetNetDataFileRowCountSql(table_id, inlined_deletion_table), transaction_snapshot);
-		auto result = fresh_conn.Query(sql);
-		if (!result || result->HasError()) {
-			return 0;
-		}
+		auto result = RunQuery(sql, "read net data file row count");
 		for (auto &row : *result) {
 			return row.GetValue<idx_t>(0);
 		}
@@ -761,10 +752,7 @@ DuckLakeCommitContext DuckLakeServerSideCommit::BuildContext(idx_t &committed_sn
 		for (auto &name : LookupInlinedTableNames(table_id)) {
 			auto sql =
 			    SubstitutePlaceholders(DuckLakeMetadataManager::GetNetInlinedRowCountSql(name), transaction_snapshot);
-			auto result = fresh_conn.Query(sql);
-			if (!result || result->HasError()) {
-				continue;
-			}
+			auto result = RunQuery(sql, "read net inlined row count");
 			for (auto &row : *result) {
 				total += row.GetValue<idx_t>(0);
 				break;
--- a/src/storage/ducklake_server_side_commit.cpp
+++ b/src/storage/ducklake_server_side_commit.cpp
@@ -720,8 +720,10 @@ DuckLakeCommitContext DuckLakeServerSideCommit::BuildContext(idx_t &committed_sn
 		                                  transaction_snapshot);
 		auto result = RunQuery(sql, "read table column schema");
 		for (auto &row : *result) {
-			schema.push_back({FieldIndex(static_cast<idx_t>(row.GetValue<int64_t>(0))),
-			                  row.GetValue<string>(1), DuckLakeTypes::FromString(row.GetValue<string>(2))});
+			// parent_column IS NULL => top-level root; otherwise a nested leaf carrying its own leaf type.
+			bool is_root = row.IsNull(3);
+			schema.push_back({FieldIndex(static_cast<idx_t>(row.GetValue<int64_t>(0))), row.GetValue<string>(1),
+			                  DuckLakeTypes::FromString(row.GetValue<string>(2)), is_root});
 		}
 		return schema;
 	};
--- a/src/storage/ducklake_server_side_commit.cpp
+++ b/src/storage/ducklake_server_side_commit.cpp
@@ -608,6 +608,20 @@ unique_ptr<DuckLakeStats> DuckLakeServerSideCommit::BuildStatsMap(vector<DuckLak
 	return result;
 }
 
+vector<string> DuckLakeServerSideCommit::LookupInlinedTableNames(TableIndex table_id) {
+	vector<string> names;
+	auto sql =
+	    SubstitutePlaceholders(DuckLakeMetadataManager::GetInlinedTableNamesSql(table_id), transaction_snapshot);
+	auto result = fresh_conn.Query(sql);
+	if (!result || result->HasError()) {
+		return names;
+	}
+	for (auto &row : *result) {
+		names.push_back(row.GetValue<string>(0));
+	}
+	return names;
+}
+
 const string &DuckLakeServerSideCommit::ResolveInlinedTableName(TableIndex table_id) {
 	auto it = inlined_table_name_cache.find(table_id.index);
 	if (it != inlined_table_name_cache.end()) {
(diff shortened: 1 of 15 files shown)


guillesd:
Okey so this won't work with strings. There is some limitation there (I did not dig much). But basically because we can't prove that the string column will be less than 8 bytes and since `StringStats` takes a truncated prefix of 8 bytes, we decide not to use stats and do a full scan. This can be worked on but I think it probably requires more than what's scoped for this PR.

pdet:
@Alex-Monahan, when you have time, could you review this PR?

Alex-Monahan:
I've started having a look and will continue over the next couple of days as I have time!

guillesd:
Linux test is failing because of a broken API in quay.io that Github is calling, unrelated to the changes

src/storage/ducklake_transaction.cpp
```diff
@@ -2555,6 +2569,215 @@ CompactionInformation DuckLakeTransaction::GetCompactionChanges(DuckLakeCommitSt
...
+		// no committed stats row exists yet - the UPDATE path of UpdateGlobalTableStats only refreshes existing rows
+		return;
+	}
+
+	DuckLakeTableStats new_stats;
+	new_stats.next_row_id = current_stats->next_row_id; // monotonic - carried forward, never recomputed
+	idx_t parquet_gross_rows = 0;
+
+	// 1. Merge the per-file stats of the post-rewrite parquet files = (pre-commit visible files - removed) + new files.
+	string query = StringUtil::Format(R"(
```
Alex-Monahan: Would it be better to have this query in ducklake_metadata_manager.cpp? I think it would allow other metadata catalogs to override this more easily. There are a few cases of queries not being centralized in that file, but maybe it would be best to try and keep things there when possible?

test/sql/stats/min_max_optimization_compaction.test
```diff
@@ -0,0 +1,171 @@
...
+# ============================================
+
+statement ok
+CREATE TABLE ducklake.t(i INTEGER);
+
+statement ok
+INSERT INTO ducklake.t FROM range(1, 201);
+
+statement ok
+CALL ducklake_flush_inlined_data('ducklake')
```
Alex-Monahan: Not needed, correct? Inlined limit is 100

test/sql/stats/min_max_optimization_time_travel.test
```diff
@@ -0,0 +1,97 @@
...
+
+# uncommitted insert that widens the range
+statement ok
+BEGIN;
+
+statement ok
+INSERT INTO ducklake.tt VALUES (-5), (9999);
+
+query II
+SELECT MIN(i), MAX(i) FROM ducklake.tt;
```
Alex-Monahan, line 61: Does this pull from metadata, or is there a scan here? Mind adding the query plan test in again?

jdctinuiti:
@guillesd I noticed one interaction with the full-file-drop/table-stats stack I’m working on in #1171/#1172/#1220.

If MIN/MAX pushdown trusts `ducklake_table_column_stats` when `ducklake_table_stats.record_count == net_count`, then full-file drops need to either recompute table-column stats or mark them unknown. Otherwise a whole-file drop can decrement `record_count` so the exactness check passes while min/max still include dropped files.

My current plan is to make the whole-file-drop paths update the table-column-stats contract safely: exact recompute where possible with conservative invalidation/unknown stats as fallback.

Do you prefer that I keep #1171/#1172/#1220 based on `main` and do a separate `v1.5-variegata` backport for the #1192 interaction, or would you rather see the stack rebased onto `v1.5-variegata`?

src/storage/ducklake_transaction.cpp
```diff
@@ -2555,6 +2569,215 @@ CompactionInformation DuckLakeTransaction::GetCompactionChanges(DuckLakeCommitSt
...
+			auto col = DuckLakeUtil::SQLIdentifierToString(field_id->Name());
+			bool is_float =
+			    field_id->Type().id() == LogicalTypeId::FLOAT || field_id->Type().id() == LogicalTypeId::DOUBLE;
+			string nan_expr =
+			    is_float ? StringUtil::Format("COALESCE(BOOL_OR(isnan(%s)), false)", col) : string("false");
+			select_list +=
+			    StringUtil::Format(", MIN(%s)::VARCHAR, MAX(%s)::VARCHAR, COUNT(%s), %s", col, col, col, nan_expr);
+		}
+		string query = StringUtil::Format(R"(
+SELECT %s
```
Alex-Monahan: Should this live in the ducklake_metadata_manager.cpp? Then we can keep the catalog SQL logic co-located and easy to override per-catalog class.

src/storage/ducklake_transaction.cpp
```diff
@@ -2555,6 +2569,215 @@ CompactionInformation DuckLakeTransaction::GetCompactionChanges(DuckLakeCommitSt
...
+LEFT JOIN {METADATA_CATALOG}.ducklake_file_column_stats stats ON stats.data_file_id = data.data_file_id
+WHERE data.table_id = %d
+  AND {SNAPSHOT_ID} >= data.begin_snapshot
+  AND ({SNAPSHOT_ID} < data.end_snapshot OR data.end_snapshot IS NULL)
+ORDER BY data.data_file_id;
+)",
+	                                  table_id.index);
+	auto result = Query(snapshot, query);
+	if (result->HasError()) {
+		return;
```
Alex-Monahan: Do we want to bubble up an error in some way?

src/storage/ducklake_transaction.cpp
```diff
@@ -2555,6 +2569,215 @@ CompactionInformation DuckLakeTransaction::GetCompactionChanges(DuckLakeCommitSt
...
+		}
+		string query = StringUtil::Format(R"(
+SELECT %s
+FROM {METADATA_CATALOG}.%s
+WHERE {SNAPSHOT_ID} >= begin_snapshot AND ({SNAPSHOT_ID} < end_snapshot OR end_snapshot IS NULL);
+)",
+		                                  select_list, DuckLakeUtil::SQLIdentifierToString(inlined_table.table_name));
+		auto result = Query(snapshot, query);
+		if (result->HasError()) {
+			return false;
```
Alex-Monahan: Anything we need to bubble up error-wise here?

test/sql/stats/min_max_optimization_basic.test
```diff
@@ -0,0 +1,129 @@
...
+
+# always-true filter over the whole range still folds
+query I
+SELECT MIN(i) FROM ducklake.test WHERE i >= -50;
+----
+-50
+
+# partial filter -> falls back to a scan
+query I
+SELECT MIN(i) FROM ducklake.test WHERE i > 50;
```
Alex-Monahan, line 91: Is it worth validating the query plan has a table scan here?

Alex-Monahan (approved):
Love this idea! My 2 main points of feedback are:
1. Should these SQL queries be kept in the ducklake_metadata_catalog.cpp file to keep things centralized and easy to override for other catalogs?
2. Can we add just a couple more validations of the query plan in a few spots?

Thanks!

src/storage/ducklake_transaction.cpp
```diff
@@ -2555,6 +2569,215 @@ CompactionInformation DuckLakeTransaction::GetCompactionChanges(DuckLakeCommitSt
...
+                                                           const CompactionInformation &rewrite_changes,
+                                                           const set<DataFileIndex> &removed_source_ids) {
+	auto snapshot = GetSnapshot();
+	auto entry = ducklake_catalog.GetEntryById(*this, snapshot, table_id);
+	if (!entry) {
+		return;
+	}
+	auto &table = entry->Cast<DuckLakeTableEntry>();
+	auto current_stats = ducklake_catalog.GetTableStats(*this, table_id);
+	if (!current_stats) {
```
Alex-Monahan: What would happen (and would current_stats exist) if we do this series of steps all in the same transaction (and is that permitted to be in the same transaction?)
1. BEGIN
2. Create a table
3. Insert data into it
4. Delete data
5. Compact deleted data 
6. COMMIT

test/sql/stats/min_max_optimization_compaction.test
guillesd: Claude doesn't seem to understand inlining! thanks for the nit!

src/storage/ducklake_transaction.cpp
guillesd: You are very right! I got lazy with this!

src/storage/ducklake_transaction.cpp
guillesd: I added a test that addresses that!

guillesd:
@Alex-Monahan thanks! I addressed your comments plus had to merge 1.5 since Pedro did a beefy refactor on DuckLake transactions (now going through DucKLakeTransactionState).

guillesd:
@Alex-Monahan FYI failing test is due to Quack using server side retries, which does not have visibility on table entries. I need to add a few things to make it work for Quack. Will be done in a couple of hours.

guillesd:
@Alex-Monahan it's ready to go! I made it work with the new Quack implementation too!

src/storage/ducklake_server_side_commit.cpp
```diff
@@ -703,6 +714,52 @@ DuckLakeCommitContext DuckLakeServerSideCommit::BuildContext(idx_t &committed_sn
...
+	};
+	ctx.get_inlined_table_names = [this](TableIndex table_id) {
+		return LookupInlinedTableNames(table_id);
+	};
+	ctx.get_net_data_file_row_count = [this](TableIndex table_id) -> idx_t {
+		// The inlined-file-deletion table is deterministically named but created lazily.
+		// Probe its existence; if absent, the SQL omits the inlined-deletion subterm.
+		auto inlined_deletion_table = DuckLakeMetadataManager::InlinedFileDeletionTableName(table_id);
+		auto probe_sql = SubstitutePlaceholders(
+		    StringUtil::Format("SELECT 1 FROM {METADATA_CATALOG}.%s LIMIT 0", inlined_deletion_table),
```
Alex-Monahan, line 738: This is somewhat of a nit, but do you think this should also be in the metadata class as well? This is the last remaining call to `Query` in the PR.

Alex-Monahan:
Thank you for addressing the feedback! I left one more comment - looks like one SQL query didn't get moved over to the metadata manager (the smallest query...). I have not thoroughly vetted the logic itself of when to use the min/max indexes (my C++ is not there just yet!), but otherwise looks ok to me!

src/storage/ducklake_server_side_commit.cpp
guillesd: Yes, this is true! The reason is that I actually want to be able to do `probe->HasError()` is to check whether the inlined table is there or not. This is a bit funky but it is a pattern that was there before (see metadata manager line 2967). We probably should think of a better way but let's keep it out of scope!

guillesd:
thanks @Alex-Monahan !

guillesd:
@pdet fixed the bug found by your digital intern. Basically recomputing stats was not working correctly for nested structs

pdet:
Thanks Gui!

PR #1193, opened 2026-05-27
[Experimental] Vortex file format support

jakthom-blw:
## Overview

  Adds support for Vortex data files in DuckLake.

  Parquet remains the default. New data files can now be written as either Parquet or Vortex, and DuckLake tracks the format per file so mixed-format tables continue to work.

  ## What changed

  - Added a `data_file_format` DuckLake option.
    - Accepted values: `parquet`, `vortex`.
    - Invalid values are rejected.
    - Uses existing option precedence: table, schema, DuckLake/global, then default `parquet`.

  - Added attach-time support for `DATA_FILE_FORMAT`.
    - `ATTACH ... (DATA_FILE_FORMAT 'vortex')` initializes the persisted global option for a new DuckLake.
    - Existing DuckLakes can change the persisted setting through `ducklake.set_option`.

  - Added per-data-file format tracking.
    - `DuckLakeDataFile` and related metadata structs now carry the data file format.
    - Metadata reads select `ducklake_data_file.file_format`.
    - Metadata writes persist the actual format instead of hardcoding `parquet`.
    - Existing/null metadata still defaults to Parquet.

  - Updated write paths to use the selected format.
    - INSERT
    - CTAS
    - MERGE INTO inserts
    - UPDATE rewrite inserts
    - flush inlined data
    - compaction

  - Updated COPY planning for format-specific behavior.
    - Parquet keeps existing field-id, encryption, compression, row group, version, and GeoParquet settings.
    - Vortex uses DuckDB’s `vortex` copy function.
    - Vortex files use the `.vortex` extension.
    - Parquet-only options are not passed to Vortex.
    - Vortex writes are rejected for encrypted DuckLakes for now.

  - Added Vortex read support.
    - Added `VortexFileScanner`, which binds DuckDB’s `read_vortex`.
    - Added `DuckLakeVortexFileReader`.
    - Integrated Vortex into DuckLake’s multi-file reader path.
    - Mixed Parquet/Vortex scans are supported through per-file reader selection.
    - Row IDs, casts, virtual columns, delete filtering, and scan filters are handled in the DuckLake reader path.

  - Fixed empty-projection Vortex scans.
    - Added an explicit `EMPTY` reader-column state so scans that do not project physical columns still read enough data to produce the correct row count.
    - This covers cases like count-style scans where the scan output has no regular data columns.

  - Updated metadata/file plumbing.
    - Multi-file list entries include `data_file_format` in extended file info.
    - Data file metadata queries now select data file format separately from delete file format.
    - Delete files remain unchanged: positional deletes are Parquet, deletion vectors are Puffin.

  - Updated cleanup.
    - Orphan file cleanup now considers both `.parquet` and `.vortex` files.

  - Added documentation.
    - `docs/vortex_backend_spec.md` describes the option surface, metadata behavior, write/read paths, compatibility notes, and current limitations.

  ## Limitations

  - Vortex data files do not currently store DuckLake field-id metadata through the DuckDB Vortex extension.
  - Vortex uses the existing missing-field-id fallback mapping path.
  - Straight append/read workloads and column renames are supported, but complex schema evolution across Vortex files is still experimental.
  - Encrypted Vortex data files are not supported yet.

  ## Testing

  Added `test/sql/settings/data_file_format.test`, covering:

  - attach-time `DATA_FILE_FORMAT 'vortex'`
  - persisted `data_file_format` metadata
  - writing and reading Vortex data files
  - `.vortex` file creation
  - `ducklake_data_file.file_format` tracking
  - switching a table back to Parquet
  - mixed Parquet/Vortex reads in the same table
  - invalid format rejection

### DIFF
--- a/docs/vortex_backend_spec.md
+++ b/docs/vortex_backend_spec.md
@@ -1,0 +1,80 @@
+# DuckLake Data File Backends
+
+## Goal
+
+DuckLake data files are selectable by format. Parquet remains the default for
+compatibility, and Vortex can be selected for new data files at DuckLake, schema,
+or table scope.
+
+## User Surface
+
+DuckLake adds a `data_file_format` option with two accepted values:
+
+- `parquet`
+- `vortex`
+
+The option follows the existing DuckLake option precedence:
+
+1. table option
+2. schema option
+3. global DuckLake option
+4. default `parquet`
+
+Examples:
+
+```sql
+CALL ducklake_set_option('ducklake', 'data_file_format', 'vortex');
+CALL ducklake_set_option('ducklake', 'data_file_format', 'parquet', table_name => 'orders');
+ATTACH 'ducklake:metadata.db' AS ducklake (DATA_PATH 'files', DATA_FILE_FORMAT 'vortex');
+```
+
+The selected format applies to newly written data files only. Existing files
+keep their original file format in metadata and remain readable. An attach-time
+`DATA_FILE_FORMAT` initializes the global DuckLake option when creating a new
+DuckLake; existing DuckLakes should use `ducklake_set_option` to change the
+persisted global default.
+
+## Metadata
+
+`ducklake_data_file.file_format` is the source of truth for each data file.
+Existing catalogs without a value, or with `parquet`, are interpreted as Parquet.
+New Vortex files store `vortex`.
+
+Delete files are unchanged. Positional delete files continue to be Parquet, and
+deletion-vector files continue to be Puffin.
+
+## Write Path
+
+Insert, CTAS, flush-inlined-data, merge, and compaction resolve the target data
+file format using `data_file_format`.
+
+For Parquet, all current writer behavior and Parquet-specific options remain
+unchanged.
+
+For Vortex:
+
+- DuckDB's `vortex` copy function is used.
+- Files use the `.vortex` extension.
+- Parquet-specific options are not passed.
+- Vortex writes are rejected for encrypted DuckLakes until the Vortex writer
+  exposes an equivalent encryption API.
+
+## Read Path
+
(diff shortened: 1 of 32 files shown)


jakthom-blw:
There's a `ducklake` <-> `duckdb submodule` <-> `vortex extension` version triangle prob - only after shimming one of those legs was I able to make this work.

pdet:
Hi @jakthom-blw, adding Vortex as a file format is a very interesting feature, but it is quite a Homeric task and most likely not a good fit for DuckLake 1.1, which is what `main` currently targets. You can see our contribution guidelines in [here](https://github.com/duckdb/ducklake#contributing).

Generally speaking, I would prefer first-time and early contributors to focus on bug fixes rather than features or performance improvements.

jakthom-blw:
Makes sense to me @pdet. Here's what I'm [seeing in the contribution guidelines](https://github.com/duckdb/ducklake#contributing):


# #

> We are happy to have outside contributions to our DuckLake extension. We currently have two active branches, and depending on your contribution, you must target the appropriate branch. Please note that bug fixes are always welcome. For new features, please coordinate with the team before starting any implementation.
> 
> 1. `main`. Our main branch is our DuckLake implementation that uses DuckDB main as a submodule. If your contribution is a new feature, or a bug fix that requires a DuckLake spec change, or if it strictly depends on a change in DuckDB main, you must target this branch.
> 2. `1.5-variegata`. Here, we use DuckDB 1.5-variegata as a submodule. Target this branch if you have a bug-fix that does not require a DuckLake spec change.


Since this isn't a `bugfix` I didn't pull to `1.5-variegata`.  Is there a preferred non-`main`, non-`bugfix` route? Or is "hurry up and wait" the right strategy here.

PR #1196, opened 2026-05-28
Fix remote execution of ducklake_delete_orphaned_files() for quack

### LINKED ISSUE #1169: Quack metadata manager remote-executes `read_blob` during `ducklake_delete_orphaned_files`
### What happens?

When using DuckLake with a Quack-backed metadata catalog, `ducklake_delete_orphaned_files(...)` appears to send the orphan-file discovery query to the remote Quack catalog backend.

That query contains:

```sql
FROM read_blob({DATA_PATH} || '**')
```

As a result, the remote Quack metadata service must be able to list/read the DuckLake data path, even though the DuckDB client process already has the filesystem/object-store credentials.

This differs from the Postgres metadata catalog path, where metadata reads fall back to the base metadata manager and execute in the local DuckDB process.

## Why this matters

For serverless Quack metadata backends, this creates an unexpected requirement:

- The DuckDB client has the `TYPE r2` / `TYPE s3` secret needed to read and delete DuckLake data files.
- `ducklake_delete_orphaned_files(...)` needs to discover physical files under `DATA_PATH`.
- With Quack, the discovery query is routed through `quack_query_by_name(...)`.
- Therefore the remote Quack backend must also have access to the same object store, or must reimplement file discovery.

This is surprising because the metadata catalog backend should ideally only store and query DuckLake metadata. For comparison, a Postgres metadata catalog does not need object-store credentials to support orphan cleanup.

## Code path observed

In `DuckLakeMetadataManager::GetOrphanFilesForCleanup(...)`, DuckLake builds a query containing:

```sql
FROM read_blob({DATA_PATH} || '**')
```

and executes it via:

```cpp
transaction.Query(query)
```

For the base metadata manager, `DuckLakeMetadataManager::Query(...)` substitutes placeholders and executes the SQL locally through:

```cpp
transaction.ExecuteRaw(query)
```

For Postgres, `PostgresMetadataManager::Query(...)` delegates to the base implementation:

```cpp
return DuckLakeMetadataManager::Query(snapshot, query);
```

So `read_blob(...)` runs locally in DuckDB, where the client-side filesystem extensions and secrets exist.

For Quack, `QuackMetadataManager::Query(...)` overrides this behavior and wraps the query in:

```sql
CALL system.main.quack_query_by_name(...)
```

As a result, `read_blob(...)` is executed by the remote Quack endpoint instead of the local DuckDB client.

## Expected behavior

`ducklake_delete_orphaned_files(...)` should discover files in the DuckDB client process, where the relevant filesystem extensions and secrets already exist.

The Quack metadata catalog should only receive the metadata queries/writes that truly belong to the metadata catalog.

This would match the behavior of the Postgres metadata manager more closely.

## Actual behavior

With a Quack metadata catalog, the remote Quack service receives a query involving `read_blob(...)`.

A remote backend that only implements metadata-catalog SQL cannot support `ducklake_delete_orphaned_files(...)` unless it also has access to the data files or implements a compatible file-listing layer.

## Impact

This would make Quack-backed DuckLake behavior align better with Postgres-backed DuckLake behavior.

It would also allow remote Quack metadata services to remain metadata-only catalog backends, without requiring direct access to the same R2/S3 buckets used by the DuckDB client.

## Reproduction shape

```sql
LOAD httpfs;
LOAD quack;
LOAD ducklake;

CREATE SECRET (
  TYPE quack,
  TOKEN '...'
);

CREATE SECRET lake_r2 (
  TYPE r2,
  KEY_ID '...',
  SECRET '...',
  ACCOUNT_ID '...',
  SCOPE 'r2://bucket/lake/'
);

ATTACH 'ducklake:quack:<host>:443' AS lake (
  DATA_PATH 'r2://bucket/lake/'
);

SELECT *
FROM ducklake_delete_orphaned_files(
  'lake',
  cleanup_all => true,
  dry_run => true
);
```

The remote Quack service receives a query containing `read_blob(...)`.

## Suggested direction

One possible fix would be similar in spirit to the fix for `ducklake_add_data_files(...)` in `duckdb/ducklake#1164`.

DuckLake could split orphan cleanup into:

1. Local DuckDB phase:
   - list files under `DATA_PATH`
   - apply file age filters
   - use the local filesystem/object-store credentials

2. Metadata catalog phase:
   - query registered data/delete/scheduled files from the metadata catalog
   - compare local physical files against registered metadata
   - remove catalog rows for scheduled cleanup where applicable

Alternatively, the Quack metadata manager could distinguish client-local helper queries involving filesystem functions from true metadata-catalog SQL and execute the former locally.

### OS:

MacOS latest

### DuckDB Version:

1.5.2

### DuckLake Version:

1.0

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

TobiLG

### Affiliation:

None

I have tested with a nightly build

- [ ] Yes, I have

- [ ] Yes, I have

tobilg:
## Summary

Fixes Quack-backed `ducklake_delete_orphaned_files()` so filesystem discovery runs in the attaching DuckDB client instead of on the remote Quack backend.

The base metadata manager keeps the existing single-query implementation used by local/SQLite/Postgres catalogs. Quack now overrides `GetOrphanFilesForCleanup()` and handles the mixed local/remote query shape explicitly:

- fetch known DuckLake data/delete/scheduled file paths through Quack's normal metadata query path
- materialize those known paths into a local temporary DuckDB table
- run `read_blob({DATA_PATH} || '**')` locally in the client
- use a local SQL anti-join against the temp table to return only orphaned files

This preserves the original `read_blob(...) NOT IN (...)` semantics while avoiding remote execution of the client data-path scan.

## Motivation

For Postgres and local catalogs, `ducklake_delete_orphaned_files()` is planned by the client DuckDB process: `read_blob(...)` scans the local data path and metadata tables are accessed as catalog scans.

Quack overrides metadata `Query(...)` by wrapping the full SQL string in `quack_query_by_name(...)`. As a result, the original orphan cleanup query sent the entire statement to the remote Quack backend, including:

```sql
read_blob({DATA_PATH} || '**')
```

That makes the remote backend scan for files that belong to the attaching client. This PR keeps Quack metadata access remote, but forces the physical file listing and orphan anti-join to happen locally.

## Tests

```sh
cmake --build build/release --config Release
make test
scripts/run_quack_tests.py --slow --no-install
```

Results:

- `make test`: all tests passed, 440 test cases, 94,500 assertions, 16 skipped
- `run_quack_tests.py --slow --no-install`: 392 passed, 0 failed, 0 timeouts, 0 infra errors

Fixes #1169.

## Disclaimer

This PR was created with the help of Codex / GPT-5.5 xhigh. Feel free to close it if this isn't acceptable.

### DIFF
--- a/src/include/metadata_manager/quack_metadata_manager.hpp
+++ b/src/include/metadata_manager/quack_metadata_manager.hpp
@@ -33,7 +33,6 @@ public:
 	unique_ptr<QueryResult> Query(string &query) override;
 	unique_ptr<QueryResult> AttachMetadata(const string &attach_query) override;
 	void ClearCache() override;
-	vector<DuckLakeFileForCleanup> GetOrphanFilesForCleanup(const string &filter, const string &separator) override;
 
 	bool MetadataExists() override;
 
--- a/src/include/metadata_manager/quack_metadata_manager.hpp
+++ b/src/include/metadata_manager/quack_metadata_manager.hpp
@@ -33,6 +33,7 @@ public:
 	unique_ptr<QueryResult> Query(string &query) override;
 	unique_ptr<QueryResult> AttachMetadata(const string &attach_query) override;
 	void ClearCache() override;
+	vector<DuckLakeFileForCleanup> GetOrphanFilesForCleanup(const string &filter, const string &separator) override;
 
 	bool MetadataExists() override;
 
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -401,6 +401,7 @@ public:
 protected:
 	string GetInlinedTableQuery(const DuckLakeTableInfo &table, const string &table_name);
 	string GetColumnType(const DuckLakeColumnInfo &col);
+	string GetKnownFilesForCleanupQuery(const string &separator) const;
 
 	//! Optimized data file writing using DuckDB Appender API (only for DuckDB metadata manager)
 	string WriteNewDataFilesWithAppender(DuckLakeSnapshot &commit_snapshot, const vector<DuckLakeFileInfo> &new_files,
--- a/src/metadata_manager/quack_metadata_manager.cpp
+++ b/src/metadata_manager/quack_metadata_manager.cpp
@@ -1,5 +1,9 @@
 #include "metadata_manager/quack_metadata_manager.hpp"
 #include "common/ducklake_util.hpp"
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/common/types/uuid.hpp"
+#include "duckdb/main/appender.hpp"
+#include "duckdb/main/client_context.hpp"
 #include "duckdb/main/connection.hpp"
 #include "storage/ducklake_catalog.hpp"
 #include "storage/ducklake_staged_commit.hpp"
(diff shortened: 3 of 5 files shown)


pdet:
Thanks for the PR.

Two things, 

1. I'm wondering if we need to annotate the temporary table with an UUID, since these are per client connection and they can hold only one query at a time, I think it should be safe?

2. We can move that solution to the ducklake_catalog instead of having a specialized implementation of it, I think it's a good solution in general for any server.

tobilg:
Thanks for the fast feedback @pdet, I'll have a look.

tobilg:
Updated the PR to generalize the orphan cleanup temp-table materialization.

What changed:

- Moved the known-file materialization flow into `DuckLakeMetadataManager::GetOrphanFilesForCleanup`.
- Kept `GetKnownFilesForCleanupQuery(separator)` as the shared helper for catalog-known data/delete/scheduled paths.
- The base cleanup now:
  - queries known paths through the metadata manager path,
  - creates `CREATE OR REPLACE TEMPORARY TABLE "__ducklake_known_cleanup_files"(full_path VARCHAR)`,
  - appends known paths through the internal metadata connection,
  - runs the local `read_blob({DATA_PATH} || '**')` scan with `NOT EXISTS` against the temp table,
  - best-effort drops the temp table on success and error.
- Removed the Quack-only `GetOrphanFilesForCleanup` override.
- Removed the Quack-only UUID/temp-table-name dependency.

I kept the temp table name fixed. The assumption is safe here because DuckDB temporary objects are connection-local, cleanup runs through the transaction-local metadata connection, and query execution on a client context is serialized.

Validation:

- `cmake --build build/release --config Release`
- Default DuckDB catalog: `make test`
  - Passed: 440 test cases, 94,500 assertions
- Focused remove-orphans SQL tests:
  - `test/sql/remove_orphans/remove_orphaned_files.test`
  - `test/sql/remove_orphans/mixed_paths.test`
  - `test/sql/remove_orphans/orphan_after_expire.test`
  - `test/sql/remove_orphans/metadata_in_data_path.test`
- Quack remove-orphans subset: `scripts/run_quack_tests.py --filter 'remove_orphans/*' --no-install`
  - Passed: 3/3
- Quack full slow suite: `scripts/run_quack_tests.py --slow --no-install`
  - Passed: 392/392
- SQLite catalog: `build/release/test/unittest --test-config test/configs/sqlite.json "test/*"`
  - Passed: 403 test cases, 93,638 assertions
- Postgres catalog: `build/release/test/unittest --test-config test/configs/postgres.json --test-dir ./ "test/sql/*"`
  - Passed: 427 test cases, 94,273 assertions
- MinIO/S3 config: `build/release/test/unittest --test-config /private/tmp/ducklake-minio-httpfs.json "test/sql/*"`
  - Passed: 430 test cases, 93,665 assertions
- `git diff --check`
  - Clean

Environment notes:

- Installed missing local `sqlite_scanner` and `postgres_scanner` DuckDB extensions before running those catalog suites.
- The first MinIO attempt failed because this local build needed `LOAD httpfs` before creating the S3 secret; rerunning with a temporary config that prepended `LOAD httpfs;` passed.

tobilg:
I honestly feel like this eventually should include more / more specific tests, as a bug could eventually be fatal regarding deleting files @pdet. Just my 2cts.

pdet:
@tobilg You have something specific in mind? I could imagine that indeed having tests with multiple connections and concurrent loops doing orphan deletion and insertion could be interesting (if those do not exist yet)

tobilg:
Yes, I think the most relevant missing coverage is a `concurrentloop` SQLLogicTest that runs regular inserts against a non-inlined table while other threads call `ducklake_delete_orphaned_files('ducklake')` without `cleanup_all`, i.e. using the normal age guard. The final assertions should verify that all committed rows remain readable and that a dry-run orphan scan reports zero orphan files.

I would avoid using `cleanup_all => true` in that concurrent-writer test unless we explicitly want to define/support that behavior, because it bypasses the age guard and can classify in-flight, not-yet-committed data files as orphaned. That seems more like a documented unsafe operational mode than a guarantee this PR should provide.

I updated my PR branch and added a test @pdet...

tobilg:
Any updates @pdet?

pdet:
I agree with (cc @guillesd)
> I would avoid using cleanup_all => true in that concurrent-writer test unless we explicitly want to define/support that behavior, because it bypasses the age guard and can classify in-flight, not-yet-committed data files as orphaned. That seems more like a documented unsafe operational mode than a guarantee this PR should provide.


LGTM, thanks for the work!

PR #1197, opened 2026-05-28
Allow Tables that are not inlined to use reserved names.

pdet:
Inlining requires certain reserved column names. (i.e., `row_id`, `begin_snapshot`, `end_snapshot`), we currently error if a user tries to create or modify a table that has a column with any of these names. This PR makes it more flexible. If inlining is disabled for a given table, that table can then use reserved names. This workes for inlining at any scope, either table, schema, or database.

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -27,20 +27,20 @@ jobs:
   duckdb-next-build:
     name: Build extension binaries
     needs: get-duckdb-version
-    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@main
+    uses: duckdb/extension-ci-tools/.github/workflows/_extension_distribution.yml@v1.5-variegata
     with:
       duckdb_version: ${{ needs.get-duckdb-version.outputs.duckdb_version }}
-      ci_tools_version: main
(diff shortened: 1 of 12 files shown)


suresh-summation:
@pdet  It looks like DuckDB 1.5.4 is expected to be released next week (June 15th). Will this fix be available in the 1.5.4 release?

pdet:
Hi @suresh-summation, indeed this fix will be available in 1.5.4

suresh-summation:
@pdet  Looks like the alter statement is failing even when data inlining is turned off. create statement is working as expected

```
CREATE OR REPLACE TABLE test(i INT);

ALTER TABLE test ADD COLUMN row_id int;

```

`SQL Error: Error while executing SQL "ALTER TABLE test ADD COLUMN row_id int": An execution error has occurred: Catalog Error: Column name "row_id" is reserved by DuckLake for internal use when data inlining is enabled. If you must use this column name, disable inlining by calling ducklake_set_option('data_inlining_row_limit', 0).`

pdet:
Good catch, I've made a PR for that. https://github.com/duckdb/ducklake/pull/1232

suresh-summation:
@pdet Thanks for the fix. I hope, this will also be part of 1.5.4 release.

PR #1200, opened 2026-05-29
fix: correct delete snapshot_id under virtual-column projection reordering (#1199)

### LINKED ISSUE #1199: table_changes / ducklake_table_deletions reports the wrong snapshot_id for deletes in real delete files, depending on virtual-column projection order
## Summary

When a data file accumulates deletes from **more than one snapshot into a single delete file**, `ducklake_table_deletions` returns the wrong `snapshot_id` for the rows whose real delete snapshot is *not* the delete file's `begin_snapshot`. Instead of the per-row delete snapshot (which is correctly stored on disk), the output column falls back to the delete file's `begin_snapshot` (the snapshot in which that data file was *first* deleted from).

The bug is triggered by the **projection order of the virtual columns**: it appears only when the `snapshot_id` virtual column is projected *before* the `rowid` virtual column. The `table_changes` macro is written as `SELECT snapshot_id, rowid, ...`, so it always hits the bug. The result is that an in-place `UPDATE` is sometimes reported as a clean `update_preimage`/`update_postimage` pair and sometimes split into a `delete` (attributed to the wrong snapshot) + `insert`.

This only manifests on the **real delete-file path** (`DATA_INLINING_ROW_LIMIT 0`); with default inlining the deletes go through `inlined_file_deletions`, which is not affected.

## Environment

- Reproduced on a local build of `main`: DuckDB `v1.5.0-dev8043` (`88277463aa`), ducklake at HEAD (commit `cb19190d`).
- Also reported against DuckDB 1.5.2 with a release ducklake build.
- Re-confirmed on the latest `v1.5-variegata` branch (ducklake `1a499a27`, DuckDB submodule `14eca11b` = `v1.5.3`): still reproduces unchanged on a clean `arm64-linux` release build — the fix has not landed upstream. The relevant code (`DuckLakeMultiFileReader::GetVirtualColumnExpression` setting `deletion_scan_snapshot_col = local_idx.GetIndex()`, consumed directly as the output-chunk column index in `GatherDeletionScanSnapshots`) is unchanged from the version where the bug was first found.
- File-based catalog, `ATTACH ... (DATA_INLINING_ROW_LIMIT 0)`.

## Minimal reproduction (pure DELETE, 3 rows)

```sql
ATTACH 'ducklake:m.ducklake' AS dl (DATA_PATH 'df', DATA_INLINING_ROW_LIMIT 0);
USE dl;
CREATE TABLE d (k INTEGER, v INTEGER);
INSERT INTO d VALUES (1,100),(2,200),(3,300);  -- snapshot 2  -> data_file 0, rows at pos 0,1,2
DELETE FROM d WHERE k=1;                         -- snapshot 3  -> deletes pos 0
DELETE FROM d WHERE k=2;                         -- snapshot 4  -> deletes pos 1 (same data file 0)
```

The output of `ducklake_table_deletions` is **deterministic** and depends only on the projection order:

| Projection                                   | Output            | Correct? |
|----------------------------------------------|-------------------|----------|
| `SELECT rowid, snapshot_id ...` (rowid first) | `r0@s3, r1@s4`    | ✅       |
| `SELECT snapshot_id, rowid ...` (snapshot first) | `s3@r0, s3@r1` | ❌ row 1 should be s4 |
| `SELECT snapshot_id ...` (no rowid at all)   | `s3, s4`          | ✅       |

```sql
-- WRONG (snapshot_id projected before rowid, as the table_changes macro does):
SELECT snapshot_id, rowid, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid;
-- snapshot_id | rowid | k
--     3       |   0   | 1
--     3       |   1   | 2     <-- should be snapshot 4

-- CORRECT (rowid projected before snapshot_id):
SELECT rowid, snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid;
-- rowid | snapshot_id | k
--   0   |     3       | 1
--   1   |     4       | 2
```

### Effect on `table_changes`

Because the macro projects `snapshot_id` before `rowid`, the same `UPDATE` is split:

```sql
CREATE TABLE c (k INTEGER, v INTEGER);
INSERT INTO c VALUES (1,100),(2,200),(3,300);  -- snap 2
UPDATE c SET v=111 WHERE k=1;                   -- snap 3
UPDATE c SET v=222 WHERE k=2;                   -- snap 4

SELECT snapshot_id, rowid, change_type, k, v
FROM ducklake_table_changes('dl','main','c',2,4) ORDER BY 1,2,3;
```

Observed (wrong):

```
snap rowid change_type        k  v
 3    0    update_preimage    1  100
 3    0    update_postimage   1  111
 3    1    delete             2  200     <-- k=2 delete attributed to snap 3
 4    1    insert             2  222     <-- k=2 update split off into snap 4
```

Expected: row 1 should be `update_preimage`@4 / `update_postimage`@4, the same as row 0.

## Storage is correct — only the output column is wrong

Dumping the metadata and the delete-file parquet for the repro above:

- `ducklake_delete_file` has a single row: `data_file_id=0`, `begin_snapshot=3`. The snapshot-4 delete was **accumulated into the same delete file** rather than creating a new record.
- The delete-file parquet carries the `_ducklake_internal_snapshot_id` column, and it is **correct per row**: `pos 0 -> snapshot 3`, `pos 1 -> snapshot 4`.

The per-row snapshot is also used correctly for *filtering*: `ducklake_table_deletions('dl','main','d',4,4)` returns only `pos 1`, and `(2,3)` returns only `pos 0`. So the only thing wrong is the value written into the output `snapshot_id` column.

## Root cause

Three pieces:

1. **The macro** — `src/functions/ducklake_table_changes.cpp:9-17` projects `SELECT snapshot_id, rowid, ...` (snapshot_id before rowid). This is the trigger.

2. **`src/storage/ducklake_multi_file_reader.cpp`**
   - `GetVirtualColumnExpression` (≈L522-585): the `snapshot_id` virtual column is initialized to a **constant expression** equal to the delete-scan entry's `snapshot_id` (≈L576-581). It also records `deletion_scan_rowid_col` / `deletion_scan_snapshot_col` as **local** virtual-column indices (`local_idx.GetIndex()`, ≈L527/L567).
   - `GatherDeletionScanSnapshots` (≈L587-640): is supposed to overwrite that constant per row — it reads each row's row_id from `chunk.data[rowid_col_idx]`, looks it up in `scan_snapshot_map`, and writes the real snapshot into `chunk.data[snapshot_col_idx]`.
   - `FinalizeChunk` (≈L642-686): in the **else branch** (≈L676-684, taken when the user explicitly projects `rowid`) it passes `output_chunk` to `GatherDeletionScanSnapshots`, which indexes it with the **local** virtual-column indices. When the virtual columns are reordered (snapshot before rowid), those indices no longer line up with the `output_chunk` layout, the lookup misses, the constant is never overwritten, and the output falls back to `begin_snapshot`.

3. **The fallback value** — `src/storage/ducklake_metadata_manager.cpp` (`GetTableDeletions`, ≈L1532/L1648): the delete-scan entry's scalar `snapshot_id = current_delete.begin_snapshot`. That is the wrong value that leaks through.

The three black-box results confirm the mechanism:
- `rowid` first → indices happen to align → correct.
- `snapshot_id` first → misalignment → overwrite no-ops → falls back to `begin_snapshot`.
- no `rowid` projected → `FinalizeChunk` takes the `internally_projected_rowid` branch (≈L649-675), which appends an internal rowid column and passes an explicit `rowid_col_override`, so it is immune to the reordering → correct.

## Why this explains the "sometimes preimage/postimage, sometimes delete+insert" behavior

The invariant is: **the emitted delete snapshot equals the `begin_snapshot` of the delete file that carries the row**. It is correct only when that equals the row's real delete snapshot.

- Single delete, or all deletes in one snapshot → the delete file's `begin_snapshot` equals the row's real delete snapshot → the constant is accidentally correct → clean preimage/postimage.
- A second row of the **same data file** deleted in a later snapshot accumulates into the same delete file (still `begin_snapshot = first delete`) → its real snapshot differs → wrong attribution → split into delete+insert.

So it is not run-to-run randomness; it is determined by whether the row's delete is co-located in a delete file whose `begin_snapshot` predates the row's own delete.

## Why existing tests don't catch it

`test/sql/table_changes/ducklake_table_deletions_partial_files.test` is named for partial delete files, but it does **not** set `DATA_INLINING_ROW_LIMIT 0`, so it actually exercises the inlined-deletion path (`ducklake_data_file` / `ducklake_delete_file` are empty in that test). It also happens to project `SELECT rowid, snapshot_id, ...` (rowid first), which avoids the misalignment. Adding `DATA_INLINING_ROW_LIMIT 0` and projecting `snapshot_id` before `rowid` reproduces the bug.

## Workaround (until fixed)

Hand-write the change feed so that **every reference to `ducklake_table_deletions` projects `rowid` before `snapshot_id`**, and materialize the scans:

```sql
WITH ins AS MATERIALIZED (
  SELECT rowid AS rid, snapshot_id AS sid, * FROM ducklake_table_insertions('dl','main','c',2,4)),
     del AS MATERIALIZED (
  SELECT rowid AS rid, snapshot_id AS sid, * FROM ducklake_table_deletions('dl','main','c',2,4))
SELECT sid AS snapshot_id, rid AS rowid,
       CASE WHEN (sid,rid) IN (SELECT sid,rid FROM del) THEN 'update_postimage' ELSE 'insert' END AS change_type, *
FROM ins
UNION ALL
SELECT sid, rid,
       CASE WHEN (sid,rid) IN (SELECT sid,rid FROM ins) THEN 'update_preimage' ELSE 'delete' END, *
FROM del;
```

This restores correct preimage/postimage pairing for the repro.

## Suggested fix

`GatherDeletionScanSnapshots` should not index `output_chunk` with local virtual-column indices. It should locate the rowid/snapshot columns in a way that is independent of projection order (e.g. the same explicit-override approach used in the `internally_projected_rowid` branch, applied to both branches). A regression test that uses `DATA_INLINING_ROW_LIMIT 0` and projects `snapshot_id` before `rowid` would lock this down.

Note: the LEFT JOIN rewrite of the macro in #869 keeps the `snapshot_id, rowid` column order, so it does not fix this issue.

KumoSiunaus:
## Summary

Fixes #1199.

`ducklake_table_deletions` — and therefore the `ducklake_table_changes` macro built on it — reported the **wrong per-row delete `snapshot_id`** on the real delete-file path (`DATA_INLINING_ROW_LIMIT 0`) whenever the `snapshot_id` virtual column was projected **before** `rowid`. The `table_changes` macro always projects `snapshot_id` first, so an in-place `UPDATE` whose rows were deleted across multiple snapshots into one accumulated delete file was mis-reported as `delete`@wrong-snapshot + `insert` instead of a matched `update_preimage`/`update_postimage` pair.

The per-row snapshot is stored correctly on disk (`_ducklake_internal_snapshot_id`) and used correctly for filtering — only the value written into the output `snapshot_id` column was wrong, falling back to the delete file's `begin_snapshot`.

## Root cause

`deletion_scan_rowid_col` / `deletion_scan_snapshot_col` were set in `GetVirtualColumnExpression` from the per-file **local** virtual-column index (`local_idx`). But `GatherDeletionScanSnapshots` uses them to index `output_chunk`, which is laid out in **output projection order**. The `snapshot_id` virtual column is emitted as a constant expression and never advances the local column counter (`local_idx = MultiFileLocalIndex(reader.column_ids.size())` in the column mapper), so the two index spaces diverge as soon as `snapshot_id` precedes `rowid`:

| Projection (delete scan)          | Before this PR        |
|-----------------------------------|-----------------------|
| `rowid, snapshot_id` (rowid 1st)  | ✅ correct (indices happen to align) |
| `snapshot_id, rowid` (snap 1st)   | ❌ falls back to `begin_snapshot` |
| `snapshot_id` only (no rowid)     | ✅ correct (internally-projected-rowid path) |

## Fix

Resolve the rowid/snapshot **output** positions from `global_column_ids` (which matches the `output_chunk` layout) in `CreateMapping`, mirroring the existing column-search in `InitializeReader`, and drop the stale `local_idx` assignments in `GetVirtualColumnExpression`. This makes both `FinalizeChunk` branches index `output_chunk` correctly regardless of projection order. No on-disk format, metadata, or filtering behavior changes.

## Tests

- New `test/sql/table_changes/ducklake_table_deletions_projection_order.test`: exercises rowid-first, snapshot-first, and no-rowid projections plus per-snapshot filtering on the real delete-file path, and a `table_changes` UPDATE scenario asserting matched preimage/postimage pairs with no spurious delete/insert split. Confirmed red before the fix, green after.
- Strengthened `test/sql/table_changes/ducklake_table_deletions_partial_files.test`: it was named for partial delete files but ran the inlined path with a rowid-first projection, so it never reached this code. Added `DATA_INLINING_ROW_LIMIT 0` and snapshot-first assertions over the multi-file accumulation. All prior assertions still pass (logical rowids are unchanged), confirming the real delete-file path produces identical rowid-first results.

The full `table_changes` group and the `update`/`insert`/`rowid`/`merge`/`delete`/`compaction` groups pass.

## Out of scope (separate, pre-existing)

While reviewing this I found an **unrelated, pre-existing** crash on the delete-file path: a query that **filters on a column not present in its SELECT projection** (e.g. `... ducklake_table_deletions(...) WHERE v > 0` where `v` is not projected) triggers DuckDB's filter-column-removal, which narrows `output_chunk` via `projection_ids`, and `GatherDeletionScanSnapshots` then indexes the wrong/out-of-range column → `INTERNAL Error`. This reproduces identically on unpatched code with both projection orders, is orthogonal to the projection-order bug fixed here, and a correct fix needs to map positions through `projection_ids` (not reachable in `FinalizeChunk` today). Left for a separate change to keep this PR focused on #1199.

### DIFF
--- a/test/sql/table_changes/ducklake_table_deletions_projection_order.test
+++ b/test/sql/table_changes/ducklake_table_deletions_projection_order.test
@@ -1,5 +1,5 @@
 # name: test/sql/table_changes/ducklake_table_deletions_projection_order.test
-# description: ducklake_table_deletions must report the correct per-row delete snapshot_id on the real delete-file path (DATA_INLINING_ROW_LIMIT 0) regardless of virtual-column projection order (issue #1199)
+# description: ducklake_table_deletions must report the correct per-row delete snapshot_id on the real delete-file path (DATA_INLINING_ROW_LIMIT 0) regardless of virtual-column projection order
 # group: [table_changes]
 
 require ducklake
@@ -33,29 +33,29 @@ DELETE FROM d WHERE k=1
 statement ok
 DELETE FROM d WHERE k=2
 
-# CORRECT (rowid before snapshot_id) - already worked before the fix, must stay correct
+# rowid projected before snapshot_id: each deleted row reports its own delete snapshot
 query III
 SELECT rowid, snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid
 ----
 0	3	1
 1	4	2
 
-# THE BUG: snapshot_id before rowid (as the table_changes macro projects).
-# Before the fix row 1 wrongly reports snapshot 3 (the delete file's begin_snapshot).
+# snapshot_id projected before rowid (the order the table_changes macro uses): each row must still
+# report its own delete snapshot, not the delete file's begin_snapshot (3)
 query III
 SELECT snapshot_id, rowid, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid
 ----
 3	0	1
 4	1	2
 
-# CORRECT (no rowid projected) - internally-projected-rowid branch, must stay correct
+# no rowid projected: exercises the internally-projected-rowid branch
 query II
 SELECT snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY snapshot_id
 ----
 3	1
 4	2
 
-# Per-snapshot filtering must keep working (guards against the fix breaking filtering)
+# Per-snapshot filtering must keep working
 query III
 SELECT snapshot_id, rowid, k FROM ducklake_table_deletions('dl','main','d',4,4) ORDER BY rowid
 ----
@@ -67,8 +67,8 @@ SELECT snapshot_id, rowid, k FROM ducklake_table_deletions('dl','main','d',2,3) 
 3	0	1
 
 # ---------------------------------------------------------------------------
-# table_changes: an in-place UPDATE must be a matched preimage/postimage pair
-# at the row's own snapshot, NOT split into delete@wrong-snapshot + insert.
+# table_changes: an in-place UPDATE must surface as a matched preimage/postimage
+# pair at the row's own snapshot.
 # ---------------------------------------------------------------------------
 
 # snapshot 5 - create second table
@@ -88,9 +88,8 @@ statement ok
 UPDATE c SET v=222 WHERE k=2
 
 # The range includes the snapshot-6 INSERT, so the three original inserts appear.
-# The key assertion is that BOTH updates are matched preimage/postimage pairs at
-# their own update snapshot. Before the fix, k=2 (rowid 1) was split into
-# delete@7 + insert@8 instead of an update_preimage/update_postimage@8 pair.
+# Both updates must surface as matched preimage/postimage pairs at their own
+# update snapshot (k=1 at snapshot 7, k=2 at snapshot 8).
 query IIIII
 SELECT snapshot_id, rowid, change_type, k, v
 FROM ducklake_table_changes('dl','main','c',6,8) ORDER BY rowid, change_type
@@ -104,7 +103,6 @@ FROM ducklake_table_changes('dl','main','c',6,8) ORDER BY rowid, change_type
 6	2	insert	3	300
 
 # An in-place UPDATE must never surface as a bare delete in this range.
-# A non-zero count here is the exact symptom of issue #1199 (delete@wrong-snapshot).
 query I
 SELECT count(*) FROM ducklake_table_changes('dl','main','c',6,8) WHERE change_type = 'delete'
 ----
--- a/test/sql/table_changes/ducklake_table_deletions_projection_order.test
+++ b/test/sql/table_changes/ducklake_table_deletions_projection_order.test
@@ -1,0 +1,111 @@
+# name: test/sql/table_changes/ducklake_table_deletions_projection_order.test
+# description: ducklake_table_deletions must report the correct per-row delete snapshot_id on the real delete-file path (DATA_INLINING_ROW_LIMIT 0) regardless of virtual-column projection order (issue #1199)
+# group: [table_changes]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+# DATA_INLINING_ROW_LIMIT 0 forces deletes onto the real delete-file path
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS dl (DATA_PATH '${DATA_PATH}/ducklake_projection_order', DATA_INLINING_ROW_LIMIT 0)
+
+statement ok
+USE dl
+
+# snapshot 1 - create table
+statement ok
+CREATE TABLE d (k INTEGER, v INTEGER)
+
+# snapshot 2 - insert three rows into one data file (positions 0,1,2)
+statement ok
+INSERT INTO d VALUES (1,100),(2,200),(3,300)
+
+# snapshot 3 - delete pos 0 (k=1)
+statement ok
+DELETE FROM d WHERE k=1
+
+# snapshot 4 - delete pos 1 (k=2); accumulates into the SAME delete file (begin_snapshot stays 3)
+statement ok
+DELETE FROM d WHERE k=2
+
+# CORRECT (rowid before snapshot_id) - already worked before the fix, must stay correct
+query III
+SELECT rowid, snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid
+----
+0	3	1
+1	4	2
+
+# THE BUG: snapshot_id before rowid (as the table_changes macro projects).
+# Before the fix row 1 wrongly reports snapshot 3 (the delete file's begin_snapshot).
+query III
+SELECT snapshot_id, rowid, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid
+----
+3	0	1
+4	1	2
+
+# CORRECT (no rowid projected) - internally-projected-rowid branch, must stay correct
+query II
+SELECT snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY snapshot_id
+----
+3	1
+4	2
+
+# Per-snapshot filtering must keep working (guards against the fix breaking filtering)
+query III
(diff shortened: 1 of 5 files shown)


dentiny:
Cc @dentiny

test/sql/table_changes/ducklake_table_deletions_projection_order.test
```diff
@@ -0,0 +1,111 @@
...
+DELETE FROM d WHERE k=2
+
+# CORRECT (rowid before snapshot_id) - already worked before the fix, must stay correct
+query III
+SELECT rowid, snapshot_id, k FROM ducklake_table_deletions('dl','main','d',2,4) ORDER BY rowid
+----
+0	3	1
+1	4	2
+
+# THE BUG: snapshot_id before rowid (as the table_changes macro projects).
```
dentiny: I'm not sure we should spend so many comments describing the current incorrect behavior and result.
For readers who don't have context, it doesn't provide information and confuses people -- they don't need to know what **WAS** wrong. I'd suggest cleanup these comments for both code change and two test files.

dentiny (approved):
Thanks!

test/sql/table_changes/ducklake_table_deletions_projection_order.test
KumoSiunaus: Got it. I just fixed the comments and also fixed and rebased #1206 .

dentiny:
Cc @pdet

dentiny:
CC @guillesd

guillesd:
Thanks!

guillesd:
@KumoSiunaus did you open an issue for the bug you encountered regarding projection_ids on delete code path?

KumoSiunaus:
> @KumoSiunaus did you open an issue for the bug you encountered regarding projection_ids on delete code path?

Yes, that is #1201 and the PR is #1206

PR #1206, opened 2026-05-29
fix: remap deletion-scan output columns under filter-column removal (#1201)

### LINKED ISSUE #1201: INTERNAL Error in ducklake_table_deletions/ducklake_table_changes when filtering on a column not in the SELECT list (real delete-file path)
## What happens

On the real (non-inlined) delete-file path — i.e. when the catalog is attached with `DATA_INLINING_ROW_LIMIT 0` — a query against `ducklake_table_deletions` (and therefore the `ducklake_table_changes` macro built on it) throws an `INTERNAL Error` and invalidates the database **whenever it filters on a column that is not in its SELECT projection**.

```
INTERNAL Error: Expected vector of type INT64, but found vector of type INT32
```

Depending on the projection it can instead surface as `Attempted to access index N within vector of size N`, or silently return wrong `rowid`/`snapshot_id` values.

## Reproduction

```sql
ATTACH 'ducklake:repro.ducklake' AS dl (DATA_PATH 'repro_data/', DATA_INLINING_ROW_LIMIT 0);
USE dl;

CREATE TABLE d (k INTEGER, v INTEGER);
INSERT INTO d VALUES (1,10),(2,20),(3,30),(4,40),(5,50);
DELETE FROM d WHERE k = 2;   -- snapshot 3
DELETE FROM d WHERE k = 4;   -- snapshot 4

-- OK: filter on a projected column
SELECT rowid, snapshot_id, k
FROM ducklake_table_deletions('dl','main','d',2,4)
WHERE k > 0;

-- CRASH: filter on column v, which is NOT in the SELECT list
SELECT rowid, snapshot_id, k
FROM ducklake_table_deletions('dl','main','d',2,4)
WHERE v > 0;
```

The first query returns the expected rows; the second throws the `INTERNAL Error` above. The crash does not depend on projection order (`rowid` first or `snapshot_id` first both fail), and it also happens when `rowid` is not projected at all (e.g. `SELECT snapshot_id, k ... WHERE v > 0`).

It does **not** reproduce under the default inlining settings — only on the real delete-file path (`DATA_INLINING_ROW_LIMIT 0`), which is the production storage form.

## Expected

`WHERE v > 0` should behave like `WHERE k > 0` (filtering on `v` without projecting it), returning the matching deleted rows with correct per-row `rowid`/`snapshot_id`.

## Analysis

`DuckLakeMultiFileReader::GatherDeletionScanSnapshots` writes the per-row `snapshot_id` into `output_chunk` using column positions that are tracked in the wrong index space. When DuckDB's filter-column-removal optimization applies, the filter-only column (`v`) is added to the scan's `global_column_ids` but removed from the output via `projection_ids`, so `output_chunk` is a narrowed/reordered subset of `global_column_ids`. The rowid/snapshot positions used to index `output_chunk` are not mapped through `projection_ids`, so they point at the wrong column (here an `INT32` data column instead of the `INT64` `snapshot_id`), causing the type-mismatch `INTERNAL Error`.

## Environment

- Re-confirmed on the latest upstream `v1.5-variegata` (HEAD `2c9f15d1`; bundled DuckDB `v1.5.3` / submodule `14eca11b`) with a fresh clean release build — still reproduces (`WHERE k > 0` returns correct rows; `WHERE v > 0` throws the `INTERNAL Error` above and invalidates the DB). Originally observed on `v1.5.0-dev8043`; also reproduces on `main`.
- This is pre-existing and **distinct** from #1199 (the delete `snapshot_id` projection-order bug): the two have different triggers — #1199 is about projection *order* (with no filter-column removal), whereas this fails specifically when filtering on an *unprojected* column — and are fixed at different sites. They are **not orthogonal/independent**, though: both originate in the same `deletion_scan_*_col` column-index addressing, and a correct fix for this issue **builds on #1199** — it assumes those indices have already been resolved to `global_column_ids` order (as #1199 does) and then additionally remaps them through `projection_ids`. This still reproduces on top of the #1199 fix, which confirms it is a separate defect rather than the same one.

KumoSiunaus:
Fixes #1201.

## Problem

On the real delete-file path (`DATA_INLINING_ROW_LIMIT 0`), `ducklake_table_deletions` / `ducklake_table_changes` crash with `INTERNAL Error: Expected vector of type INT64, but found vector of type INT32` (or an out-of-bounds index, or silently wrong `rowid`/`snapshot_id`) whenever a query filters on a column that is **not** in its SELECT projection — e.g. `WHERE v > 0` with `v` not projected.

DuckDB's filter-column-removal optimization keeps the filter-only column in the scan's `global_column_ids` but drops it from the output via `projection_ids`, so `output_chunk` is laid out in `projection_ids` order while `GatherDeletionScanSnapshots` addressed it with `global_column_ids`-order indices. When `projection_ids` is non-identity the two diverge and the per-row snapshot overwrite lands on the wrong column (here an `INT32` data column instead of the `INT64` `snapshot_id`).

## Fix

`FinalizeChunk` now remaps the rowid/snapshot column indices to their true `output_chunk` positions by matching expression pointers between `reader_data.expressions` and the per-file `executor` (new `RemapDeletionScanOutputColumn` helper). `projection_ids` is not reachable from `FinalizeChunk`, but `executor.expressions[k]` is the same `Expression*` as `reader_data.expressions[projection_ids[k]]`, so the output position is recoverable. The remap is a no-op when `projection_ids` is identity, so the common (unfiltered) path is unchanged.

The internally-projected-rowid branch (rowid not in the user's projection) additionally evaluates the internal rowid expression explicitly, because under filter-column removal that column is absent from `projection_ids` and the base executor leaves the appended temp column uninitialized.

## Stacked on #1200

This builds on #1200 (the #1199 fix): the remap assumes `deletion_scan_*_col` already hold `global_column_ids` output indices, which #1200 establishes. **#1200 should be merged first.** Until then this PR's diff includes #1200's commits — only the final commit (`fix: remap deletion-scan output columns under filter-column removal`) is the #1201 change. (See the issue for why the two bugs are distinct but layered.)

## Testing

- New regression test `test/sql/table_changes/ducklake_table_deletions_filter_unprojected.test`: filter on an unprojected column across projection orders, the no-rowid (internally-projected-rowid) path, multiple data files (per-file `row_id_start`), empty results, and `ducklake_table_changes`. Fails (crashes) before this fix, passes after.
- Added that test to the `deletion_vectors` config skip list — under deletion vectors (puffin bitmaps) per-position snapshot tracking is unavailable, so it fails there for the same inherent reason as the existing per-row-snapshot tests in that list (unrelated to this fix).
- Full ducklake suite green on the latest `v1.5-variegata` (15621 assertions / 435 cases, default config).


dentiny:
Cc @dentiny

pdet:

pdet:
Hi @KumoSiunaus the PR LGTM, one nit, can we reduce the comments to max one line, and if extremely necessary two? (including in tests).

KumoSiunaus:
> Hi @KumoSiunaus the PR LGTM, one nit, can we reduce the comments to max one line, and if extremely necessary two? (including in tests).

@pdet Done.

pdet:
Ahh shoot, sorry I merges sth that created a conflict for you, can you resolve this?

KumoSiunaus:
> Ahh shoot, sorry I merges sth that created a conflict for you, can you resolve this?

@pdet Done. The CI failure seems to has nothing to do with my code.

pdet:
Thanks! Looks great!

PR #1208, opened 2026-05-30
fix: prevent deleted rows reappearing when parquet and inlined tombstones overlap (#1084)

### LINKED ISSUE #1084: Deleted rows reappear in scans when parquet and inlined tombstones overlap on the same data file
## What happens?

DuckLake stores tombstones in two places: parquet delete files (for
batches of more than ~10 deletes per data file) and an
inlined-deletes catalog table (for smaller batches). If a single
data file ends up with the same row position recorded in both, the
read-time filter mishandles the duplicate, gets stuck on it, and
lets that row plus every other tombstoned row later in the same
vector chunk pass through to the caller. Worse, `SELECT COUNT(*)`
(computed from metadata stats) and `SELECT *` (which scans the
file) report different row counts for the same query because the
two paths account for the duplicate differently.

### Detailed summary

`DuckLakeMultiFileReader::InitializeReader`
(`src/storage/ducklake_multi_file_reader.cpp:259-274` at HEAD,
commit `202c966`) builds the per-file delete filter in two steps:

1. Loads the parquet delete file via
   `DuckLakeDeleteFilter::Initialize(context, delete_file)`. This
   populates `delete_data->deleted_rows` with sorted positions read
   from the parquet file and asserts strict monotonicity, so the
   parquet contribution is internally deduplicated.
2. If `file_entry.inlined_file_deletions` is non-empty, builds a
   `DuckLakeInlinedDataDeletes` and calls
   `DuckLakeDeleteFilter::Initialize(const DuckLakeInlinedDataDeletes &)`
   (`src/storage/ducklake_delete_filter.cpp:246-255`):

   ```cpp
   void DuckLakeDeleteFilter::Initialize(
           const DuckLakeInlinedDataDeletes &inlined_deletes) {
       D_ASSERT(std::is_sorted(delete_data->deleted_rows.begin(),
                               delete_data->deleted_rows.end()));
       auto mid_idx = delete_data->deleted_rows.size();
       for (auto &idx : inlined_deletes.rows) {
           delete_data->deleted_rows.push_back(idx);
       }
       delete_data->snapshot_ids.clear();
       std::inplace_merge(delete_data->deleted_rows.begin(),
                          delete_data->deleted_rows.begin() + mid_idx,
                          delete_data->deleted_rows.end());
   }
   ```

   This merges the two sorted ranges but does not deduplicate. If
   the inlined positions overlap with positions already in the
   parquet delete file, `deleted_rows` is left with duplicate values
   (e.g. `[10, 10, 100, 200]`).

Then `DuckLakeDeleteData::Filter`
(`src/storage/ducklake_delete_filter.cpp:53-85`) walks the chunk
position by position and only advances `delete_idx` on an exact
match:

```cpp
auto entry = std::lower_bound(deleted_rows.begin(), deleted_rows.end(),
                              start_row_index);
auto delete_idx = NumericCast<idx_t>(entry - deleted_rows.begin());
...
for (idx_t i = 0; i < count; i++) {
    if (delete_idx < deleted_rows.size() &&
        start_row_index + i == deleted_rows[delete_idx]) {
        bool is_deleted = true;
        ...
        delete_idx++;
        if (is_deleted) continue;
    }
    result_sel.set_index(result_count++, i);
}
```

If `deleted_rows[delete_idx]` is a duplicate value that the scan
has already moved past, `delete_idx` never advances again for the
rest of the chunk: every subsequent position is compared to the
duplicate value, never matches, and is treated as visible. So a
single duplicate in `deleted_rows` causes every later-tombstoned
position in that chunk to leak.

A second symptom: `SELECT COUNT(*)` is satisfied from metadata
stats (`record_count - sum(delete_count) - inlined_delete_count`),
which double-counts the overlapping positions in the opposite
direction. The stats-based count is too low; the scan count is too
high; the same logical query returns different row counts depending
on which path the planner picks.

### Real-world impact

We discovered this on a production table that had been live for
several years and had received a large bulk load early on plus many
smaller daily appends/deletes. Diagnostic state at the time we
investigated:

- 79 active data files, 6 active parquet delete files. Total
  physical rows 290,093, total parquet tombstones 230,411, total
  inlined-delete entries 342.
- Of those 342 inlined entries, 107 (all attached to one specific
  data file) point to positions already recorded in that file's
  parquet delete file.
- Visible row count via `SELECT COUNT(*)`: 70,307. Distinct id
  count: 59,447. Surviving primary-key duplicates: 10,860.
- Per-file scan visibility: every other big data file returns 0
  visible rows (correctly fully tombstoned). The one file with the
  107 overlapping positions leaks 10,860 of its 42,426 rows. That
  exactly accounts for the dup residue (`70,307 = 59,447 fresh +
  10,860 leaked`).

Three runs of the same `delete + insert` merge produced identical
10,860-row residues, confirming that the bug is deterministic and
data-shaped rather than racy.

### How does this happen in practice?

We were not able to identify the DuckLake code path that produced
the parquet/inlined overlap on the affected file. Hypotheses:

- An older DuckLake version (or a previous schema-evolution path)
  that wrote inlined entries which a later parquet-delete-file
  write did not clean up. The current
  `ducklake_inlined_delete_<table_id>` schema has no `end_snapshot`
  column, so once an entry is written it persists indefinitely.
- An `AUTOMATIC_MIGRATION`-time conversion that touched
  `ducklake_inlined_delete_*`.
- A manual catalog edit during incident response.
- An interaction we have not yet traced.

What is clear is that once the overlap exists, it cannot be cleaned
by the user-facing maintenance functions:
`ducklake_rewrite_data_files` skips data files that have any inlined
deletions (`src/functions/ducklake_compaction_functions.cpp:257-260`
short-circuits on `candidate.has_inlined_deletions` before
`delete_threshold` is consulted, so even
`delete_threshold => 0.0::DOUBLE` returns zero rows; this is a
separate bug https://github.com/duckdb/ducklake/issues/1085), and
`ducklake_merge_adjacent_files` skips files with delete files. So
even after we discovered the leak there was no in-product way to
fix the table; we ended up dropping and re-syncing it from the
upstream source.

### Proposed fix

We do not know the codebase well enough to be confident about the
right fix; the maintainers are in a better position to judge. Two
ideas we explored while reasoning about the bug, either of which
appears to restore correctness in our local repro:

1. **Deduplicate inside
   `Initialize(const DuckLakeInlinedDataDeletes &)`** so
   `deleted_rows` is a strictly increasing sequence (matching what
   `Initialize(context, delete_file)` already enforces for the
   parquet path):

   ```cpp
   delete_data->deleted_rows.erase(
       std::unique(delete_data->deleted_rows.begin(),
                   delete_data->deleted_rows.end()),
       delete_data->deleted_rows.end());
   ```

2. **Make `Filter` tolerant of duplicates** by advancing
   `delete_idx` past any values the scan has already moved past,
   before checking for a match:

   ```cpp
   for (idx_t i = 0; i < count; i++) {
       while (delete_idx < deleted_rows.size() &&
              deleted_rows[delete_idx] < start_row_index + i) {
           delete_idx++;
       }
       if (delete_idx < deleted_rows.size() &&
           start_row_index + i == deleted_rows[delete_idx]) {
           ...
       } else {
           result_sel.set_index(result_count++, i);
       }
   }
   ```

   This is more defensive and would also cover any other path that
   ends up with non-unique entries in `deleted_rows`.

We do not have strong opinions on which (if either) is the right
shape; there may be a better approach that fits the codebase or the
broader semantics that we are not aware of.

For the COUNT(*)-vs-scan inconsistency, one possibility is to align
both paths on whichever count is intended to be authoritative. If
the metadata stats are the source of truth, the scan would need to
deduplicate positions on read (which (1) above achieves). If the
scan is authoritative, the metadata-stats path could be updated to
subtract `distinct(parquet ∪ inlined)` rather than
`parquet_count + inlined_count`. Aligning the deduplication in
`Initialize` would make both paths agree as a side effect, which is
why we've been thinking about it as the more conservative option,
but again we'd defer to the maintainers' judgement.

## To Reproduce

The repro injects the duplicate inlined entries directly into the
postgres-backed catalog because we did not find the natural DuckLake
API path that produces the overlap. The read-time bug, however,
fires regardless of how the overlap is created.

Prerequisites: Python 3.11+ with `duckdb>=1.5.1,<1.6` and
`psycopg2-binary` installed; a postgres on `localhost:5433`
reachable as `user=postgres password=postgres dbname=postgres`. A
one-liner to spin one up with Docker:

```sh
docker run -d --name ducklake-pg -p 5433:5432 \
    -e POSTGRES_PASSWORD=postgres \
    -e POSTGRES_USER=postgres \
    -e POSTGRES_DB=postgres \
    postgres:16
```

Save the script below to a file and run it. It should print
`BUG REPRODUCED.` and exit 1.

```python
"""Minimal repro: parquet delete file + inlined deletions that
overlap on the same data file -> read-time scan leaks rows."""
from __future__ import annotations

import shutil
import sys
from pathlib import Path

import duckdb
import psycopg2

WS = Path("/tmp/ducklake-overlap-repro")
PG_DSN = "dbname=ducklake host=localhost port=5433 user=postgres password=postgres"


def reset_pg() -> None:
    """Drop and recreate the catalog database for a clean run."""
    admin = psycopg2.connect(
        host="localhost", port=5433, user="postgres",
        password="postgres", dbname="postgres",
    )
    admin.autocommit = True
    with admin.cursor() as c:
        c.execute('DROP DATABASE IF EXISTS ducklake WITH (FORCE)')
        c.execute('CREATE DATABASE ducklake')
    admin.close()


def reset_workspace() -> None:
    if WS.exists():
        shutil.rmtree(WS)
    WS.mkdir(parents=True)
    (WS / "data").mkdir()


def attach(con: duckdb.DuckDBPyConnection) -> None:
    con.execute("INSTALL postgres; LOAD postgres")
    con.execute(
        f"ATTACH 'ducklake:postgres:{PG_DSN}' AS dl "
        f"(DATA_PATH '{WS/'data/'}/')"
    )
    con.execute("USE dl")


def main() -> int:
    reset_pg()
    reset_workspace()

    # 1. Populate a 1000-row table.
    con = duckdb.connect()
    attach(con)
    con.execute("CREATE SCHEMA IF NOT EXISTS main")
    con.execute("CREATE TABLE main.t (id BIGINT)")
    con.execute("INSERT INTO main.t SELECT i FROM range(0, 1000) AS r(i)")
    con.close()

    # 2. Issue a >10-row DELETE. This crosses
    # ducklake_default_data_inlining_row_limit (default 10), so the
    # tombstones go to a parquet delete file rather than the
    # inlined-deletes table.
    con = duckdb.connect()
    attach(con)
    con.execute("DELETE FROM main.t WHERE id < 800")
    con.close()

    # 3. Sanity-check the pre-injection state: 200 rows visible, ids
    # in [800, 999], COUNT(*) and scan agree.
    con = duckdb.connect()
    attach(con)
    print("STATE 1: parquet delete file only, no inlined")
    state1_count = con.execute("SELECT COUNT(*) FROM main.t").fetchone()[0]
    state1_ids = con.execute(
        "SELECT id FROM main.t ORDER BY id"
    ).fetchdf()["id"]
    print(f"  COUNT(*) = {state1_count}")
    print(
        f"  scan rows = {len(state1_ids)}  "
        f"ids range [{state1_ids.min()}..{state1_ids.max()}]"
    )
    assert state1_count == 200, f"pre-injection broken: {state1_count}"
    con.close()

    # 4. Inject 100 inlined-delete entries that point to positions
    # already in the parquet delete file. We use a direct catalog
    # write because we did not find a public DuckLake API that
    # produces this overlap; the read-time bug, however, fires
    # regardless of how the overlap is created.
    admin = psycopg2.connect(
        host="localhost", port=5433, user="postgres",
        password="postgres", dbname="ducklake",
    )
    admin.autocommit = True
    with admin.cursor() as c:
        c.execute("SELECT MAX(snapshot_id) FROM ducklake_snapshot")
        cur_snap = c.fetchone()[0]
        c.execute(
            "SELECT data_file_id FROM ducklake_data_file "
            "WHERE table_id = 1 AND end_snapshot IS NULL"
        )
        data_file_id = c.fetchone()[0]
        c.execute(
            "CREATE TABLE IF NOT EXISTS ducklake_inlined_delete_1 "
            "(file_id BIGINT, row_id BIGINT, begin_snapshot BIGINT)"
        )
        # Positions 0, 8, 16, ..., 792: 100 positions, all already in
        # the parquet delete file (which covers 0..799).
        for pos in range(0, 800, 8):
            c.execute(
                "INSERT INTO ducklake_inlined_delete_1 VALUES (%s, %s, %s)",
                (data_file_id, pos, cur_snap),
            )
    admin.close()

    # 5. Read again. Adding tombstones for positions that are already
    # tombstoned should be a no-op: COUNT(*) should still be 200 and
    # the scan should still return ids in [800, 999].
    con = duckdb.connect()
    attach(con)
    print("\nSTATE 2: parquet + 100 overlapping inlined entries")
    state2_count = con.execute("SELECT COUNT(*) FROM main.t").fetchone()[0]
    state2_ids = con.execute(
        "SELECT id FROM main.t ORDER BY id"
    ).fetchdf()["id"]
    print(f"  COUNT(*) = {state2_count}")
    print(
        f"  scan rows = {len(state2_ids)}  "
        f"ids range [{state2_ids.min()}..{state2_ids.max()}]"
    )
    print(f"  first 10 ids: {state2_ids.head(10).tolist()}")
    print(f"  last 10 ids:  {state2_ids.tail(10).tolist()}")
    print(
        "\nExpected (in both states): "
        "COUNT(*) = 200, scan rows = 200, ids in [800, 999]."
    )

    bug_count = state2_count != 200
    bug_scan = len(state2_ids) != 200 or state2_ids.min() < 800

    if bug_count or bug_scan:
        print("\nBUG REPRODUCED.")
        return 1
    print("\nNo bug observed.")
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

Output observed:

```
STATE 1: parquet delete file only, no inlined
  COUNT(*) = 200
  scan rows = 200  ids range [800..999]

STATE 2: parquet + 100 overlapping inlined entries
  COUNT(*) = 100
  scan rows = 999  ids range [1..999]
  first 10 ids: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
  last 10 ids:  [990, 991, 992, 993, 994, 995, 996, 997, 998, 999]

Expected (in both states): COUNT(*) = 200, scan rows = 200, ids in [800, 999].

BUG REPRODUCED.
```

Expected output: in both states, `COUNT(*) = 200` and the scan
returns 200 rows with ids in `[800, 999]`. Adding inlined-delete
entries for positions that are already tombstoned in the parquet
delete file should be a no-op.

### OS:

macOS / Darwin 25.4.0 on aarch64 (Apple Silicon)

### DuckDB Version:

1.5.2

### DuckLake Version:

ducklake extension version 415a9ebd (the build shipped with DuckDB 1.5.2) also reproduced against duckdb/ducklake @ commit 202c966

### DuckDB Client:

Python (duckdb 1.5.2 wheel from PyPI)

### Hardware:

_No response_

### Full Name:

Giordano Mattoni

### Affiliation:

Tensor Energy

and also core_nightly (extension https://github.com/duckdb/ducklake/commit/4c841e7625649342c374b43b710686c97a86159d)

Yes

- [x] Yes, I have

- [x] Yes, I have

GavinPizza:
Closes #1084

## What happens
When a single data file has the same row position recorded in both a parquet delete file and the inlined-delete catalog table (`ducklake_inlined_delete_<table_id>`), deleted rows reappear in scans.

`DuckLakeDeleteFilter::Initialize(const DuckLakeInlinedDataDeletes &)` merges the two sorted position ranges in one of two paths (the `std::inplace_merge` fast path, or a snapshot-tracking merge), but does not deduplicate, so `deleted_rows` can end up with a repeated value (e.g. `[10, 10, 11, ...]`). `DuckLakeDeleteData::Filter` advances its cursor only on an exact match, so once the scan passes the duplicated value the cursor gets stuck on it: every later tombstoned position in the chunk is compared against the stale value, never matches, and leaks through to the caller.

## The fix
Deduplicate `deleted_rows` after the merge (in both paths) so it stays strictly increasing, the same invariant the parquet delete path already enforces (it asserts strictly-increasing positions on read). This restores the precondition `Filter` relies on.

```cpp
delete_data->deleted_rows.erase(
    std::unique(delete_data->deleted_rows.begin(), delete_data->deleted_rows.end()),
    delete_data->deleted_rows.end());
```

## Test
Adds `test/sql/issues/issue_1084_parquet_inlined_delete_overlap.test`, which reproduces the overlap through ordinary SQL, with no catalog editing. Two concurrent transactions delete the same row of the same parquet data file with delete sizes that straddle `data_inlining_row_limit` (default 10): the single-row delete is stored as an inlined file delete, while the larger delete is stored as a parquet delete file. Both commit, so the row's position ends up recorded in both stores. The test asserts the overlap is actually present in the catalog and that the deleted positions are not visible to a scan. Verified the test fails without the fix and passes with it.

## Scope / follow-up
This fixes the scan correctness bug (deleted rows leaking into results). It does not address the related symptom that stats-based `SELECT COUNT(*)` double-subtracts the overlapping positions: that path sums `delete_count + inlined_delete_count` from the catalog and cannot dedupe, because the parquet delete positions live in the parquet file, not the catalog. Reconciling `COUNT(*)` needs separate write-path or stats-semantics work. The write-path origin of these overlaps, namely concurrent inlined and parquet deletes of the same row not being detected as a conflict, is tracked separately in #1215; this PR fixes the read-time symptom regardless of how the overlap arises. Calling it out here rather than silently implying both paths now agree.


dentiny:
CC @dentiny

test/sql/issues/issue_1084_parquet_inlined_delete_overlap.test
```diff
@@ -0,0 +1,85 @@
...
+0	100
+
+query I
+SELECT delete_count FROM ducklake_meta.ducklake_delete_file
+----
+30
+
+# inject inlined tombstones that OVERLAP positions already in the parquet delete file.
+# this overlap cannot be produced through normal SQL (a deleted row cannot be re-deleted),
+# so we write it directly into the catalog to mirror the corrupted state from issue #1084.
```
dentiny: Emmm I'm not sure this is a good repro -- if it's a real problem can we use user-facing SQL, instead of hacking catalog table?

GavinPizza: Yep I should have recreated the problem with user-facing SQL. I hit this in a concurrent workload but had a brain fart and reduced the regression test to a single transaction, which is where the wrong "can't happen in plain SQL" framing came from. I've rewritten the test around the concurrent path (no catalog editing) and updated the PR description.

The path: `data_inlining_row_limit` (default 10) routes a small DELETE to inlined file deletes and a large one to a parquet delete file. Two concurrent transactions deleting the same row, one small and one large, leave that position recorded in both stores and both commit. The new test does this (`con1` deletes one row, `con2` deletes a 30-row range covering it) and checks the deleted rows stay hidden. Verified it leaks pre-fix and passes post-fix.

I also created #1215 for the underlying behavior, since I believe those two deletes arguably should conflict (parquet+parquet and inlined+inlined already do).

chaegordon:
Thanks for this fix — we hit the bug in production and the dedup is exactly right for the scan path. We posted detailed findings on the issue ([here](https://github.com/duckdb/ducklake/issues/1084#issuecomment-4898024082) and [here](https://github.com/duckdb/ducklake/issues/1084#issuecomment-4900759112)) that suggest three extra assertions this PR's regression test could pick up cheaply, since the overlapped state is already constructed:

1. `count(*)` agrees with a forced scan — the metadata/stats path double-counts the overlapped position and will still return N−1 after this fix (each of the two delete transactions decremented stats).
2. A **subsequent** `DELETE` against positions after the overlap actually takes effect on scans — pre-fix, the stalled cursor swallows future tombstones too, making the file delete-proof.
3. An `AT (VERSION => …)` read at/after the overlap snapshot returns the same results as the equivalent live read did — time-travel uses the same filter and leaks identically pre-fix.

GavinPizza:
Based on @chaegordon comment I added 2 and 3, both pass.

1 is real (count(*) says 69, a scan finds 70) but the cause isn't the stats table, record_count never gets decremented. It's the count(*) pushdown: sum(record_count) - sum(delete_count) - count(inlined deletes), which subtracts the overlapped position twice. Time travel skips the shortcut and counts fine. The catalog doesn't know which positions a parquet delete file covers, so this isn't fixable from the scan path. Filed as #1329. (#1285 prevents new overlaps from committing, so only already-overlapped catalogs hit this.)

GavinPizza:
@pdet Merge conflicts are resolved

PR #1210, opened 2026-05-30
Evict DuckLake schema cache on snapshot expiration

jzillmann:
## Summary

This fixes part of the cache growth described in #852 by evicting DuckLake schema ObjectCache entries that become unreachable after `ducklake_expire_snapshots` removes snapshots.

- retarget to `v1.5-variegata`
- reuse the existing `DuckLakeCatalog::InvalidateSchemaCache` helper
- invalidate each expired snapshot `schema_version` after snapshot deletion succeeds
- add a small regression test that verifies dry-run expiration leaves cache memory intact and real expiration reduces ObjectCache memory

## Tests

- `cmake --build build/reldebug --target ducklake_loadable_extension unittest --parallel 8`
- `build/reldebug/test/unittest test/sql/issues/issue_852_schema_cache_growth.test --durations yes`
- `build/reldebug/test/unittest test/sql/compaction/expire_snapshots.test --durations yes`
- `build/reldebug/test/unittest test/sql/compaction/expire_snapshots_metadata_cleanup.test --durations yes`
- `build/reldebug/test/unittest test/sql/compaction/expire_snapshots_invalid_params.test --durations yes`
- `git diff --check`

### DIFF
--- a/src/functions/ducklake_expire_snapshots.cpp
+++ b/src/functions/ducklake_expire_snapshots.cpp
@@ -111,6 +111,10 @@ void DuckLakeExpireSnapshotsExecute(ClientContext &context, TableFunctionInput &
 	if (!state.executed && !data.dry_run) {
 		auto &transaction = DuckLakeTransaction::Get(context, data.catalog);
 		transaction.DeleteSnapshots(data.snapshots);
+		auto &ducklake_catalog = data.catalog.Cast<DuckLakeCatalog>();
+		for (auto &snapshot : data.snapshots) {
+			ducklake_catalog.InvalidateSchemaCache(snapshot.schema_version);
+		}
 		state.executed = true;
 	}
 
--- a/test/sql/issues/issue_852_schema_cache_growth.test
+++ b/test/sql/issues/issue_852_schema_cache_growth.test
@@ -1,0 +1,82 @@
+# name: test/sql/issues/issue_852_schema_cache_growth.test
+# description: https://github.com/duckdb/ducklake/issues/852 - DuckLakeCatalog::GetSchemaCacheEntry inserts into the ObjectCache keyed by schema_version. On a long-lived connection, repeated DDL grows the cache, and ducklake_expire_snapshots should release entries for expired schema versions.
+# group: [issues]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION {TEST_DIR}/{UUID}.db
+
+test-env DATA_PATH {TEST_DIR}
+
+statement ok
+ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/issue_852_schema_cache_growth')
+
+statement ok
+CREATE TABLE ducklake.anchor (id INTEGER)
+
+# Force one schema cache entry to be materialized so baseline > 0
+statement ok
+SELECT * FROM ducklake.anchor LIMIT 0
+
+# Baseline: snapshot ObjectCache memory after the initial schema load.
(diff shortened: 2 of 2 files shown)


dentiny:
CC @dentiny

src/functions/ducklake_expire_snapshots.cpp
```diff
@@ -111,6 +112,19 @@ void DuckLakeExpireSnapshotsExecute(ClientContext &context, TableFunctionInput &
 	if (!state.executed && !data.dry_run) {
 		auto &transaction = DuckLakeTransaction::Get(context, data.catalog);
 		transaction.DeleteSnapshots(data.snapshots);
+		auto &ducklake_catalog = data.catalog.Cast<DuckLakeCatalog>();
+		unordered_set<idx_t> schema_versions;
```
dentiny:
```suggestion
schema_versions.reserve(data.snapshots.size());
next_file_ids.reserve(data.snapshots.size());
```

dentiny: But what's the blocker to invalidate cache entries directly, so we don't need both unordered_sets here?

dentiny (commented):
overall looks good to me, but I guess you need to re-target to v1.5.
Recently I've split stats cache to per-table: https://github.com/duckdb/ducklake/pull/1167

src/functions/ducklake_expire_snapshots.cpp
```diff
@@ -111,6 +111,10 @@ void DuckLakeExpireSnapshotsExecute(ClientContext &context, TableFunctionInput &
 	if (!state.executed && !data.dry_run) {
 		auto &transaction = DuckLakeTransaction::Get(context, data.catalog);
 		transaction.DeleteSnapshots(data.snapshots);
+		auto &ducklake_catalog = data.catalog.Cast<DuckLakeCatalog>();
```
dentiny, line 114: I'm wondering why don't we handle stats cache entries this time? I'm ok though to consider it as a followup.

test/sql/issues/issue_852_schema_cache_growth.test_slow
```diff
@@ -0,0 +1,82 @@
+# name: test/sql/issues/issue_852_schema_cache_growth.test_slow
```
dentiny: Curious why do we need so many iterations, which makes it a slow test? To check the effect of eviction, comparison between before/after snapshot eviction should be enough?

jzillmann: Updated in `4aad92dc`: the regression is now a regular `.test`, the churn is down from 12 CREATE/DROP pairs to 3, and the assertion no longer depends on a coarse `> 40000` threshold. It now just checks that churn grows ObjectCache memory, dry-run expiration does not shrink it, and real expiration does. Locally this runs in ~0.11s.

src/functions/ducklake_expire_snapshots.cpp
jzillmann: I left stats cache eviction out of this PR and narrowed the title/body to schema cache only. On `v1.5-variegata` the stats cache is per-table and keyed by `(next_file_id, table_id)`, while `ducklake_expire_snapshots` currently carries `DuckLakeSnapshotInfo` with the expired `schema_version` only. Evicting stats exactly would require adding the missing key data and table-id enumeration/loading for expired snapshots, so I would rather keep this PR to the direct schema-cache invalidation and handle stats in a follow-up.

dentiny (approved):
Thanks!
CC @pdet to trigger CI and have a final look.

pdet:
Thanks!

PR #1213, opened 2026-05-31
Evict stats cache entries on snapshot expiration

dentiny:
Counterpart for https://github.com/duckdb/ducklake/pull/1210

This PR evicts useless stats cache entries when snapshot is expired explicitly.


pdet:
Thanks!

Issue #1212, opened 2026-05-31
Table defaults appear broken


billziss-gh:
### What happens?

Table defaults appear broken as the value `"NULL"` is special-cased in [`DuckLakeMetadataManager::GetCatalogForSnapshot`](https://github.com/duckdb/ducklake/blob/v1.5-variegata/src/storage/ducklake_metadata_manager.cpp#L672-L679):

```cpp
		if (!row.IsNull(COLUMN_INDEX_START + 4)) {
			auto value = row.GetValue<string>(COLUMN_INDEX_START + 4);
			if (value == "NULL") {
				column_info.default_value = Value();
			} else {
				column_info.default_value = Value(value);
			}
		}
```

### To Reproduce

```
% duckdb ducklake:test.ducklake
DuckDB v1.5.2 (Variegata)
Enter ".help" for usage hints.
test D CREATE TABLE t (id INTEGER, t0 VARCHAR, t1 VARCHAR DEFAULT 'NULL');
test D INSERT INTO t (id) VALUES (1);
test D from t;
┌───────┬─────────┬─────────┐
│  id   │   t0    │   t1    │
│ int32 │ varchar │ varchar │
├───────┼─────────┼─────────┤
│     1 │ NULL    │ NULL    │
└───────┴─────────┴─────────┘
test D from t where t0 is null;
┌───────┬─────────┬─────────┐
│  id   │   t0    │   t1    │
│ int32 │ varchar │ varchar │
├───────┼─────────┼─────────┤
│     1 │ NULL    │ NULL    │
└───────┴─────────┴─────────┘
test D from t where t1 is null;
┌───────┬─────────┬─────────┐
│  id   │   t0    │   t1    │
│ int32 │ varchar │ varchar │
├───────┼─────────┼─────────┤
│     1 │ NULL    │ NULL    │
└───────┴─────────┴─────────┘
```

### OS:

macOS Tahoe

### DuckDB Version:

v1.5.2 (Variegata)

### DuckLake Version:

415a9ebd

### DuckDB Client:

CLI

### Full Name:

Bill Zissimopoulos

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

[...]

PR #1214, opened 2026-05-31
Fix NULL interpret

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1212

The issue description has already pointed out the root cause, but I found read side doesn't have enough information to tell apart between `'NULL'` and `NULL`, the easiest fix might be making column type an expression instead of literal.


pdet:
I'm wondering how this will affect a system if a table with the default value was created prior to the fix and if we will need a migration for that.

pdet:

pdet:

PR #1217, opened 2026-06-03
Evict name map cache entries on snapshot expiration

dentiny:
Followup PR for https://github.com/duckdb/ducklake/pull/1210 and https://github.com/duckdb/ducklake/pull/1213

When a table is dropped and its snapshot expired, we delete related rows for `ducklake_column_mapping` and `ducklake_name_mapping`
https://github.com/duckdb/ducklake/blob/d897bc5a35f887c9087403bc20efd92d99272e69/src/storage/ducklake_metadata_manager.cpp#L4499-L4504

There's no reason we keep their in-memory cache entries, which leads to memleak.


pdet:

dentiny:
> Hi, thanks for opening the PR, we are now moving all PRs from 1.5-variegata to main. Could you please retarget?

Sure! This PR is fairly simple, maybe @guillesd could take a look when you have some time? Thanks!

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -5245,9 +5245,25 @@ WHERE NOT EXISTS (
...
+FROM {METADATA_CATALOG}.ducklake_name_mapping tbl
+WHERE NOT EXISTS (
+    SELECT 1 FROM {METADATA_CATALOG}.ducklake_column_mapping m
+    WHERE m.mapping_id = tbl.mapping_id
+);)");
+		if (result->HasError()) {
+			result->GetErrorObject().Throw("Failed to list deleted name mappings in DuckLake: ");
+		}
+		for (auto &row : *result) {
+			catalog.InvalidateNameMapCache(MappingIndex(row.GetValue<idx_t>(0)));
```
pdet: What happens in a rollback? Can we add a test that verifies the data is properly re-added to the cache?
 e.g.,
```
  ATTACH 'ducklake:a.db' AS ducklake;
  CREATE TABLE ducklake.to_drop (col_0 INTEGER, col_1 VARCHAR);
  COPY (SELECT 42 col_0, 'duck' col_1) TO '/tmp/x/file.parquet';
  CALL ducklake_add_data_files('ducklake', 'to_drop', '/tmp/x/file.parquet');
  SELECT * FROM ducklake.to_drop;                     
  SET VARIABLE v = (SELECT max(snapshot_id) FROM ducklake_snapshots('ducklake'));
  DROP TABLE ducklake.to_drop;
  BEGIN; CALL ducklake_expire_snapshots('ducklake', older_than => now()); ROLLBACK;
  SELECT * FROM ducklake.to_drop AT (VERSION => getvariable('v'));
```

pdet (commented):
Hi, thanks for the PR, I have one comment!

src/storage/ducklake_metadata_manager.cpp
dentiny: Yeah that's valid point! In the latest commit I buffer mapping id to invalidate, and only apply when transaction commits.

test/configs/quack.json
```diff
@@ -117,6 +117,12 @@
         "test/sql/deletion_inlining/test_deletion_inlining_transaction.test"
       ]
     },
+    {
```
dentiny, line 120: I filed an issue here: https://github.com/duckdb/ducklake/issues/1267, should be a separate issue to investigate.

dentiny:
Hey @guillesd / @pdet , wondering if you have some time to take another look? Thank you!

guillesd:
Looks great! Thanks!

Issue #1168, opened 2026-05-18
Rows inserted by MERGE are not inlined


glenjamin:
### What happens?

When inserting rows using merge, they're added to a parquet file - not inlined. Updating existing inlined rows using merge does write to the inlined table as expected.

### To Reproduce

Can be run piped into `duckdb` CLI.

```sql
.mode box
.shell rm -rf /tmp/ducklake_merge_repro
.shell mkdir -p /tmp/ducklake_merge_repro

INSTALL ducklake;
LOAD ducklake;

ATTACH 'ducklake::memory:' AS lake (DATA_PATH '/tmp/ducklake_merge_repro/');
USE lake;

CREATE TABLE t (id INT, val VARCHAR);

----------------------------------------------------------------------
-- Step 1 — plain INSERT. Expected: row goes to inlined buffer, no
-- parquet file. Observed: matches expectation.
----------------------------------------------------------------------
INSERT INTO t VALUES (1, 'plain-insert');

SELECT '-- after plain INSERT: data files (expect 0 rows) --' AS step;
SELECT data_file_id, path, record_count
FROM   __ducklake_metadata_lake.main.ducklake_data_file;

SELECT '-- after plain INSERT: inlined buffer (expect id=1) --' AS step;
SELECT row_id, id, val
FROM   __ducklake_metadata_lake.main.ducklake_inlined_data_1_1
ORDER BY row_id;

----------------------------------------------------------------------
-- Step 2 — MERGE that only inserts a single new row. Expected: same as
-- plain INSERT — row joins the inlined buffer. Observed: a parquet file
-- is written instead, and the buffer is unchanged.
----------------------------------------------------------------------
MERGE INTO t AS tgt
USING (VALUES (2, 'merge-insert')) AS src(id, val)
    ON tgt.id = src.id
WHEN NOT MATCHED THEN INSERT (id, val) VALUES (src.id, src.val);

SELECT '-- after MERGE insert: data files (BUG: expected 0, got 1) --' AS step;
SELECT data_file_id, path, record_count
FROM   __ducklake_metadata_lake.main.ducklake_data_file;

SELECT '-- after MERGE insert: inlined buffer (BUG: expected id=1 and id=2, got only id=1) --' AS step;
SELECT row_id, id, val
FROM   __ducklake_metadata_lake.main.ducklake_inlined_data_1_1
ORDER BY row_id;

----------------------------------------------------------------------
-- Step 3 — MERGE that only updates an existing inlined row. Expected:
[...]

PR #1218, opened 2026-06-03
Show inlinining issue

jauco:
This isn't really a PR, just the testcase that shows the issue for #1168 (I don't know what's the proper way to submit things like this)


guillesd:
hey @jauco You can add this to the existing issue! I will close this since this is not the right place for it. Thanks!!

PR #1220, opened 2026-06-03
Update table stats for compaction rewrites

jdctinuiti:
Stacked on #1172 (drop-data-files-procedure).

This updates compaction commits to adjust ducklake_table_stats.record_count and file_size_bytes for both ducklake_rewrite_data_files and ducklake_merge_adjacent_files. The change computes removed/added data-file deltas and applies them through the existing global table-stats update path introduced in #1172, preserving next_row_id and existing column stats.

Adds a focused SQL regression covering rewrite-delete with output, merge-adjacent, inlined-file-delete rewrite, and zero-output rewrite after flushing fully-deleted inlined data.

### DIFF
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -97,6 +97,9 @@ static void LoadInternal(ExtensionLoader &loader) {
 	auto add_files = DuckLakeAddDataFilesFunction::GetFunctions();
 	loader.RegisterFunction(add_files);
 
+	auto drop_files = DuckLakeDropDataFilesFunction::GetFunctions();
+	loader.RegisterFunction(drop_files);
+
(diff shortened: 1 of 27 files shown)


dentiny:
Cc @dentiny I will take a look later

jdctinuiti:
Rebased onto the updated #1172. The stats delta update now lives in `DuckLakeTransactionState::CommitChanges`. It runs before the exact REWRITE_DELETES recompute that landed on main (4de135fd), so the exact stats win when that fires. The deltas still cover merge_adjacent and partial rewrites. The diff now also shows the last two #1172 commits because this branch sits on the full #1172 branch. Tests pass locally.

pdet:
See https://github.com/duckdb/ducklake/pull/1171

PR #1223, opened 2026-06-05
Fix data-file partition id remapping on commit retry

jdctinuiti:
## Summary

This PR originally also made identical `SET PARTITIONED BY` alters a no-op; after retargeting and rebasing onto `main`, that part is already covered there (fd4d186), so the PR now contains the remaining fixes:

- **Fix partition id remapping on commit retry.** `GetNewPartitionKey` overwrites `partition_data->partition_id` with the committed id during a commit attempt. If that attempt fails retryably (e.g. a snapshot id collision with a concurrent committer), the retry records the remap as `committed_partition_ids[previous_attempt_id] = new_id`, while data files written by the transaction still carry the original transaction-local partition id — the remap misses and the file is committed referencing a nonexistent partition (found in production). The fix preserves the original transaction-local id on `DuckLakePartition` and keys the commit-time remap off it, making the remap idempotent across commit retries.
- **Defensive commit-time check** that a data file never retains a transaction-local partition id after remapping, raising an `InternalException` instead of silently corrupting metadata.
- **Tests.** `test/sql/transaction/partition_commit_retry_remap.test` deterministically reproduces both retry scenarios (first-time `SET PARTITIONED BY` + INSERT, and re-partition + INSERT, each with a concurrent commit forcing a retry); without the fix the retried commit fails. Also extends `partition_nop.test` to cover a redundant partition alter combined with writes in the same transaction.

### DIFF
--- a/src/include/storage/ducklake_partition_data.hpp
+++ b/src/include/storage/ducklake_partition_data.hpp
@@ -39,6 +39,10 @@ struct DuckLakePartitionField {
 
 struct DuckLakePartition {
 	idx_t partition_id = 0;
+	//! The transaction-local id this partition was assigned when it was created. Data files written in
+	//! the same transaction reference this id. Commit attempts overwrite partition_id with the id of
+	//! that attempt, so on retry the data-file remap must key off the original transaction-local id.
+	optional_idx local_partition_id;
 	vector<DuckLakePartitionField> fields;
 };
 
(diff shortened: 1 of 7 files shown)


dentiny:
Cc @dentiny

pdet:

jdctinuiti:
> Hi, thanks for opening the PR, we are now moving all PRs from `1.5-variegata` to `main`. Could you please retarget?

Hello! Will do thank you! On another note, I have another stack of PRs starting with #1171 that already target `main`, is there anything I need to do for those to get looked at?

jdctinuiti:
@pdet Retargeted to `main` and rebased. The no-op partition alter part was already handled on `main` (fd4d186), so this PR is now just the commit-retry partition id remap fix, the defensive commit-time check, and the tests.

pdet:
Thanks!

PR #1230, opened 2026-06-09
Add option do hide metadata catalog

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/8762

### DIFF
--- a/src/include/common/ducklake_options.hpp
+++ b/src/include/common/ducklake_options.hpp
@@ -30,6 +30,7 @@ struct DuckLakeOptions {
 	DuckLakeEncryption encryption = DuckLakeEncryption::AUTOMATIC;
 	bool create_if_not_exists = true;
 	bool automatic_migration = false;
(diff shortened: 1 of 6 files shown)


src/include/common/ducklake_options.hpp
```diff
@@ -30,6 +30,7 @@ struct DuckLakeOptions {
 	DuckLakeEncryption encryption = DuckLakeEncryption::AUTOMATIC;
 	bool create_if_not_exists = true;
 	bool automatic_migration = false;
+	bool hide_metadata_catalog = true;
```
Mytherin, line 33: I think this should default to hiding only when the `metadata_catalog` is not explicitly defined by the user, i.e. if the user specifies `metadata_catalog '....'` we should not hide the metadata catalog unless `HIDE_METADATA_CATALOG true` is specified, and if the user does not specify `metadata_catalog` we should hide the metadata catalog unless `HIDE_METADATA_CATALOG false` is specified.

PR #1232, opened 2026-06-10
Ensure data inining doesnt cover reserved words for alter

pdet:


suresh-summation:
@pdet Thanks for the quick fix. Just to confirm, will this fix also be included in the 1.5.4 release?

PR #1234, opened 2026-06-10
Fix add-files partition key lookup for repeated transforms

jdctinuiti:
## Summary
- Carry the exact ducklake_add_data_files partition_key_index through hive partition mapping instead of reconstructing it from source field and transform type.
- Validate parsed hive partition values against unique partition key indexes and the full partition transform before writing metadata.
- Add regression coverage for repeated bucket transforms over the same source column.

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -113,6 +113,22 @@ struct HivePartition {
 	DuckLakeTransformType transform_type;
 };
 
+struct HivePartitionLookupKey {
+	FieldIndex field_index;
+	DuckLakeTransformType transform_type;
+
+	bool operator==(const HivePartitionLookupKey &other) const {
+		return field_index.index == other.field_index.index && transform_type == other.transform_type;
+	}
+};
+
+struct HivePartitionLookupKeyHash {
+	hash_t operator()(const HivePartitionLookupKey &key) const {
+		auto result = Hash(key.field_index.index);
+		return CombineHash(result, Hash(static_cast<uint8_t>(key.transform_type)));
+	}
+};
+
 struct ParquetFileMetadata {
 	string filename;
 	vector<unique_ptr<ParquetColumn>> columns;
@@ -1217,7 +1233,8 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 			for (const auto &hive_partition_value : file.hive_partition_values) {
 				bool found_field = false;
 				for (const auto &field : partition_data->fields) {
-					if (field.field_id.index == hive_partition_value.field_index.index) {
+					if (field.field_id.index == hive_partition_value.field_index.index &&
+					    field.transform.type == hive_partition_value.transform_type) {
 						found_field = true;
 						break;
 					}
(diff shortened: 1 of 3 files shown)


src/functions/ducklake_add_data_files.cpp
```diff
@@ -1232,13 +1249,19 @@ DuckLakeDataFile DuckLakeFileProcessor::AddFileToTable(ParquetFileMetadata &file
 			throw InvalidInputException("File \"%s\" contains an invalid partition value for the table configuration.",
 			                            file.filename);
 		}
-		unordered_map<idx_t, idx_t> field_partition_key_map;
+		unordered_map<HivePartitionLookupKey, idx_t, HivePartitionLookupKeyHash> field_partition_key_map;
 		for (auto &partition_fields : partition_data->fields) {
-			field_partition_key_map[partition_fields.field_id.index] = partition_fields.partition_key_index;
+			field_partition_key_map[{partition_fields.field_id, partition_fields.transform.type}] =
+			    partition_fields.partition_key_index;
```
pdet: What if you have the same column and the same transformation type? e.g.,

```sql
# name: test/sql/add_files/add_file_repeated_bucket.test
# description: ducklake_add_data_files must map each partition field to the correct partition_key_index when the same column feeds two BUCKET transforms with different bucket counts
# group: [add_files]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/add_file_repeated_bucket/', METADATA_CATALOG 'metadata');

statement ok
USE ducklake;

statement ok
CREATE TABLE double_bucket_t (id INT, c INT);

statement ok
ALTER TABLE double_bucket_t SET PARTITIONED BY (bucket(4, c), bucket(8, c));

statement ok
COPY (SELECT 10 AS id, 42 AS c, 1 AS bucket, 5 AS bucket_c)
TO '{DATA_PATH}/repeated_bucket_staged'
(FORMAT parquet, PARTITION_BY (bucket, bucket_c));

statement ok
CALL ducklake_add_data_files('ducklake', 'double_bucket_t', '{DATA_PATH}/repeated_bucket_staged/bucket=1/bucket_c=5/*.parquet')

statement ok
SET VARIABLE bucket_file_id = (
	SELECT data_file_id
	FROM metadata.ducklake_data_file
	WHERE path LIKE '%bucket_c=5%'
)

query II
SELECT partition_key_index, partition_value
FROM metadata.ducklake_file_partition_value
WHERE data_file_id = getvariable('bucket_file_id')
ORDER BY partition_key_index
----
0	1
1	5
```

pdet:
Thanks for the PR, I have a comment

src/functions/ducklake_add_data_files.cpp
jdctinuiti: Good catch ty! I updated the PR to avoid reconstructing the partition key from `(field_id, transform_type)` entirely. Each parsed hive partition now carries the exact `partition_key_index` from the matching `DuckLakePartitionField`.

jdctinuiti: Also added the repeated bucket regression case for bucket(4, c), bucket(8, c)

pdet:
Thanks! LGTM!

PR #1243, opened 2026-06-16
Ensure REWRITE_DELETES respects target_file_size

### LINKED ISSUE #1241: `ducklake_rewrite_data_files` merges files instead of rewriting each file individually
### What happens?

When calling `ducklake_rewrite_data_files` on a table where multiple data files have deletions, it will merge all data files per per partition into a single target file. The expected behavior would be that each data file is rewritten separately into a new data file, and that `ducklake_merge_adjacent_files` is necessary to merge the resulting smaller files together.

From the [following comment in the source](https://github.com/duckdb/ducklake/blob/main/src/functions/ducklake_compaction_functions.cpp#L297), this seems intentional: "For REWRITE_DELETES, generate one compaction command per partition group". However, @pdet's reply on Discord seems to imply that each rewritten data file should result in a separate target file, not that they are merged together.

The fix should be simple. In `DuckLakeCompactor::GenerateCompactions`, instead of generating a compaction command for each compaction group, we would create one for each candidate file. Happy to submit a PR for that if you're interested.

### To Reproduce

First, create a DuckLake table with two data files, each of which also has a delete file:
```sql
ATTACH 'ducklake:test.ducklake';
USE test;

CREATE TABLE t (i BIGINT);
INSERT INTO t FROM range(1000);
INSERT INTO t FROM range(1000);

DELETE FROM t WHERE i < 100;
```
To confirm:
```sql
SELECT parse_filename(data_file), parse_filename(delete_file) FROM ducklake_list_files('test', 't');
```
shows two data files with associated delete files:
```
┌───────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────────┐
│               parse_filename(data_file)               │                 parse_filename(delete_file)                  │
│                        varchar                        │                           varchar                            │
├───────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
│ ducklake-019ecb49-b741-7973-a8f8-4a24654c6a54.parquet │ ducklake-019ecb49-baac-763b-8aba-ecf46dcfc022-delete.parquet │
│ ducklake-019ecb49-b74f-76c7-9c6a-6390887c8fc1.parquet │ ducklake-019ecb49-baab-7958-a382-7a9dbf8f4a2b-delete.parquet │
└───────────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────┘
```
Then, rewrite the data files:
```sql
CALL ducklake_rewrite_data_files('test', 't', delete_threshold => 0);
```
The result already indicates that the files were merged together:
```
┌─────────────┬────────────┬─────────────────┬───────────────┐
│ schema_name │ table_name │ files_processed │ files_created │
│   varchar   │  varchar   │      int64      │     int64     │
├─────────────┼────────────┼─────────────────┼───────────────┤
│ main        │ t          │               2 │             1 │
└─────────────┴────────────┴─────────────────┴───────────────┘
```
`ducklake_list_files` now also shows a single data file:
```sql
SELECT parse_filename(data_file), parse_filename(delete_file) FROM ducklake_list_files('test', 't');
```
results in:
```
┌───────────────────────────────────────────────────────┬─────────────────────────────┐
│               parse_filename(data_file)               │ parse_filename(delete_file) │
│                        varchar                        │           varchar           │
├───────────────────────────────────────────────────────┼─────────────────────────────┤
│ ducklake-019ecb4b-47ac-711c-9617-c8922bedeeb6.parquet │ NULL                        │
└───────────────────────────────────────────────────────┴─────────────────────────────┘
```

### OS:

x86_64

### DuckDB Version:

`v1.6.0-dev8546`

### DuckLake Version:

`478761c1`

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Gijs Hendriksen

### Affiliation:

Radboud University

Yes

- [x] Yes, I have

- [x] Yes, I have

gijshendriksen:
Fixes #1241 

Changed `ducklake_rewrite_data_files` to process each input file individually, rather than merging them together. @pdet please let me know if this is the desired behavior.

The change simplifies the checks in `DuckLakeCompactor::GenerateCompactions`, because the paths for `REWRITE_DELETES` and `MERGE_ADJACENT_TABLES` are separated more clearly. One note: I left in the check for `end_snapshot.IsValid()`, but the query in `metadata_manager.GetFilesForCompaction` does also check for `end_snapshot IS NULL`. Do we want to leave the check in here as well (for clarity/robustness), or can it be removed?

I also updated the relevant tests that rewrite deletes and count the resulting files, manually verifying whether the changes made sense. CI completed successfully [on my fork](https://github.com/gijshendriksen/ducklake/pull/1/checks).

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -295,30 +295,13 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 
 		candidates[group].candidate_files.push_back(file_idx);
 	}
-	if (type == CompactionType::REWRITE_DELETES) {
-		// For REWRITE_DELETES, generate one compaction command per partition group
-		for (auto &entry : candidates) {
-			auto &candidate_list = entry.second.candidate_files;
-			if (candidate_list.empty()) {
-				continue;
-			}
-			vector<DuckLakeCompactionFileEntry> partition_files;
-			for (auto &candidate_idx : candidate_list) {
-				partition_files.push_back(std::move(files[candidate_idx]));
-			}
-			auto compaction_command = GenerateCompactionCommand(std::move(partition_files));
-			if (compaction_command) {
-				compactions.push_back(std::move(compaction_command));
-			}
-		}
-		return;
-	}
+
 	// we have gathered all the candidate files per compaction group
 	// iterate over them to generate actual compaction commands
 	uint64_t compacted_files = 0;
 	for (auto &entry : candidates) {
 		auto &candidate_list = entry.second.candidate_files;
-		if (candidate_list.size() <= 1) {
+		if (type == CompactionType::MERGE_ADJACENT_TABLES && candidate_list.size() <= 1) {
 			// we need at least 2 files to consider a merge
 			continue;
 		}
@@ -327,15 +310,17 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 			idx_t current_file_size = 0;
 			idx_t compaction_idx;
 			for (compaction_idx = start_idx; compaction_idx < candidate_list.size(); compaction_idx++) {
-				if (current_file_size >= target_file_size) {
-					// we hit the target size already - stop
-					break;
-				}
 				auto candidate_idx = candidate_list[compaction_idx];
 				auto &candidate = files[candidate_idx];
 				idx_t file_size = candidate.file.data.file_size_bytes;
-				if (file_size >= target_file_size) {
-					// don't consider merging if the file is larger than the target size
+				if (type == CompactionType::REWRITE_DELETES) {
+					// estimate size of remaining rows
+					file_size *= (1 - candidate.delete_ratio);
+				}
+				const int64_t current_size_diff = NumericCast<int64_t>(current_file_size) - target_file_size;
+				const int64_t merged_size_diff = NumericCast<int64_t>(current_file_size + file_size) - target_file_size;
+				if (current_file_size > 0 && std::abs(merged_size_diff) >= std::abs(current_size_diff)) {
+					// adding this file would move away from target_file_size - stop
 					break;
 				}
 				// this file can be compacted along with the neighbors
@@ -344,8 +329,8 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 
 			if (start_idx < compaction_idx) {
 				idx_t compaction_file_count = compaction_idx - start_idx;
-				if (compaction_file_count == 1) {
-					// If we only have one file to compact, we have nothing to compact
+				if (type == CompactionType::MERGE_ADJACENT_TABLES && compaction_file_count == 1) {
+					// If we only have one file to merge, we have nothing to compact
 					compacted_files++;
 					if (compacted_files >= options.max_files) {
 						break;
@@ -360,11 +345,11 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 				start_idx += compaction_file_count - 1;
 			}
 			compacted_files++;
-			if (compacted_files >= options.max_files) {
+			if (type == CompactionType::MERGE_ADJACENT_TABLES && compacted_files >= options.max_files) {
 				break;
 			}
 		}
-		if (compacted_files >= options.max_files) {
+		if (type == CompactionType::MERGE_ADJACENT_TABLES && compacted_files >= options.max_files) {
 			break;
 		}
 	}
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -462,6 +462,7 @@ struct DuckLakeCompactionFileEntry {
 	set<idx_t> inlined_file_deletions;
 	//! Whether this file has any inlined deletions (cheap flag; set for all compaction types).
 	bool has_inlined_deletions = false;
+	double delete_ratio = 0;
 };
 
 struct DuckLakeRewriteFileEntry {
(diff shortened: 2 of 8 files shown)


pdet:
Hi @gijshendriksen thanks for the PR.

Now that I'm looking at the code, I think the original idea was that since we are rewriting files anyway, we could also take that moment to do merge_adjacent if applicable, I guess your original issue was that the file_size was not being respected by that?

gijshendriksen:
Hi @pdet, thanks for the quick response! Indeed, `target_file_size` was not taken into account when building the compaction commands, so that mainly caused the issues of all data being rewritten to a single file.

I considered just incorporating `target_file_size` in the `REWRITE_DELETES` path, but since I thought the intention was that `REWRITE_DELETES` was supposed to rewrite the files individually I applied this fix instead. If you want, I can update the PR to follow the other behavior instead.

pdet:
Yeah, let's just add `target_file_size` and test that, sorry for the confusion!

gijshendriksen:
Alright, makes sense! No worries :)

Quick follow-up question: do you want to use the delete ratio to estimate the new file size, or just stick with the old size as an upper bound?

pdet:
> Alright, makes sense! No worries :)
> 
> Quick follow-up question: do you want to use the delete ratio to estimate the new file size, or just stick with the old size as an upper bound?

I think estimating makes more sense, as this compaction should mainly be used with high delete ratios. Assume all rows have the same size, should be good enough, with the exception of extreme corner cases.

gijshendriksen:
I've updated the PR to respect `target_file_size` in `ducklake_rewrite_data_files`. I've slightly changed the generation of compaction commands to check for `target_file_size` _before_ adding a new file, rather than after. This should prevent overshooting of `target_file_size`. Otherwise, two 500MB files might be merged into a 1GB file, even if `target_file_size` is set to 512MB. I had to change a couple of tests to account for this change. I've also added some tests to check that my fix works as intended.

Should be ready for review now!

pdet:
Hi @gijshendriksen, CI seems to be failing?

gijshendriksen:
Yes, sorry! I didn't account for tests that disabled inlining. I also found another flaw in my changes which I've fixed and incorporated in the tests. I'm running CI on my fork now before updating the PR again.

gijshendriksen:
@pdet should be good to go now!

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -327,15 +310,15 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
...
 				auto &candidate = files[candidate_idx];
 				idx_t file_size = candidate.file.data.file_size_bytes;
-				if (file_size >= target_file_size) {
-					// don't consider merging if the file is larger than the target size
+				if (type == CompactionType::REWRITE_DELETES) {
+					// estimate size of remaining rows
+					file_size *= (1 - candidate.delete_ratio);
+				}
+				if (current_file_size > 0 && current_file_size + file_size >= target_file_size) {
+					// adding this file would make the target too large - stop
```
pdet: This is a behavior change in the merging code, not necessarily wrong, but before that we would merge files if they were smaller than the target size, this could cause an overshoot of up to 2x the target size, the tradeoff is that we wouldnt have files with size between 0.5 and target size. 

This changes the behavior to the opposite direction.

@guillesd from a user perspective, what you think is more desirable?

guillesd: So if I understand this correctly if current_file_size = 500MB and file_size = 13MB, with the new behaviour this 13MB file will not be compacted?

gijshendriksen: Yeah, it felt like this was a more appropriate choice for the `REWRITE_DELETES` compaction, but I agree it might be too conservative in some cases. We could perhaps also do a check like:
```
abs(current_file_size + file_size - target_file_size) >= abs(current_file_size - target_file_size)
```
So if the new size overshoots `target_file_size`, but with a smaller difference than that the current size undershoots, we can still merge the files together. That would mean the merge will undershoot/overshoot depending on which value is closer to `target_file_size`, and the overall file sizes may be more balanced as a result. What do you think?

pdet: say
`target_file_size = 100mb`
`current_file_size = 60mb`
`file_size = 60mb`

`current_file_size` and `file_size` wouldn't be merged together.

In the previous code, they would, but the final file would be 120mb, hence an overshoot.

pdet: > Yeah, it felt like this was a more appropriate choice for the `REWRITE_DELETES` compaction, but I agree it might be too conservative in some cases. We could perhaps also do a check like:
> 
> ```
> abs(current_file_size + file_size - target_file_size) >= abs(current_file_size - target_file_size)
> ```
> 
> So if the new size overshoots `target_file_size`, but with a smaller difference than that the current size undershoots, we can still merge the files together. That would mean the merge will undershoot/overshoot depending on which value is closer to `target_file_size`, and the overall file sizes may be more balanced as a result. What do you think?

I like it

guillesd: I like the idea too. The only thing that I would also keep in mind is: is there a check on the last file? I'd be worries that if there is one file left that doesn't match this condition but it is considerably small, then that one lingers around. Although to be fair it can get picked up in the next rewrite/compaction no? I'm not super familiar with the code path just speaking out loud!

gijshendriksen: > I like the idea too. The only thing that I would also keep in mind is: is there a check on the last file? I'd be worries that if there is one file left that doesn't match this condition but it is considerably small, then that one lingers around. Although to be fair it can get picked up in the next rewrite/compaction no? I'm not super familiar with the code path just speaking out loud!

Yes, I think that issue remains. Though I think this is also a problem with the current implementation, and it's mainly caused by the fact that we're greedily selecting data files that can be compacted (and may also be a problem in general with writes that need to accommodate `target_file_size`?).

We probably cannot easily find the optimal set of compaction groups, so maybe we should indeed just accept that the final file may not be included and could be picked up in a next compaction round.

guillesd: cool! I'm happy with the `abs` solution then!

gijshendriksen: Great, thanks for the feedback! I've applied the fix in b32c157.

pdet:
Thanks!! Looks great!

PR #1247, opened 2026-06-17
Update table stats when DELETEs drop whole files

jdctinuiti:
When a `DELETE` removes every row of a data file, the file is dropped from the active set, but the table stats were left untouched. `record_count` and `file_size_bytes` kept counting rows and bytes that no longer exist, so repeated delete/reinsert cycles inflated cardinality estimates and the size reported by `duckdb_tables`.

This subtracts the dropped files' stats at commit time:
  - track row count and file size for each dropped committed data file.
  - subtract those amounts from the table stats during commit, including the server-side commit path.
  - clear the table column stats when a drop leaves the recorded min/max stale. Exact min/max cannot be recomputed without rescanning the surviving files, so they are reset to unknown and rebuilt on the next write.
  - `next_row_id` stays monotonic. The stats cache is invalidated after delete-only commits, since those commits do not advance `next_file_id`.

### Whole-file deletes under inlining

A fully-deleted file is now dropped before a small delete would be routed into inlined file deletes. Before this change, with the default inline limit, deleting every row of a small file inlined the deletes and left the dead file in place with stale stats. Dropping it is cheaper and keeps the stats correct. Only the case with no existing delete file takes this path; deletes that merge with an existing delete file are unchanged.

### DIFF
--- a/src/storage/ducklake_server_side_commit.cpp
+++ b/src/storage/ducklake_server_side_commit.cpp
@@ -47,6 +47,17 @@ struct EntryShell {
 	bool hive_partition;
 };
 
+static string JoinIds(const vector<idx_t> &ids) {
+	string result;
+	for (auto id : ids) {
+		if (!result.empty()) {
+			result += ",";
+		}
+		result += to_string(id);
+	}
+	return result;
+}
+
 unique_ptr<DuckLakeNameMapEntry> BuildNameMapEntry(idx_t id, const std::map<idx_t, const EntryShell *> &shell_by_id,
                                                    const std::map<idx_t, vector<idx_t>> &children_of) {
 	auto &shell = *shell_by_id.at(id);
@@ -212,6 +223,20 @@ void DuckLakeServerSideCommit::ReadColumnTypes() {
 	auto inlined = ScanStagedTable(DuckLakeStagedTableType::INLINED_DATA);
 	for (auto &row : *inlined) {
 		table_ids.insert(AsIdx(row, 0));
+	}
+	vector<idx_t> dropped_file_ids;
+	auto dropped_files = ScanStagedTable(DuckLakeStagedTableType::DROPPED_FILE);
+	for (auto &row : *dropped_files) {
+		dropped_file_ids.push_back(AsIdx(row, 1));
+	}
+	if (!dropped_file_ids.empty()) {
+		auto dropped_tables = RunQuery(
+		    StringUtil::Format("SELECT DISTINCT table_id FROM %s.ducklake_data_file WHERE data_file_id IN (%s)",
+		                       schema_id, JoinIds(dropped_file_ids)),
+		    "read dropped file table ids");
+		for (auto &row : *dropped_tables) {
+			table_ids.insert(AsIdx(row, 0));
+		}
 	}
 	if (table_ids.empty()) {
 		return;
@@ -445,8 +470,23 @@ void DuckLakeServerSideCommit::ReadStagedDeleteFiles() {
 
 void DuckLakeServerSideCommit::ReadStagedDroppedFiles() {
 	auto dropped = ScanStagedTable(DuckLakeStagedTableType::DROPPED_FILE);
+	vector<idx_t> dropped_file_ids;
 	for (auto &row : *dropped) {
-		state->dropped_files.emplace(row.GetValue<string>(0), DataFileIndex(AsIdx(row, 1)));
+		auto file_id = AsIdx(row, 1);
+		state->dropped_files.emplace(row.GetValue<string>(0), DataFileIndex(file_id));
+		dropped_file_ids.push_back(file_id);
+	}
+	if (!dropped_file_ids.empty()) {
+		auto dropped_stats = RunQuery(
+		    StringUtil::Format(
+		        "SELECT table_id, record_count, file_size_bytes FROM %s.ducklake_data_file WHERE data_file_id IN (%s)",
+		        schema_id, JoinIds(dropped_file_ids)),
+		    "read dropped file stats");
+		for (auto &row : *dropped_stats) {
+			auto &stats = state->dropped_file_stats[TableIndex(AsIdx(row, 0))];
+			stats.row_count += AsIdx(row, 1);
+			stats.file_size_bytes += AsIdx(row, 2);
+		}
 	}
 	auto tables = ScanStagedTable(DuckLakeStagedTableType::TABLES_DELETED_FROM);
 	for (auto &row : *tables) {
--- a/src/storage/ducklake_server_side_commit.cpp
+++ b/src/storage/ducklake_server_side_commit.cpp
@@ -48,14 +48,7 @@ struct EntryShell {
 };
 
 static string JoinIds(const vector<idx_t> &ids) {
-	string result;
-	for (auto id : ids) {
-		if (!result.empty()) {
-			result += ",";
-		}
-		result += to_string(id);
-	}
-	return result;
+	return StringUtil::Join(ids, ids.size(), ",", [](const idx_t &id) { return to_string(id); });
 }
 
 unique_ptr<DuckLakeNameMapEntry> BuildNameMapEntry(idx_t id, const std::map<idx_t, const EntryShell *> &shell_by_id,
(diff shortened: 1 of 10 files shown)


src/storage/ducklake_transaction_state.cpp
```diff
@@ -934,9 +934,97 @@ void DuckLakeTransactionState::RecomputeGlobalStatsAfterRewrite(string &batch_qu
...
+	if (entry == attempt_dropped_file_stats.end()) {
+		return false;
+	}
+	auto &stats = new_stats.stats;
+	stats.record_count = SaturatingSubtract(stats.record_count, entry->second.row_count);
+	stats.table_size_bytes = SaturatingSubtract(stats.table_size_bytes, entry->second.file_size_bytes);
+	attempt_dropped_file_stats.erase(entry);
+	// File drops can make table-level min/max stale. If live rows remain in files we did not drop, the caller
+	// deletes the column-stats rows so later inserts do not rebuild stats from only the newly inserted files.
+	// If the table is now empty, there are no surviving rows, so reset the column stats to NULL in place instead.
```
pdet: Let's avoid comments that span more than one line

test/sql/delete/full_file_delete_stats.test
```diff
@@ -0,0 +1,112 @@
+# name: test/sql/delete/full_file_delete_stats.test
```
pdet, line 1: Can we add tests with compaction?

e.g.,
```sql
CREATE TABLE dl.t AS SELECT range::INT id FROM range(50);
INSERT INTO dl.t SELECT range::INT id FROM range(50, 100);
INSERT INTO dl.t SELECT range::INT id FROM range(100, 150);
INSERT INTO dl.t SELECT range::INT id FROM range(150, 200);
CALL ducklake_merge_adjacent_files('dl'); 
DELETE FROM dl.t;   
-- check stats are correct 
-- do the same for rewrite_data

```

src/storage/ducklake_transaction_state.cpp
```diff
@@ -934,9 +934,97 @@ void DuckLakeTransactionState::RecomputeGlobalStatsAfterRewrite(string &batch_qu
 	    DuckLakeTransaction::ConvertNewGlobalStats(table_id, new_globals));
 }
 
+static idx_t SaturatingSubtract(idx_t value, idx_t decrement) {
+	return decrement > value ? 0 : value - decrement;
```
pdet: Is `decrement > value` a valid value?

src/storage/ducklake_transaction_state.cpp
```diff
@@ -934,9 +934,97 @@ void DuckLakeTransactionState::RecomputeGlobalStatsAfterRewrite(string &batch_qu
...
+		if (attempt_dropped_file_stats.find(table_id) == attempt_dropped_file_stats.end()) {
+			continue;
+		}
+
+		optional_ptr<DuckLakeTableStats> current_stats;
+		shared_ptr<DuckLakeTableStats> current_stats_pin;
+		if (dl_stats) {
+			auto dl_stats_entry = dl_stats->table_stats.find(table_id);
+			if (dl_stats_entry != dl_stats->table_stats.end()) {
+				current_stats = dl_stats_entry->second.get();
```
pdet, line 1002: A file's table always has table_stats no? Is this a valid state?

src/storage/ducklake_server_side_commit.cpp
```diff
@@ -47,6 +47,17 @@ struct EntryShell {
...
+static string JoinIds(const vector<idx_t> &ids) {
+	string result;
+	for (auto id : ids) {
+		if (!result.empty()) {
+			result += ",";
+		}
+		result += to_string(id);
+	}
+	return result;
+}
```
pdet, line 52: Use StringUtil::Join

src/storage/ducklake_server_side_commit.cpp
```diff
@@ -213,6 +224,20 @@ void DuckLakeServerSideCommit::ReadColumnTypes() {
 	for (auto &row : *inlined) {
 		table_ids.insert(AsIdx(row, 0));
 	}
+	vector<idx_t> dropped_file_ids;
+	auto dropped_files = ScanStagedTable(DuckLakeStagedTableType::DROPPED_FILE);
```
pdet: This is also read in `ReadStagedDroppedFiles`, should we fold it into one scan + one ducklake_data_file query?

pdet (commented):
Hi @jdctinuiti, thanks for the PR, I have a few comments!

src/storage/ducklake_server_side_commit.cpp
jdctinuiti: swapped now

test/sql/delete/full_file_delete_stats.test
jdctinuiti: added!

src/storage/ducklake_transaction_state.cpp
jdctinuiti: Good point. I changed this to strict subtraction that throws if dropped-file stats exceed current table stats.

The only special case is when the delete empties the table: record_count is still validated strictly, and file_size_bytes is normalized to 0 because that is the exact final state when no active data files remain.

src/storage/ducklake_transaction_state.cpp
jdctinuiti: Changed this to throw an internal exception.

src/storage/ducklake_server_side_commit.cpp
jdctinuiti: Done! The server-side commit path now reads staged dropped-file rows once, queries ducklake_data_file once for table/stat data, and reuses that cache in both ReadColumnTypes and ReadStagedDroppedFiles.

src/storage/ducklake_transaction_state.cpp
jdctinuiti: sounds good!

jdctinuiti:
@pdet should hopefully be ready now!

pdet:
Hi @jdctinuiti, my clanker signaled the following test failure, could you have a look:

```sql
# name: test/sql/delete/zzbreak_sametxn_anyvalid.test
# description: BREAK - same-transaction drop-to-empty + insert loses column stats -> wrong pruned results
# group: [delete]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/zzbreak', DATA_INLINING_ROW_LIMIT 0)

statement ok
CREATE TABLE ducklake.t AS SELECT range::INT id FROM range(100)

statement ok
SET VARIABLE t_id = (SELECT table_id FROM __ducklake_metadata_ducklake.ducklake_table WHERE table_name = 't')

# Same transaction: drop the whole file (empties table) then insert a NEW disjoint range 200..209.
statement ok
BEGIN

statement ok
DELETE FROM ducklake.t

statement ok
INSERT INTO ducklake.t SELECT range::INT id FROM range(200, 210)

statement ok
COMMIT

# record_count is correctly decremented to 10 ...
query I
SELECT record_count FROM __ducklake_metadata_ducklake.ducklake_table_stats WHERE table_id = getvariable('t_id')
----
10

# ... but the column-level min/max are WRONGLY reset to NULL (should be 200 / 209).
# On HEAD this row is (NULL, NULL).
query TT
SELECT min_value, max_value FROM __ducklake_metadata_ducklake.ducklake_table_column_stats WHERE table_id = getvariable('t_id')
----
200	209

# Now a NARROWER insert in a separate txn: 50..59. Table then holds {50..59, 200..209}.
statement ok
INSERT INTO ducklake.t SELECT range::INT id FROM range(50, 60)

# A point lookup that relies on min/max pruning. Row 205 exists.
# On HEAD the stats have collapsed to {50..59}, so the file holding 205 is pruned and the count is 0.
query I
SELECT COUNT(*) FROM ducklake.t WHERE id = 205
----
1

query I
SELECT COUNT(*) FROM ducklake.t WHERE id > 100
----
10

query I
SELECT COUNT(*) FROM ducklake.t
----
20
```

jdctinuiti:
@pdet Good catch, ty! Fixed by setting `any_valid = false` on the reset so a same-commit insert seeds the min/max instead. Also added the relevant test.

pdet:
Hi @jdctinuiti thanks. I think it looks good!

One thing to keep in mind is that I think we can actually recalculate the min/max by just checking the file stats table at commit time and keep them valid, but we can also do that in a follow-up.

jdctinuiti:
> Hi @jdctinuiti thanks. I think it looks good!
> 
> One thing to keep in mind is that I think we can actually recalculate the min/max by just checking the file stats table at commit time and keep them valid, but we can also do that in a follow-up.

Hey @pdet ty! Did the follow up in https://github.com/duckdb/ducklake/pull/1260

PR #1250, opened 2026-06-18
Implement inlining for the MERGE INTO ... INSERT case (fixes #1186)

### LINKED ISSUE #1186: UHUGEINT/HUGEINT columns stored as DOUBLE in Parquet, corrupting values
## What happens

DuckLake writes `UHUGEINT` and `HUGEINT` columns as `DOUBLE` in the underlying Parquet data files, causing silent precision loss for any value with more than ~15 significant digits. This affects **all** write paths (INSERT, MERGE INSERT) once the batch is large enough to produce a Parquet file. Small inserts appear correct because DuckLake inlines them into catalog metadata instead of writing Parquet.

## Minimal reproduction

```sql
ATTACH 'ducklake:sqlite:repro.sqlite' AS lake
  (DATA_PATH 'repro_data', METADATA_CATALOG 'meta');
CREATE TABLE lake.main.t (val uhugeint);

-- Force Parquet write (batch > inline threshold)
INSERT INTO lake.main.t SELECT md5_number(i::varchar) FROM range(5000) t(i);

-- Check first value
SELECT val, md5_number('0') AS expected FROM lake.main.t LIMIT 1;
-- ┌─────────────────────────────────────────┬─────────────────────────────────────────┐
-- │                   val                   │                expected                 │
-- │                 uint128                 │                 uint128                 │
-- ├─────────────────────────────────────────┼─────────────────────────────────────────┤
-- │ 290293690697406555575523292621523910656 │ 290293690697406552751975134056109100495 │
-- └─────────────────────────────────────────┴─────────────────────────────────────────┘
-- Values differ in the low-order digits — consistent with DOUBLE precision loss.
```

Inspecting the Parquet file confirms the type:

```sql
SELECT name, type FROM parquet_schema('repro_data/main/t/*.parquet');
-- ┌─────┬────────┐
-- │ val │ DOUBLE │  ← should be FIXED_LEN_BYTE_ARRAY or similar 128-bit type
-- └─────┴────────┘
```

## Why small INSERTs appear correct

DuckLake inlines small inserts into catalog metadata (`inlined_insert` in snapshot changes), bypassing Parquet entirely. Only once the batch exceeds the inline threshold does the bug manifest. This makes it look like INSERT works correctly while MERGE doesn't — but both paths produce DOUBLE Parquet files.

```sql
-- Small insert: inlined, correct
INSERT INTO lake.main.t VALUES (12345678901234567890);
SELECT changes FROM ducklake_snapshots('lake') ...;
-- {inlined_insert=[1]}  ← no Parquet file written

-- Large insert: Parquet, corrupted
INSERT INTO lake.main.t SELECT md5_number(i::varchar) FROM range(5000) t(i);
SELECT changes FROM ducklake_snapshots('lake') ...;
-- {tables_inserted_into=[1]}  ← Parquet file written with DOUBLE type
```

## Impact

Any DuckLake table using 128-bit integer columns (hash keys, surrogate keys, UUIDs as UHUGEINT) silently corrupts data once rows spill to Parquet. This also breaks MERGE-based upsert workflows, since stored values don't match recomputed hashes, causing duplicate inserts instead of updates.

## Environment

- DuckDB v1.5.2
- DuckLake extension (version bundled with v1.5.2)
- Linux x86_64
- Catalog backends tested: SQLite, DuckDB

jauco:
Commit https://github.com/duckdb/ducklake/commit/c89438ff added the update implementation for MERGE INTO with support for inlining, but the INSERT case never had it. I think it was simply forgotten or deferred. So I added the insert handling analogous to what is already happening.

This is my first PR for anything duckdb related and it took me quite some time to grok the duckdb/ducklake codebase but the end result seemed so obvious given the UPDATE branch paths that I think it should be correct.

I didn't see a HACKING.md but feel free to let me know if there are things I should take into account.

### DIFF
--- a/src/storage/ducklake_merge_into.cpp
+++ b/src/storage/ducklake_merge_into.cpp
@@ -155,6 +155,24 @@ static void SinkInlineOrCopy(ExecutionContext &context, DataChunk &execute_input
 	}
 }
 
+static void DrainInlineData(optional_ptr<DuckLakeInlineData> inline_data_op, ExecutionContext &context, OperatorSinkCombineInput &input, PhysicalOperator &physical_copy, DuckLakeMergeSinkGlobalState &gstate, DuckLakeMergeSinkLocalState &lstate) {
+	auto &inline_output = lstate.inline_output;
+	while (true) {
+		lstate.inline_output.Reset();
+		auto fresult = inline_data_op->FinalExecute(context, inline_output, *gstate.inline_data_gstate,
+		                                            *lstate.inline_data_lstate);
+		if (inline_output.size() > 0) {
+			ProjectAndCastForCopy(context.client, inline_output, physical_copy, lstate.expression_executor.get(),
+			                      lstate.projected_chunk, lstate.cast_chunk);
+			OperatorSinkInput copy_input {*physical_copy.sink_state, *lstate.copy_sink_state, input.interrupt_state};
+			physical_copy.Sink(context, lstate.cast_chunk, copy_input);
+		}
+		if (fresult == OperatorFinalizeResultType::FINISHED) {
+			break;
+		}
+	}
+}
+
 SinkResultType DuckLakeMergeInsert::Sink(ExecutionContext &context, DataChunk &chunk, OperatorSinkInput &input) const {
 	auto &gstate = input.global_state.Cast<DuckLakeMergeSinkGlobalState>();
 	auto &lstate = input.local_state.Cast<DuckLakeMergeSinkLocalState>();
@@ -169,21 +187,7 @@ SinkCombineResultType DuckLakeMergeInsert::Combine(ExecutionContext &context, Op
 
 	// drain inline data
 	if (inline_data_op) {
-		auto &inline_output = lstate.inline_output;
-		while (true) {
-			lstate.inline_output.Reset();
-			auto fresult = inline_data_op->FinalExecute(context, inline_output, *gstate.inline_data_gstate,
-			                                            *lstate.inline_data_lstate);
-			if (inline_output.size() > 0) {
-				ProjectAndCastForCopy(context.client, inline_output, physical_copy, lstate.expression_executor.get(),
-				                      lstate.projected_chunk, lstate.cast_chunk);
-				OperatorSinkInput copy_input {*physical_copy.sink_state, *lstate.copy_sink_state, input.interrupt_state};
-				physical_copy.Sink(context, lstate.cast_chunk, copy_input);
-			}
-			if (fresult == OperatorFinalizeResultType::FINISHED) {
-				break;
-			}
-		}
+		DrainInlineData(inline_data_op, context, input, physical_copy, gstate, lstate);
 	}
 
 	OperatorSinkCombineInput combine_input {*physical_copy.sink_state, *lstate.copy_sink_state, input.interrupt_state};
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_merge_into.cpp
```diff
@@ -100,25 +104,82 @@ static void ProjectAndCastForCopy(ClientContext &context, DataChunk &input_chunk
...
+				OperatorSinkInput copy_input {*copy.sink_state, *local_state.copy_sink_state, input.interrupt_state};
+				copy.Sink(context, local_state.cast_chunk, copy_input);
+			}
+		}
+	} else {
+		ProjectAndCastForCopy(context.client, chunk, copy, local_state.expression_executor.get(), local_state.chunk,
+													local_state.cast_chunk);
+		OperatorSinkInput sink_input {*copy.sink_state, *local_state.copy_sink_state, input.interrupt_state};
+		return copy.Sink(context, local_state.cast_chunk, sink_input);
+	}
```
pdet, line 155: Is this structurally the same as `DuckLakeMergeUpdate::Sink`? If so, can we use a helper function to dedup the code?

src/storage/ducklake_merge_into.cpp
```diff
@@ -100,25 +104,82 @@ static void ProjectAndCastForCopy(ClientContext &context, DataChunk &input_chunk
...
+				ProjectAndCastForCopy(context.client, inline_output, copy, local_state.expression_executor.get(),
+				                      local_state.chunk, local_state.cast_chunk);
+				OperatorSinkInput copy_input {*copy.sink_state, *local_state.copy_sink_state, input.interrupt_state};
+				copy.Sink(context, local_state.cast_chunk, copy_input);
+			}
+			if (fresult == OperatorFinalizeResultType::FINISHED) {
+				break;
+			}
+		}
+	}
```
pdet, line 191: Is this structurally the same as `DuckLakeMergeUpdate::Combine`? If so, can we use a helper function to dedup the code?

src/storage/ducklake_merge_into.cpp
```diff
@@ -539,12 +614,32 @@ static unique_ptr<MergeIntoOperator> DuckLakePlanMergeIntoAction(DuckLakeCatalog
 		auto &copy_dummy =
 		    planner.Make<PhysicalDummyScan>(physical_copy.Cast<PhysicalCopyToFile>().expected_types, 1);
 		physical_copy.children.push_back(copy_dummy);
+
+		// maybe wrap with InlineData if we hit the row limit
+		optional_ptr<DuckLakeInlineData> inline_data;
+		idx_t data_inlining_row_limit = catalog.GetInliningLimit(context, ducklake_table);
+		if (data_inlining_row_limit > 0) {
+			// the inlined chunks hold only the table's physical columns - the generated partition columns are
+			// only projected (through the extra projections) for data that is passed through to the copy
```
pdet: nit: lets keep comments single line (also in tests) when possible.

pdet (commented):
Hi @jauco, thanks for the PR, I have a couple of comments.

src/storage/ducklake_merge_into.cpp
jauco: A good point. I tried to keep them as similar as possible, but IIRC there were some small differences. I'll check if I can make it a helper function! (also for the second one)

jauco:
@pdet I think I resolved all your requests.

The change got a bit bigger, because I renamed a bunch of variables to be certain that the code was the same and I figured that that's a net improvement. If you want the final patch to be smaller I can revert most of it. (If I revert the DuckLakeMergeSink{Global,Local}State bit then the the helper function signature will get a lot bigger)

I did try and make it seperately reviewable commits

pdet:
Thanks! LGTM!

PR #1255, opened 2026-06-22
Invalidate stale schema entries

dentiny:
Hi team, IIRC, schema cache entries are only invalidated on (1) snapshot expiration; or (2) inline table superseded.
On DDL, new version of schema is loaded into cache, but stale ones are never cleaned and used.

### DIFF
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -39,6 +39,7 @@ struct NewNameMapInfo;
 struct CompactionInformation;
 struct DuckLakePath;
 struct DuckLakeCommitState;
+struct DuckLakeSchemaCacheEntry;
 class DuckLakeFieldId;
 class LocalTableChangeIterationHelper;
 class DuckLakeTransactionState;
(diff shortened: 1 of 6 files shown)


Dletta:
@dentiny Do you have a timeline on when your PR might be able to be merged and we can test if this resolves the memory leak issues?

src/include/storage/ducklake_transaction.hpp
```diff
@@ -348,6 +353,11 @@ class DuckLakeTransaction : public Transaction, public enable_shared_from_this<D
 	idx_t local_catalog_id;
 	//! Set when this transaction inlines into a table that does not yet have an inlined-data table
 	atomic<bool> requires_new_inlined_table {false};
+	//! Schema cache entries referenced by this transaction's catalog entries. These pins keep transaction-local
+	//! parent references valid even if another transaction invalidates the global object-cache entry.
+	//! Maps from cache entry pointer to the shared pointer to the cache entry.
+	mutex schema_pin_lock;
+	unordered_map<DuckLakeSchemaCacheEntry *, shared_ptr<DuckLakeSchemaCacheEntry>> schema_pins;
```
pdet: This looks fairly similar to 

```
 class DuckLakeSchemaPinState {
    mutex lock;
    unordered_map<DuckLakeSchemaCacheEntry *, shared_ptr<DuckLakeSchemaCacheEntry>> pins;
    void Pin(shared_ptr<DuckLakeSchemaCacheEntry> entry); // D_ASSERT; lock; emplace(raw, move)
  };
```

Maybe we can reuse?

pdet (commented):
Thanks, I just have one minor comment.

src/include/storage/ducklake_transaction.hpp
dentiny: Yeah you're right, updated

Dletta:
Sorry to nudge again @pdet. I understand this is open source, so no pressure intended.
I just wanted to see if you had time to approve and merge this one, as the memory leak is a real problem when using ducklake at the moment.

pdet:
Thanks, LGTM!

PR #1256, opened 2026-06-22
Fix compile on clang

dentiny:
I met a compilation error on clang/linux.
```sh
/workspace/src/storage/ducklake_transaction.cpp:1531:16: error: could not convert 'result'
from 'unique_ptr<duckdb::MaterializedQueryResult>' to 'unique_ptr<duckdb::QueryResult>'
  return result;
```

### DIFF
--- a/.github/workflows/MainDistributionPipeline.yml
+++ b/.github/workflows/MainDistributionPipeline.yml
@@ -33,9 +33,51 @@ jobs:
       ci_tools_version: v1.5-variegata
       extension_name: ducklake
 
+  # Regression test, see https://github.com/duckdb/ducklake/pull/1256 for more details.
+  gcc12-compatibility-build:
+    name: Build with GCC 12
+    runs-on: ubuntu-22.04
+    env:
+      VCPKG_TARGET_TRIPLET: x64-linux
+      GEN: Ninja
+      VCPKG_TOOLCHAIN_PATH: ${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
+      CC: gcc-12
+      CXX: g++-12
+      CMAKE_BUILD_PARALLEL_LEVEL: 2
+    steps:
+      - name: Install required ubuntu packages
+        run: |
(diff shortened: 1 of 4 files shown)


qsliu2017:
hey @dentiny, I've encountered this issue and fixed it for a couple time. (https://github.com/duckdb/ducklake/pull/908) but CI uses GCC, and it seems never trigger this warning:

```
# Warn when returning unique_ptr<Derived> as unique_ptr<Base> without std::move
# (Clang-specific; GCC does not need an explicit flag for this pattern)
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wreturn-std-move)
endif()
```

should we consider to strict this compile option in CI?

pdet:
I agree that this would be nice to verify in CI, could we also do this here?

dentiny:
Thank you both @qsliu2017 and @pdet for the followup! Did some experiments on the platform-compiler combinations and confirmed the issue repro on a few.
```md
| Platform             | Compiler        | Result for `return result;` |
|----------------------|-----------------|-----------------------------|
| Ubuntu 18.04 aarch64 | Clang 6         | ❌ Fails                    |
| Ubuntu 20.04 aarch64 | Clang 10        | ❌ Fails                    |
| Ubuntu 22.04 aarch64 | Clang 14        | ✅ Passes                   |
| Ubuntu 22.04 aarch64 | GCC 11          | ❌ Fails                    |
| Ubuntu 22.04 aarch64 | GCC 12          | ❌ Fails                    |
| Ubuntu 24.04 aarch64 | GCC 13          | ✅ Passes                   |
| macOS arm64          | Apple Clang 17  | ✅ Passes                   |
```

Tweeted some compilation options but no luck, so moved to adding extra CI steps as regression test.
In the latest commit, I add an extra build step in the CI, which I have verified to repro the issue before fix: https://github.com/dentiny/ducklake/pull/8

pdet:
Thanks!

PR #1257, opened 2026-06-22
Fixed D_ASSERT tripped by missing SetVectorType

NiclasHaderer:
Set vector to const if all inputs are const.

Fixes https://github.com/duckdblabs/motherduck/issues/550


pdet:
Hi @NiclasHaderer, I believe this is already fixed in `main`, also `main` has a relassert CI run to prevent assertions breaking in the future.

AFAIK, we do not have plans to release a new v1.5, so I just wanted to check if there is a different reason for this PR to target that.

NiclasHaderer:
No, just did not see that it was fixed on main. Sounds good to me.

`main has a relassert CI` ❤️

PR #1258, opened 2026-06-22
Preserve externally added transaction-local files

jdctinuiti:
## Summary
- preserve externally registered transaction-local data files on disk when DuckLake drops them from metadata
- keep deleting DuckLake-created files and delete files during cleanup
- add coverage for delete-to-empty and transaction-local drop-table paths

The `created_by_ducklake flag` is already present on main; this PR applies the remaining guards at the transaction-local physical removal sites exercised by externally added files.


pdet:
Thanks, LGTM!

PR #1260, opened 2026-06-23
Recompute column stats after dropping data files

jdctinuiti:
## Summary

Depends on #1459, which builds on #1458.

After a full-file deletion, recover exact global column statistics from the remaining files when the metadata is sufficient. Otherwise keep the conservative unknown-statistics fallback.

- Carry delete-adjusted row counts for dropped files through local and server-side commit paths so the exactness check uses the post-change live row count.
- Extend the shared statistics reducer to full-file drops, reading only surviving files and including compatible inlined data.
- Cover concurrent retries, same-commit inserts, sparse/skipped statistics, inexact bounds and inlined types with the existing regression cases.

Table-total refresh, snapshot/retry safety and rewrite-stat recovery are separated into #1459.

## Review

[Review only this PR's column-statistics change](https://github.com/jdctinuiti/ducklake/compare/3d4db22f...f8cd2dcf).

The Files tab includes the prerequisites until they merge; the comparison above isolates this change.

## Validation

The local incremental build and six targeted column-statistics SQL tests passed. `make format-fix` and `git diff --check` passed. Quack-specific execution was not rerun locally.


jdctinuiti:
Hey @pdet would you be able to review this PR as well as #1261 whenever you get a chance? Don't mean to rush you just want to make sure they aren't forgotten, thank you!

src/storage/ducklake_multi_file_list.cpp
```diff
@@ -145,6 +149,24 @@ DuckLakeTableEntry &DuckLakeMultiFileList::GetTable() {
...
+	}
+	if (read_info.table.IsTransactionLocal()) {
+		return false;
+	}
+	if (HasTransactionLocalData()) {
+		return false;
+	}
+	auto transaction = read_info.GetTransaction();
+	auto current_snapshot = transaction->GetSnapshot();
+	if (read_info.snapshot.snapshot_id != current_snapshot.snapshot_id || transaction->GetCatalog().CatalogSnapshot()) {
```
pdet, line 165: Can we add the following test

```sql
CREATE TABLE ducklake.t(i INTEGER);
INSERT INTO ducklake.t SELECT range FROM range(100);
INSERT INTO ducklake.t SELECT range FROM range(1000, 1100); 

-- con2
BEGIN;
SELECT COUNT(*) FROM ducklake.t;                 -- 200

-- con1
DELETE FROM ducklake.t WHERE i < 100;            -- drops file A; stats recomputed to [1000,1099]

-- con2 (file A still visible in this transaction!)
SELECT COUNT(*) FROM ducklake.t WHERE i = 50;    -- must return 1
```

src/storage/ducklake_transaction_state.cpp
```diff
@@ -1013,10 +1051,21 @@ string DuckLakeTransactionState::UpdateStatsForDroppedFiles(
...
 		new_globals.initialized = true;
+		auto dropped_stats = attempt_dropped_file_stats.find(table_id)->second;
 		bool delete_column_stats = ApplyDroppedFileStats(table_id, new_globals, attempt_dropped_file_stats);
 		if (delete_column_stats) {
-			// the rows are deleted below - drop them from the update so we do not write values we then delete
-			new_globals.stats.column_stats.clear();
+			vector<DuckLakeFileInfo> no_new_files;
+			DuckLakeNewGlobalStats recomputed_globals;
+			auto expected_data_file_rows =
+			    SubtractDroppedFileStat(context.get_net_data_file_row_count(table_id), dropped_stats.live_row_count);
```
pdet, line 1106: Can we add the following test:
```sql
CREATE TABLE ducklake.t(i INTEGER);
INSERT INTO ducklake.t SELECT range FROM range(100); 
INSERT INTO ducklake.t SELECT range FROM range(1000, 1100); 
INSERT INTO ducklake.t SELECT range FROM range(5000, 5100);  

-- con1
BEGIN;
DELETE FROM ducklake.t WHERE i < 100;              

-- con2
DELETE FROM ducklake.t WHERE i BETWEEN 1000 AND 1099;      

 -- con1
COMMIT;
```

pdet (commented):
Hi, thanks for the PR, two comments!

src/storage/ducklake_transaction_state.cpp
jdctinuiti: Added. Commit retries also now select a fresh base snapshot per attempt and reload stats inside that attempt’s metadata transaction. Recalculation therefore excludes both committed drops rather than reusing the first attempt’s stale stats.

src/storage/ducklake_multi_file_list.cpp
jdctinuiti: Added. The fix for this test keys cached stats by snapshot_id and validates a cache miss against the latest metadata snapshot. An older reader will never consume newer narrowed stats, and if its exact stats are unavailable, it scans without stats pruning.

jdctinuiti:
Hey @pdet , added both regression tests and updated the fix to handle old snapshots and concurrent commit retries correctly.

Longer term, we could version global stats by snapshot in the metadata spec. That would make historical stats directly queryable, but it requires a metadata/spec migration and compatibility work across backends, so I kept that out of this PR. What are your thoughts?

jdctinuiti:
Hi @pdet just removed all merge conflicts with main, I would really appreciate a review on this and #1261 as I've been trying to get this feature set in for the past few months now after following your direction to split it into multiple PRs. If there is a more formal process for doing so please let me know. I can also have my company reach out.

jdctinuiti:
Hi @pdet — I’ve rebased onto main and separated the changes further for review:

- #1458: validate `SET NOT NULL` when NULL statistics are unknown.
- #1459: refresh table totals after file-set changes.
- #1260: recover exact column statistics after full-file drops.

The descriptions include dependencies and isolated review links. #1261 remains separate. I’d appreciate your feedback when you have time.

jdctinuiti:
I split the table-total maintenance out of this PR into #1459 to make review easier, and this PR now builds on it. I'm moving this back to draft until #1459 merges, then I'll rebase it onto main and mark it ready.

PR #1261, opened 2026-06-23
Add metadata deletes for identity partition predicates

jdctinuiti:
Use metadata-only DELETE when the complete predicate can be evaluated from each file's current identity partition values. For a table partitioned by `(source, batch_id)`, a predicate such as `source IN ('a', 'b') AND batch_id = 202601` can drop matching files without scanning Parquet or writing row-delete files.

The planner maps scan/projection references to DuckLake field IDs and uses DuckDB's native expression evaluator, including predicates spanning multiple identity keys. Metadata substitution is limited to types with exact representations. Unsupported types, calendar transforms, unproven row predicates, volatile expressions, old partition specifications, incomplete metadata, inlined rows, and existing row deletes retain the row-deletion fallback. Zero-match deletes still record the attempt for concurrent-write conflict checks.

Statistics recomputation remains in #1260. Calendar-transform support is deferred to a follow-up after this PR merges.

Validation on the unchanged DuckDB pin `1b3c92a18881286356dd526dcbe5408d675d5e63`:

- `make format-fix` passed.
- Delete suite: 715 assertions passed across 26 test cases; one HTTPFS-dependent test skipped.
- Tests cover physical plans, survivor rows, representation-sensitive predicates, schema mapping, prepared execution, transaction-local files, rollback, and concurrent commits.

### DIFF
--- a/src/include/storage/ducklake_delete.hpp
+++ b/src/include/storage/ducklake_delete.hpp
@@ -167,7 +167,9 @@ public:
 	static PhysicalOperator &PlanDelete(ClientContext &context, PhysicalPlanGenerator &planner,
 	                                    DuckLakeTableEntry &table, PhysicalOperator &child_plan,
 	                                    vector<idx_t> row_id_indexes, string encryption_key,
-	                                    bool allow_duplicates = true);
+	                                    bool allow_duplicates = true,
+	                                    unordered_map<idx_t, vector<Value>> metadata_delete_partition_values = {},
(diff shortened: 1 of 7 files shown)


jdctinuiti:
Hey @pdet , I went ahead and refactored this PR. It now uses DuckDB’s native expression evaluator instead of custom predicate extraction, slimming it down by 322 lines while supporting more single-partition-key predicates. Almost half of the PR’s LOC are now tests.

PR #1265, opened 2026-06-24
Fix test warning

dentiny:
I get a ton of warning messages like below
```sh
Replacing deprecated loop iterator ${threadid} in test "test/sql/remove_orphans/concurrent_insert_orphan_cleanup.test" - please use the new loop iterator {threadid}
Replacing deprecated loop iterator ${j} in test "test/sql/remove_orphans/concurrent_insert_orphan_cleanup.test" - please use the new loop iterator {j}
```


pdet:
Thanks!

PR #1268, opened 2026-06-29
Add max_compacted_files option to ducklake_rewrite_data_files

gijshendriksen:
Like `ducklake_merge_adjacent_files`, `ducklake_rewrite_data_files` can be expensive to run if there are many files that need to be rewritten -- especially if files are also merged, rows are sorted, etc. For `ducklake_merge_adjacent_files`, the `max_compacted_files` option can be used to limit the number of compacted files (and thus, memory usage). This PR extends support for this option to `ducklake_rewrite_data_files`, so users can choose to control the compaction scale there as well.


pdet:
Thanks! LGTM!

PR #1272, opened 2026-06-30
Resolve path-like DuckLake secret names before metadata paths

### LINKED ISSUE #1270: ATTACH 'ducklake:<name>' silently treats a secret name containing '/' as a filesystem path
### Summary

When attaching via `ATTACH 'ducklake:<name>'`, DuckLake decides whether `<name>` is a **secret name** or a **metadata path** using `PathIsSecret()`, which only accepts `[A-Za-z0-9_]`. Any other character — `/`, `-`, `.`, `:` — makes it fall through to "treat as metadata path", **even when a `ducklake` secret with that exact name exists**. The result is a confusing `Cannot open file` error and no way to reference such a secret.

### Reproduction (DuckDB v1.5.4)

```sql
CREATE SECRET "a/b" (TYPE ducklake, METADATA_PATH 'metadata.ducklake');
ATTACH 'ducklake:a/b' AS db1;
-- IO Error: Failed to attach DuckLake MetaData ... Cannot open file ".../a/b"

-- Contrast: an alphanumeric/underscore name resolves as a secret:
CREATE SECRET "a_b" (TYPE ducklake, METADATA_PATH 'metadata.ducklake');
ATTACH 'ducklake:a_b' AS db2;  -- OK, resolved as a secret
```

### Expected

Since the secret `"a/b"` exists (and the secret manager stores/looks it up fine — secret names are quoted identifiers that may contain `/`), `ATTACH 'ducklake:a/b'` should resolve it as a secret rather than a path.

### Root cause (main @ `96a4574d`)

- [`src/storage/ducklake_secret.cpp:18-30`](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_secret.cpp#L18) — `PathIsSecret()` returns `false` for any non-`[A-Za-z0-9_]` character.
- [`src/storage/ducklake_storage.cpp:82-93`](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_storage.cpp#L82) — when `PathIsSecret` is false, the value becomes `options.metadata_path` and the secret lookup is never attempted.

Notably there's also no explicit escape hatch: `ATTACH 'ducklake:' (TYPE ducklake, SECRET "a/b")` throws `Unsupported option secret for DuckLake`.

### Why it matters

I'm storing DuckDB secrets in a PostgreSQL secret store with row-level security (RLS) enabled. The secrets are named like `<tenant>/<name>` to go with an RLS scheme. `s3`/`rds`/`postgres` secrets resolve fine with such names (by scope, or by name reference in AWS_RDS_SECRET / METADATA_PARAMETERS). But, `ducklake` secrets become unusable because attach is the only way to reference them by name.

### Suggested fixes (either one helps)

1. In `DuckLakeAttach`, attempt an exact secret-name lookup *first*; only fall back to treating `info.path` as a metadata path if no secret matches.
2. Broaden `PathIsSecret` to accept the characters valid in quoted secret identifiers.

### Environment

DuckDB v1.5.4 (Variegata), ducklake extension; reproduced on macOS arm64.

hutiefang76:
## Summary

`ATTACH ducklake:<name>` now tries an exact DuckLake secret lookup before treating path-like names as metadata paths. This lets quoted secret identifiers such as `tenant/ducklake` resolve correctly instead of being interpreted as filesystem paths.

Fixes #1270.

## Tests

- `PATH="$HOME/Library/Python/3.9/bin:$PATH" python3 duckdb/scripts/format.py --all --check --directories src test`
- `./build/release/test/unittest test/sql/secrets/ducklake_secrets.test`

### DIFF
--- a/src/storage/ducklake_storage.cpp
+++ b/src/storage/ducklake_storage.cpp
@@ -79,17 +79,19 @@ static unique_ptr<Catalog> DuckLakeAttach(optional_ptr<StorageExtensionInfo> sto
 			throw InvalidInputException(
 			    "Default secret was not found - either specify a path to attach to directly, or create the secret");
 		}
-	} else if (DuckLakeSecret::PathIsSecret(info.path)) {
-		// if the path is a plain name - load the secret name
+	} else {
+		// Secret names can be quoted identifiers, so first try an exact lookup before treating path-like names as
+		// paths.
 		secret = DuckLakeSecret::GetSecret(context, info.path);
 		if (!secret) {
-			throw InvalidInputException(
-			    "Secret \"%s\" was not found - if this was meant to be a path to a DuckDB file, use duckdb:%s instead",
-			    info.path, info.path);
+			if (DuckLakeSecret::PathIsSecret(info.path)) {
+				throw InvalidInputException("Secret \"%s\" was not found - if this was meant to be a path to a DuckDB "
+				                            "file, use duckdb:%s instead",
+				                            info.path, info.path);
+			}
+			// otherwise set the remainder of the path as the metadata path
+			options.metadata_path = info.path;
 		}
-	} else {
-		// otherwise set the remainder of the path as the metadata path
-		options.metadata_path = info.path;
 	}
 	if (secret) {
 		// if we have a secret - handle the options
--- a/src/storage/ducklake_storage.cpp
+++ b/src/storage/ducklake_storage.cpp
@@ -82,17 +82,19 @@ static unique_ptr<Catalog> DuckLakeAttach(optional_ptr<StorageExtensionInfo> sto
 			throw InvalidInputException(
 			    "Default secret was not found - either specify a path to attach to directly, or create the secret");
 		}
-	} else if (DuckLakeSecret::PathIsSecret(info.path)) {
-		// if the path is a plain name - load the secret name
+	} else {
+		// Secret names can be quoted identifiers, so first try an exact lookup before treating path-like names as
+		// paths.
 		secret = DuckLakeSecret::GetSecret(context, info.path);
 		if (!secret) {
-			throw InvalidInputException(
-			    "Secret \"%s\" was not found - if this was meant to be a path to a DuckDB file, use duckdb:%s instead",
-			    info.path, info.path);
+			if (DuckLakeSecret::PathIsSecret(info.path)) {
+				throw InvalidInputException("Secret \"%s\" was not found - if this was meant to be a path to a DuckDB "
+				                            "file, use duckdb:%s instead",
+				                            info.path, info.path);
+			}
+			// otherwise set the remainder of the path as the metadata path
+			options.metadata_path = info.path;
 		}
-	} else {
-		// otherwise set the remainder of the path as the metadata path
-		options.metadata_path = info.path;
 	}
 	if (secret) {
 		// if we have a secret - handle the options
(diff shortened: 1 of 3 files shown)


src/storage/ducklake_storage.cpp
```diff
@@ -79,17 +79,19 @@ static unique_ptr<Catalog> DuckLakeAttach(optional_ptr<StorageExtensionInfo> sto
 			throw InvalidInputException(
 			    "Default secret was not found - either specify a path to attach to directly, or create the secret");
 		}
-	} else if (DuckLakeSecret::PathIsSecret(info.path)) {
-		// if the path is a plain name - load the secret name
+	} else {
+		// Secret names can be quoted identifiers, so first try an exact lookup before treating path-like names as
+		// paths.
```
dentiny, line 84: I'm wondering if we should verify the secret type is "ducklake"?

kinghuang: Agreed, the type should be verified. Since `GetSecretByName` matches on name across all secret types, a same-named s3/aws/postgres secret can get picked up here — it then hits the `HandleDuckLakeOption` loop and fails with a misleading `Unsupported option key_id for DuckLake`, or shadows a legitimate metadata path.

I suggest filtering by type in the lookup rather than throw, so a wrong-type match falls through to path handling (`info.path` here is legitimately either a secret name or a metadata path). In `DuckLakeSecret::GetSecret`:

```cpp
auto secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "memory");
if (secret_entry && secret_entry->secret->GetType() == "ducklake") {
    return secret_entry;
}
secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "local_file");
if (secret_entry && secret_entry->secret->GetType() == "ducklake") {
    return secret_entry;
}
return nullptr;
```

(`Identifier` defines `operator==(const char *)`. The default-secret path is unaffected.)

Worth a regression test for the collision in `ducklake_secrets.test`:

```
require httpfs

# a non-ducklake secret with a path-like name must not hijack a metadata path
statement ok
CREATE SECRET "{TEST_DIR}/collision.db" (TYPE S3, KEY_ID 'x', SECRET 'y');

statement ok
ATTACH 'ducklake:{TEST_DIR}/collision.db' AS ducklake_collision

statement ok
CREATE TABLE ducklake_collision.t(i INTEGER);

statement ok
DETACH ducklake_collision
```

Without the filter this errors with `Unsupported option ... for DuckLake`; with it, the s3 secret is ignored and the value is treated as the metadata path.

*Generated by Claude Code.*

dentiny: Curious did you plan to make the code change, or you think it's not worth it?

hutiefang76: I made the change in ff60e6b5. The lookup now accepts a named entry only when its secret type is `ducklake`, for both memory and local-file persistence. I also added a regression case with a same-named HTTP secret; it previously reached DuckLake option parsing and failed with `Unsupported option bearer_token`, and now correctly reports that the DuckLake secret was not found. Verified with the targeted SQL test (37 assertions), the release build, and the repository format check.

hutiefang76: Rebased this onto the current main in 1110734e. Main now resolves named secrets across all registered secret storages, so I kept that behavior and applied the `ducklake` type guard to the returned entry. The wrong-type regression remains covered. A clean release rebuild, the focused secret test (37 assertions), format check, and `git diff --check` all pass.

kinghuang (approved):
LGTM. The check failure on the `Dispatch to pr-tracker` step looks unrelated to this PR.

src/storage/ducklake_storage.cpp
kinghuang: @dentiny Anything else needed to push this over the line?

pdet:
Thanks!

PR #1273, opened 2026-06-30
Resolve naming collision for repeated bucket partitions

oleg-milanov:
Resolves a directory naming collision when partitioning a table by the same column multiple times using the same transform (e.g. `bucket(4, c), bucket(8, c), bucket(16, c)`) 

Changes:
- Updated  `GetPartitionKeyName`  in  `ducklake_partition_data.cpp`  to resolve folder name collisions by incrementing a suffix (producing `bucket_c_2`, `bucket_c_3` , etc.) instead of overwriting the fallback name (`bucket_c`).                                                                                                                                                             
- Added a regression test case  in `add_file_repeated_bucket.test` to verify correct indexing for mixed identity and repeated bucket partition keys.


pdet:
Thanks, LGTM!

PR #1274, opened 2026-07-01
Fix delete positions when flushing partitioned inlined data

### LINKED ISSUE #1266: BUG: incorrect result when flushing partitioned table
### What happens?

### To Reproduce

```sql
memory D ATTACH 'ducklake::memory:' AS db (DATA_PATH ':memory:', DATA_INLINING_ROW_LIMIT 1000000);
memory D USE db;
db D CREATE TABLE t(part INT, i INT);
db D ALTER TABLE t SET PARTITIONED BY (part);
db D INSERT INTO t SELECT i%4, i FROM range(0,400) tbl(i);
db D DELETE FROM t WHERE i%5=0;
db D SELECT count(), sum(i) FROM t;
┌──────────────┬────────┐
│ count_star() │ sum(i) │
│ int64 │ int128 │
├──────────────┼────────┤
│ 320 │ 64000 │
└──────────────┴────────┘
db D CALL ducklake_flush_inlined_data('db');
┌─────────────┬────────────┬──────────────┐
│ schema_name │ table_name │ rows_flushed │
│ varchar │ varchar │ int128 │
├─────────────┼────────────┼──────────────┤
│ main │ t │ 400 │
└─────────────┴────────────┴──────────────┘
db D SELECT count(), sum(i) FROM t; -- EXPECTED 320 | 64000, the same as before flushing
┌──────────────┬────────┐
│ count_star() │ sum(i) │
│ int64 │ int128 │
├──────────────┼────────┤
│ 320 │ 63704 │
└──────────────┴────────┘
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

7e3c8e97cc5acddbcd2a1ebfb8530e6c52efdacf

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

hutiefang76:
Fixes #1266.

When flushing inlined data for a partitioned table, DuckLake writes data files and then creates delete files for rows that were already deleted while the data was still inlined. The delete-position query assumes the flushed file order is `row_id, begin_snapshot`, but partitioned COPY output can write a partition file in a different internal order. That can make the delete file point at the wrong positions and produce an incorrect result after `ducklake_flush_inlined_data`.

This sets COPY `order_columns` for the partitioned flush path, when no table `SORTED BY` order is active, so each partition file is ordered by the embedded row id and snapshot id. That keeps the file order aligned with the existing delete-position calculation.

Added a regression test that reproduces the wrong result with a partitioned table: before flush the visible result is `count=320, sum=64000`; after the fix the result stays the same after flushing.

Local checks:
- `GEN=ninja CMAKE_PREFIX_PATH=/opt/homebrew/opt/croaring make release`
- `PATH="$HOME/Library/Python/3.9/bin:$PATH" python3 duckdb/scripts/format.py --all --check --directories src test`
- `./build/release/test/unittest test/sql/issues/issue_1266.test`
- `./build/release/test/unittest test/sql/deletion_inlining/test_deletion_inlining_partitions.test`
- `./build/release/test/unittest test/sql/delete/deletion_vector_inlined_flush.test`
- `./build/release/test/unittest test/sql/data_inlining/data_inlining_flush_sequential_updates.test`
- selected deletion/partition/compaction/sorted flush regression files

### DIFF
--- a/src/functions/ducklake_flush_inlined_data.cpp
+++ b/src/functions/ducklake_flush_inlined_data.cpp
@@ -20,6 +20,7 @@
 #include "duckdb/planner/operator/logical_filter.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp"
 #include "duckdb/function/function_binder.hpp"
 #include "duckdb/planner/expression/bound_aggregate_expression.hpp"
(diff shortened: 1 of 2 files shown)


hutiefang76:
Rebased this branch onto the current `main` (`27c14caf`) and resolved the `LogicalCopyToFile::names` API conflict in `7f3bdb76`.

The resolution keeps the current upstream identifier handling while preserving the fix: partitioned inlined flushes without a table sort order are ordered by the appended row ID and snapshot ID columns, so delete positions remain aligned with the output files.

Verified locally on Apple Silicon with:

- `make format-fix`
- `make GEN=ninja release` (675 build steps completed)
- `./build/release/test/unittest test/sql/issues/issue_1266.test` (1 test case, 11 assertions, all passed)
- `git diff --check origin/main...HEAD`

PR #1275, opened 2026-07-01
Fix empty files from inlined per-thread inserts

### LINKED ISSUE #1251: Incorrect results with per_thread_output when inserts are absorbed by data inlining
### What happens?

When `per_thread_output = true`, inserts that are absorbed by **data inlining** (on by default, `data_inlining_row_limit = 10`) can lead to incorrect query results. The `CREATE TABLE AS` is fine, but the next inlined write into the table (a single `INSERT` is enough — no `DELETE`/churn needed) leaves the snapshot in a state where reads are wrong. Two shapes show up depending on file layout:

- **Equality-filter pushdown returns wrong rows + garbled values** — the row count is right, but `WHERE part_id = 1` matches every row, `WHERE part_id = 0` matches none, and `SELECT *` returns corrupted values.
- **Missing rows** — after a delete + reinsert cycle, `SELECT count(*)` is lower than the real row count.

In both cases the catalog gains one empty data file per write (`record_count = 0`, NULL column stats). Using `per_thread_output = false` (or `data_inlining_row_limit = 0`) avoids it. Reproduces with `threads` set to both 1 and 4.

Likely related: #525 / #864 / #1025 were the compaction-crash form of the same underlying behavior (a writer inheriting `per_thread_output` from the lake config), addressed in #866 for the compaction path. The INSERT and flush-inlined-data paths still inherit it, which seems to be what surfaces here.

### To Reproduce

Minimal — wrong filter results + garbled values, no churn needed:

```sql
ATTACH 'ducklake:/tmp/repro.ducklake' AS lake;
USE lake;
CALL lake.set_option('per_thread_output', true);
-- data_inlining_row_limit defaults to 10, so these tiny writes are inlined

CREATE TABLE t AS SELECT i AS part_id, i*10 AS v FROM range(4) tt(i);   -- fine
INSERT INTO t BY NAME SELECT 1 AS part_id, 99 AS v;                     -- inlined

SELECT count(*) FROM t;                     -- 5 (count is correct)
SELECT count(*) FROM t WHERE part_id = 1;   -- expected 2, got 5 (matches everything)
SELECT count(*) FROM t WHERE part_id = 0;   -- expected 1, got 0
SELECT part_id, v FROM t ORDER BY part_id, v;
-- expected (0,0),(1,10),(1,99),(2,20),(3,30)
-- got      (1,99),(1,266),(2,276),(3,286),(256,256)
```

Variant — missing rows, via delete + reinsert:

```sql
-- (fresh DB, per_thread_output=true as above)
CREATE TABLE t AS SELECT i AS part_id, i*10 AS v FROM range(4) tt(i);
DELETE FROM t WHERE part_id = 0;
INSERT INTO t BY NAME SELECT 0 AS part_id, 0 AS v;
DELETE FROM t WHERE part_id = 0;
INSERT INTO t BY NAME SELECT 0 AS part_id, 0 AS v;

SELECT count(*) FROM t;                      -- expected 4, got 1

-- the empty files that get registered (record_count = 0):
SELECT data_file_id, record_count, row_id_start
FROM __ducklake_metadata_lake.ducklake_data_file
ORDER BY data_file_id;
```

### Root Cause

The insert plan is `DuckLakeInsert <- PhysicalCopyToFile <- DuckLakeInlineData <- scan`.

1. `DuckLakeInsert::GetCopyOptions` (`src/storage/ducklake_insert.cpp`) sets, for non-partitioned writes, `write_empty_file = false` (line 578) and `per_thread_output` from the lake config (line 592). `PlanCopyForInsert` builds the `PhysicalCopyToFile` operator directly.

2. This combination isn't reachable from a normal SQL `COPY` — the binder rejects it in `duckdb/src/planner/binder/statement/bind_copy.cpp` (`"Can't combine WRITE_EMPTY_FILE false with PER_THREAD_OUTPUT"`). Since DuckLake constructs the physical operator directly, that path isn't exercised.

3. For a small insert, `DuckLakeInlineData` (`src/storage/ducklake_inline_data.cpp`) collects all rows into inlined storage and forwards none downstream, so `PhysicalCopyToFile` receives 0 rows.

4. With `per_thread_output` and 0 rows, DuckDB's COPY finalize step writes an empty schema-only file even though `write_empty_file = false` — the branch commented `// no rows from source, write schema to file` in `duckdb/src/execution/operator/persistent/physical_copy_to_file.cpp` (`PhysicalCopyToFile::Finalize` on duckdb main; `FinalizeInternal` on the 1.5.x line).

5. `DuckLakeInsert::AddWrittenFiles` (`src/storage/ducklake_insert.cpp:118`) then registers that empty file. The `row_count == 0` skip currently lives only in the flush branch (`set_snapshot_id == true`, line 221, added in `cb9d17ae` "skip empty files"); the INSERT / CTAS path doesn't have it. The registered empty file takes a `row_id` range and carries zeroed column stats, which overlaps the inlined rows — which seems to produce the garbled values, wrong pruning, and missing rows.

### Suggested Fix

One option that resolves the incorrect results is to extend the existing empty-file skip so it also covers the INSERT / CTAS path — skip `row_count == 0` files in `AddWrittenFiles` for all callers (the flush path already does this since `cb9d17ae`):

```cpp
data_file.row_count = chunk.GetValue(1, r).GetValue<idx_t>();
if (data_file.row_count == 0) {
    continue;
}
```

This leaves the empty ~121-byte schema-only parquet on disk as an unreferenced file (the COPY operator still writes it); avoiding the write itself would need either honoring `write_empty_file = false` in DuckDB's `per_thread_output` path, or not pairing the two options in DuckLake.

I'd be glad to open a PR if this looks like the right direction — or take one of the alternatives above instead if you'd prefer. Let me know which you'd want.

### OS:

macOS 26.5.1, arm64

### DuckDB Version:

1.5.4 (also reproduced on current main)

### DuckLake Version:

released `d318a545`; also current `main` and `v1.5-variegata`

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Ilia Ablamonov

### Affiliation:

Rick.ai

- [x] Yes, I have

- [x] Yes, I have

hutiefang76:
This fixes #1251.

When a small INSERT is fully absorbed by data inlining, the downstream COPY can still report a zero-row schema-only file when per_thread_output is enabled. The INSERT path then registered that empty file in the catalog, which could overlap the inlined rows and make filter pruning return incorrect results.

The fix keeps zero-row file metadata for normal INSERT/compaction paths, but filters schema-only zero-row COPY output in DuckLakeInsert::Finalize when DuckLakeInlineData has already absorbed rows into inlined storage. This preserves merge-adjacent/compaction behavior while avoiding bogus live data-file metadata for inlined inserts.

Verification:
- added test/sql/data_inlining/per_thread_output_inline_insert.test
- watched the new test fail before the fix: part_id = 1 returned 5 rows instead of 2
- added a metadata assertion that active ducklake_data_file count stays 0 after the inlined per-thread insert
- ./build/release/test/unittest "test/sql/data_inlining/per_thread_output_inline_insert.test"
- ./build/release/test/unittest "test/sql/compaction/repro_merge_adjacent_zero_output.test"
- ./build/release/test/unittest "test/sql/compaction/compaction_per_thread_output.test"
- ./build/release/test/unittest "test/sql/data_inlining/inlining_issue_on_empty_inline.test"
- ./build/release/test/unittest "test/sql/data_inlining/basic_data_inlining.test"
- ./build/release/test/unittest "test/sql/data_inlining/data_inlining_filter.test"
- ./build/release/test/unittest "test/sql/data_inlining/data_inlining_flush.test"
- PATH=$HOME/Library/Python/3.9/bin:$PATH python3 duckdb/scripts/format.py --all --check --directories src test
- git diff --check

Current GitHub workflow runs on the latest fork head are action_required and need maintainer approval before they can execute.

### DIFF
--- a/src/storage/ducklake_insert.cpp
+++ b/src/storage/ducklake_insert.cpp
@@ -34,6 +34,19 @@
 
 namespace duckdb {
 
+static bool PerThreadOutputEnabled(DuckLakeCatalog &catalog, DuckLakeTableEntry &table) {
+	string per_thread_output_str;
+	return catalog.TryGetConfigOption("per_thread_output", per_thread_output_str, table) &&
+	       per_thread_output_str == "true";
+}
+
+static bool IsSchemaOnlyInlinedInsertOutput(const DuckLakeInsertGlobalState &global_state,
+                                            const DuckLakeDataFile &file) {
+	// DuckLakeInlineData has already absorbed rows into inlined storage, so a zero-row
+	// file reported here is the schema-only COPY output produced by per_thread_output.
+	return global_state.total_insert_count > 0 && file.row_count == 0;
+}
+
 DuckLakeInsert::DuckLakeInsert(PhysicalPlan &physical_plan, const vector<LogicalType> &types, DuckLakeTableEntry &table,
                                optional_idx partition_id, string encryption_key_p)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::EXTENSION, types, 1), table(&table), schema(nullptr),
@@ -253,6 +266,17 @@ SinkFinalizeType DuckLakeInsert::Finalize(Pipeline &pipeline, Event &event, Clie
                                           OperatorSinkFinalizeInput &input) const {
 	auto &global_state = input.global_state.Cast<DuckLakeInsertGlobalState>();
 
+	auto &catalog = global_state.table.catalog.Cast<DuckLakeCatalog>();
+	if (PerThreadOutputEnabled(catalog, global_state.table)) {
+		// Do not register schema-only COPY output as a live data file after inlining.
+		// Zero-row files from non-inlined inserts are still preserved for merge_adjacent_files.
+		global_state.written_files.erase(std::remove_if(global_state.written_files.begin(),
+		                                                global_state.written_files.end(),
+		                                                [&](const DuckLakeDataFile &file) {
+			                                                return IsSchemaOnlyInlinedInsertOutput(global_state, file);
+		                                                }),
+		                                 global_state.written_files.end());
+	}
 	for (auto &data_file : global_state.written_files) {
 		global_state.total_insert_count += data_file.row_count;
 	}
--- a/test/sql/data_inlining/per_thread_output_inline_insert.test
+++ b/test/sql/data_inlining/per_thread_output_inline_insert.test
@@ -1,0 +1,59 @@
+# name: test/sql/data_inlining/per_thread_output_inline_insert.test
+# description: test inlined inserts with per_thread_output enabled
+# group: [data_inlining]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION {TEST_DIR}/{UUID}.db
+
+test-env DATA_PATH {TEST_DIR}
+
+statement ok
+ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/per_thread_output_inline_insert', DATA_INLINING_ROW_LIMIT 10)
+
+statement ok
+USE ducklake
+
+statement ok
+CALL ducklake.set_option('per_thread_output', 'true')
+
+statement ok
+CREATE TABLE t AS SELECT i AS part_id, i * 10 AS v FROM range(4) tt(i)
+
+query I
+INSERT INTO t BY NAME SELECT 1 AS part_id, 99 AS v
+----
+1
+
+query II
+SELECT count(*), COALESCE(sum(record_count), 0)
+FROM __ducklake_metadata_ducklake.ducklake_data_file
+WHERE end_snapshot IS NULL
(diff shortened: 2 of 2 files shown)


hutiefang76:
Resolved the merge conflict and synced this branch with main at 7b52aea8. The PR diff remains limited to the insert fix and its SQL regression test.

Validation on macOS with the pinned DuckDB revision 1b3c92a1:
- `make reldebug` and `make format-fix` completed successfully.
- The SQL regression fails without the insert fix: metadata contains two active zero-row files instead of zero.
- With the fix restored, all 20 assertions pass. I rebuilt and reran the test after the final main sync; all 20 assertions still pass.

Test command: `build/reldebug/test/unittest 'test/sql/data_inlining/per_thread_output_inline_insert.test'`.

This is focused local validation; the full test suite and upstream CI have not been verified for this update.

PR #1285, opened 2026-07-09
fix: prevent concurrent inlined and parquet deletes of the same row from both committing (#1215)

### LINKED ISSUE #1215: BUG: concurrent inlined and parquet deletes on the same row commit without conflict
### What happens?

Two concurrent transactions that delete the **same row of the same data file** can both commit when their delete sizes straddle `data_inlining_row_limit`. One delete is small (≤ the limit), so its tombstone is written as an *inlined* file delete; the other is large (> the limit), so its tombstone is written as a *parquet delete file*. The commit conflict check does not treat these two as conflicting, so both transactions succeed and the row's position ends up recorded in **both** stores.

For comparison, two concurrent deletes of the same row *do* correctly conflict when both land in the same store (parquet+parquet or inlined+inlined) — only the cross-store (inlined+parquet) combination slips through, in either commit order.

Once the overlap exists, the read path makes the deleted rows reappear in scans. This appears to be a **write-path origin of the overlap** behind #1084 — that issue (and PR #1208) addresses the read-time symptom, while this report is about the overlap being creatable through ordinary concurrent SQL in the first place.

### To Reproduce

```python
import duckdb, tempfile, os

ws = tempfile.mkdtemp(prefix="ducklake-overlap-")
db = duckdb.connect()
db.execute("INSTALL ducklake; LOAD ducklake")
# inlining row limit 10: a DELETE of <=10 rows is stored as inlined file deletes,
# a DELETE of >10 rows is stored as a parquet delete file
db.execute(f"ATTACH 'ducklake:{ws}/catalog.ducklake' AS dl (DATA_PATH '{ws}/data', DATA_INLINING_ROW_LIMIT 10)")
# 100 rows -> parquet data file (100 > inlining limit)
db.execute("CREATE TABLE dl.t AS SELECT i AS id FROM range(100) t(i)")

# two independent transactions on the same catalog
c1 = db.cursor()
c2 = db.cursor()
c1.execute("BEGIN")
c2.execute("BEGIN")

# con1: single-row delete (1 <= 10) -> inlined file delete for the row at position 50
c1.execute("DELETE FROM dl.t WHERE id = 50")
# con2: 30-row delete incl. value 50 (30 > 10) -> parquet delete file covering positions 50..79
c2.execute("DELETE FROM dl.t WHERE id >= 50 AND id < 80")

c1.execute("COMMIT")
c2.execute("COMMIT")  # deletes the same row (50) concurrently -> should conflict, but does not

# both tombstones now exist for position 50: one inlined, one in a parquet delete file
print(db.execute("SELECT count(*) FROM dl.t WHERE id >= 50 AND id < 80").fetchone()[0])  # expect 0
```

**Observed:** `c2` commits with no conflict, and the final count is `29` (rows 51–79 reappear) instead of `0`.

**Expected:** two concurrent deletes that both target row `50` should conflict, the same way `parquet+parquet` and `inlined+inlined` deletes of the same row already do; the second commit should be rejected.

In `DuckLakeTransaction::CheckForConflicts` (`src/storage/ducklake_transaction.cpp`), the file-level delete conflict check only runs when **both** transactions are in `tables_deleted_from` (i.e. both wrote parquet delete files). A transaction that wrote *inlined* deletes lands in `tables_deleted_inlined`, which is only checked at table granularity and is never compared against the other transaction's parquet `deleted_from_files` (nor vice-versa). So:

| transaction A | transaction B | result |
| --- | --- | --- |
| parquet delete | parquet delete | conflict (correct) |
| inlined delete | inlined delete | conflict (correct) |
| inlined delete | parquet delete | **both commit → overlap** |
| parquet delete | inlined delete | **both commit → overlap** |

Symmetric in both commit orders. (On a SQLite catalog the second writer is blocked by `database is locked` before reaching this check, so it only manifests on catalog backends that let both transactions reach commit, e.g. DuckDB.)

### OS:

Linux (WSL2)

### DuckDB Version:

v1.5.3 (also reproduced on a source build of duckdb `main`)

### DuckLake Version:

extension bundled with DuckDB 1.5.3; also ducklake `main`

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Gavin Brand

### Affiliation:

N/A

Tested on the latest `core_nightly` (ducklake extension [`424eaa47`](https://github.com/duckdb/ducklake/commit/424eaa47)) — the bug **still reproduces** (both transactions commit; the deleted rows then reappear on read). Also reproduced on a stable release (DuckDB 1.5.3) and a source build of ducklake `main`.

Yes

- [x] Yes, I have

- [x] Yes, I have

GavinPizza:
Closes #1215

Two concurrent transactions deleting the same row of the same data file could both commit when their delete sizes straddle `data_inlining_row_limit`: one tombstone is written as an inlined file delete, the other as a parquet delete file, and `CheckForConflicts` only compared parquet deletes against parquet deletes. This is the write-path origin of the overlap read back in #1084.

The fix extends the file-level conflict check to span both tombstone stores, so cross-store deletes of the same file conflict (in either commit order) while deletes to different files still commit.

On Postgres, the check for the other transaction's (lazily-created) inlined-delete table reads the live catalog via `postgres_query`, since DuckDB's cached view of an attached catalog can be stale for a table another transaction just created.

The new test is skipped on the SQLite catalog. SQLite allows only one writer, so the two transactions cannot both reach commit and the scenario is unreachable there rather than unfixed. It joins the sibling concurrency tests already listed for that reason.

Related read-path fix: #1208 (for #1084). It should land first, since this change makes #1208's current test construction conflict; I'll update that test in whichever PR merges second.

### DIFF
--- a/src/include/metadata_manager/postgres_metadata_manager.hpp
+++ b/src/include/metadata_manager/postgres_metadata_manager.hpp
@@ -37,6 +37,8 @@ public:
 
 	unique_ptr<QueryResult> Query(DuckLakeSnapshot snapshot, string &query) override;
 
+	string InlinedDeleteTableExistsQuery(const string &table_name) const override;
+
 protected:
 	string GetLatestSnapshotQuery() const override;
 	string GenerateFileColumnStatsCTEBody(const CTERequirement &req, TableIndex table_id) override;
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -175,6 +175,11 @@ public:
 
 	virtual string MetadataExistsQuery() const;
 
+	//! Query that returns >=1 row iff a lazily-created inlined-file-delete table exists (cross-store conflict check).
+	//! Overridden per backend so the check reads live catalog state even when DuckDB's view of an attached
+	//! catalog is stale for cross-transaction DDL.
+	virtual string InlinedDeleteTableExistsQuery(const string &table_name) const;
+
 	//! Initialize a new DuckLake
 	virtual void InitializeDuckLake(bool has_explicit_schema, DuckLakeEncryption encryption);
 	//! Get the CREATE TABLE statements for all metadata tables
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -176,6 +176,8 @@ public:
 
 	virtual string MetadataExistsQuery() const;
 
+	virtual string InlinedDeleteTableExistsQuery(const string &table_name) const;
+
 	//! Initialize a new DuckLake
 	virtual void InitializeDuckLake(bool has_explicit_schema, DuckLakeEncryption encryption);
 	//! Get the CREATE TABLE statements for all metadata tables
--- a/src/include/storage/ducklake_transaction_state.hpp
+++ b/src/include/storage/ducklake_transaction_state.hpp
@@ -26,6 +26,9 @@ struct DuckLakeColumnSchemaEntry {
 struct DuckLakeCommitContext {
 	//! Runs a metadata-DB query during conflict resolution.
 	std::function<unique_ptr<QueryResult>(string)> conflict_query_executor;
+	//! Backend-specific query returning >=1 row iff the named inlined-file-delete table exists (see
+	//! DuckLakeMetadataManager::InlinedDeleteTableExistsQuery). Used by the cross-store conflict check.
+	std::function<string(const string &table_name)> inlined_delete_exists_query;
 	//! Returns the latest snapshot for the first commit attempt.
 	std::function<DuckLakeSnapshot()> get_snapshot;
 	//! Executes the batched snapshot/changes SQL against the metadata DB.
@@ -123,10 +126,12 @@ public:
 
 	SnapshotAndStats CheckForConflicts(DuckLakeSnapshot transaction_snapshot,
 	                                   const TransactionChangeInformation &changes,
-	                                   const std::function<unique_ptr<QueryResult>(string)> &executor);
+	                                   const std::function<unique_ptr<QueryResult>(string)> &executor,
+	                                   const std::function<string(const string &)> &inlined_delete_exists_query);
 	void CheckForConflicts(const TransactionChangeInformation &changes, const SnapshotChangeInformation &other_changes,
 	                       DuckLakeSnapshot transaction_snapshot,
-	                       const std::function<unique_ptr<QueryResult>(string)> &executor) const;
+	                       const std::function<unique_ptr<QueryResult>(string)> &executor,
+	                       const std::function<string(const string &)> &inlined_delete_exists_query) const;
 
 	static SnapshotDeletedFromFiles
 	GetFilesDeletedOrDroppedAfterSnapshot(const std::function<unique_ptr<QueryResult>(string)> &executor);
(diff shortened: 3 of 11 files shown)


principalwater:
Built this on top of `main` (736646ee) together with #1208 and exercised both. Three things worth reporting.

**The branch needs two one-line fixes to compile.** `MetadataSchemaName()` returns `const Identifier &` and its string conversion is explicit, so both `SQLLiteralToString` calls in `postgres_metadata_manager.cpp` fail. Line 101 is already fixed on `main` and a rebase would bring it along; line 134 is new here and needs it either way. Opened GavinPizza/ducklake#1 against this branch with both.

**Your merge-order note is right, and this is what it looks like in practice.** With both PRs applied, `issue_1215_concurrent_inlined_parquet_delete_conflict.test` passes 65/65, while `issue_1084_parquet_inlined_delete_overlap.test` fails 1 of 10 — its `COMMIT` at line 60 now raises exactly what this PR is designed to raise:

```
Transaction conflict - attempting to delete from file with index "0"
 - but another transaction has deleted from it
```

#1208's test constructs the overlap by racing two transactions that delete from the same file, which is precisely what this change forbids. So once this lands, that test needs a different way to reach the state it verifies — writing the catalog rows directly rather than racing.

**Independent confirmation that this fixes a real corruption, not just a theoretical one.** On stock (DuckDB 1.5.5, ducklake `d8a1881e`, Postgres catalog), `DATA_INLINING_ROW_LIMIT 10`, 100 rows in a Parquet data file, `con1` deleting `i = 50` (1 row, inlined) while `con2` deletes `i ∈ [50, 80)` (30 rows, parquet delete file):

| | stock | main + #1208 + this |
|---|---|---|
| `con2` COMMIT | succeeds | conflicts |
| `count(*)` from metadata | 69 | — |
| actual scan | 99 | 99, consistent |
| correct answer | 70 | — |

On stock the two read paths disagree with each other and both disagree with reality, in opposite directions. Catalog afterwards: `record_count` 100, `delete_count` 30, one inlined mark — row 50 counted as deleted twice while 29 genuinely deleted rows stay visible.

We hit this in production, unpartitioned, at a much smaller scale; details in duckdb/ducklake#1364.

GavinPizza:
@principalwater Thank you for the fixes and the production repro.

Both fixes are on the branch now and rebased onto current main (3d8e24a8).

Your merge-order finding matches what I expected. This change forbids exactly the race #1208's test uses to build the overlap, so that test needs to construct the state directly in the catalog instead of racing for it. This will be done for whichever PR ends up merging second.

principalwater:
Rebuilt the rebased branch from scratch and re-ran the suite. Everything checks out.

**Build.** `540f6a3b` on top of `3d8e24a8`, clean build on macOS arm64 (Apple Silicon), Release + Ninja. Both `SQLLiteralToString` call sites survived the move to `src/metadata_manager/postgres_metadata_manager.cpp` correctly — lines 101 and 130 now go through `GetIdentifierName()`, and nothing else needed touching. No warnings from the changed files.

**Regression test.** `issue_1215_concurrent_inlined_parquet_delete_conflict.test` — 65 assertions, all pass, same as before the rebase.

**Full suite.** 534 test cases / 19 513 assertions, all pass.

One caveat that is worth knowing but has nothing to do with this change: on a default macOS shell `ulimit -n` is 256, and `compaction/merge_adjacent_tiered.test_slow` fails there with

```
IO Error: Cannot open file "...ducklake-<uuid>.parquet": Too many open files
```

It creates 512 files upfront and then merges them. With `ulimit -n 8192` it passes (531 assertions). Might be worth a note in the test or the contributing docs — it reads as a real failure until you look at the message.

**Housekeeping.** Since both fixes are on your branch now, GavinPizza/ducklake#1 is superseded and I have closed it. Thanks for carrying them over with authorship intact.

Happy to re-run any of this after the test rework, whichever way the merge order lands.

Issue #239, opened 2025-06-20
ducklake attach using sql init file fails to query table


hmeriann:
This issue is copied from the [duckdb-java](https://github.com/duckdb/duckdb-java) issue tracker. [Original issue](https://github.com/duckdb/duckdb-java/issues/284)

> this is what i use for my ducklake init file:
> ```
> ATTACH 'ducklake:postgres:postgresql://user:pwd@127.0.0.1:5432/dbname' as ducklake;
> /* DUCKDB_CONNECTION_INIT_BELOW_MARKER */
> USE ducklake;
> SET s3_endpoint='localhost:9000';
> SET s3_access_key_id='minioadmin';
> SET s3_secret_access_key='minioadmin';
> SET s3_url_style='path';
> SET s3_use_ssl=false;
> the connection was successful on dbeaver but when i tried to do a query on a table it lead to this error:
> ```
> SQL Error: IO Error: Cannot open file "s3:\\ducklake-data\main\test_table\ducklake-01978af3-3251-736e-acbf-a319ff3c42be.parquet": The filename, directory name, or volume label syntax is incorrect.
> as can be seen, it turns the s3 path seperator from forward slash to backslash.

Comment from Alex Kasko:

> The problem there is not JDBC-related, reproducible with duckdb.exe on Windows with local Minio (with [default settings](https://min.io/docs/minio/windows/index.html)) and the SQL snippet from the issue.

staticlibs:
Root cause: something is wrong with autoloading `httpfs` extension on Windows.

Reproducing step with `1.3.1`:

1. run Minio with default settings:

```
minio server C:\apps\minio\data
```

2. login to `http://127.0.0.1:9000` with `minioadmin/minioadmin` and create a bucket with name `ducklake-data`

3. create DuckLake with:

```
INSTALL httpfs;
LOAD httpfs;
ATTACH 'ducklake:postgres:postgresql://wilton:wilton@127.0.0.1:5432/lake_minio' AS ducklake (DATA_PATH 's3://ducklake-data');
```

4. create a table in DuckLake:

```
LOAD httpfs;
ATTACH 'ducklake:postgres:postgresql://wilton:wilton@127.0.0.1:5432/lake_minio' AS ducklake;
USE ducklake;
SET s3_endpoint='localhost:9000';
SET s3_access_key_id='minioadmin';
SET s3_secret_access_key='minioadmin';
SET s3_url_style='path';
SET s3_use_ssl=false;
[...]

PR #1286, opened 2026-07-09
Fix: Cloud Path Separator corruption when httpfs not loaded

### LINKED ISSUE #1032: ATTACH without DATA_PATH and https corrupts cloud path separator
### What happens?

Not sure if this can be called a bug, since it’s more of a slightly non-standard scenario, but it might still be worth investigating.

When a DuckLake catalog has a cloud DATA_PATH (e.g. gs://) and is re-opened with ATTACH without the DATA_PATH specified and without the `httpfs` extension already loaded, DuckLake permanently caches `\` as the path separator for the session. Every file path it subsequently constructs replaces all `/` with `\`, turning `gs://bucket/path/schema/table/file.parquet` into `gs:\\bucket\path\schema\table\file.parquet`. Windows then rejects this as a malformed local path before any GCS call is even attempted.

The metadata table itself (ducklake_metadata) still shows the originally stored gs:// value with forward slashes - the corruption is purely in-memory, which makes it invisible to a simple health-check query.

Note: Does not reproduce on DuckDB v1.4.4 + DuckLake v0.3 (commit 3f1b372).

### To Reproduce

```
import duckdb, tempfile, os

tmp = tempfile.mktemp(suffix=".duckdb")
try:
    # Step 1 — create catalog with a gs:// data path (httpfs loaded)
    con = duckdb.connect()
    con.execute("LOAD ducklake; LOAD httpfs;")
    con.execute(f"ATTACH 'ducklake:{tmp}' AS lake (DATA_PATH 'gs://my-bucket/my-data/')")
    con.close()

    # Step 2 — re-attach WITHOUT httpfs and WITHOUT DATA_PATH
    con = duckdb.connect()
    con.execute("LOAD ducklake;")                    # httpfs intentionally absent
    con.execute(f"ATTACH 'ducklake:{tmp}' AS lake")  # DATA_PATH intentionally omitted

    # Step 3 — any operation that writes a file exposes the corrupted separator
    con.execute("CREATE TABLE lake.t AS SELECT 1 AS id")
finally:
    con.close()
    for p in [tmp, tmp + ".wal"]:
        if os.path.exists(p): os.remove(p)
```

Actual result:

```
IO Error: Failed to create directory
  "gs:\\my-bucket\my-data\main\t":
  The filename, directory name, or volume label syntax is incorrect.
```
Maintenance operations, i.e. `CHECKPOINT`, also fail.

### OS:

Windows

### DuckDB Version:

1.5.2

### DuckLake Version:

1.0

### DuckDB Client:

Python, CLI

### Hardware:

_No response_

### Full Name:

Vitaliy Grinyuk

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

rmblk:
Fixes #1032
Reimplements the fix for #239 by reapplying the same change from #244.
The change helps ensure relevant extensions such as httpfs are loaded on ATTACH to prevent the default file system path separators from being applied (i.e. Windows '\' separators used in a s3 path, corrupting the data path value).

### DIFF
--- a/src/include/storage/ducklake_partition_data.hpp
+++ b/src/include/storage/ducklake_partition_data.hpp
@@ -30,6 +30,10 @@ struct DuckLakePartitionField {
 
 struct DuckLakePartition {
 	idx_t partition_id = 0;
+	//! The transaction-local id this partition was assigned when it was created. Data files written in
+	//! the same transaction reference this id. Commit attempts overwrite partition_id with the id of
+	//! that attempt, so on retry the data-file remap must key off the original transaction-local id.
+	optional_idx local_partition_id;
 	vector<DuckLakePartitionField> fields;
 };
 
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -189,8 +189,14 @@ void DuckLakeInitializer::LoadExistingDuckLake(DuckLakeTransaction &transaction)
 		}
 		if (tag.key == "data_path") {
 			if (options.data_path.empty()) {
-				options.data_path = metadata_manager.LoadPath(tag.value);
+				// Proper path extensions may not be loaded
+				options.data_path = tag.value;
+				// Autoloads relevant extensions if the data path uses a known pattern (e.g. starts with "s3://")
 				InitializeDataPath();
+				// LoadPath may modify the path (e.g. by adding/modifying a separator)
+				// On Windows, if the proper extension is not loaded, the path may be set to the wrong separator
+				// InitializeDataPath needs to be called before LoadPath to set the correct separator
+				options.data_path = metadata_manager.LoadPath(tag.value);
 			} else {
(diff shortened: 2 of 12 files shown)


dentiny (commented):
LG, meanwhile I'm wondering if you could
- Add a regression test to repro, so we don't break it again in the future?
- Could you please rebase? I believe there were some windows-related issues

rmblk:
Rebase to main branch? And should I create a new folder in the tests for this or add to one like 'initialize' perhaps?

dentiny:
> Rebase to main branch?

Yes, rebase / merge with latest v1.5 (since your PR compares against v1.5); it's a compilation issue on windows and I met it in the past.

> And should I create a new folder in the tests for this or add to one like 'initialize' perhaps?

If you don't find a proper directory, you could also put it here https://github.com/duckdb/ducklake/tree/main/test/sql/issues

PR #1292, opened 2026-07-11
Validate partition when adding data files

### LINKED ISSUE #1291: BUG: incorrect query result when hive partition doesn't conform DuckLake table partition
### What happens?

### To Reproduce

```sql
memory D ATTACH 'ducklake:/private/tmp/dl_bucket_invalid_repro.db' AS dl (
           DATA_PATH '/private/tmp/dl_bucket_invalid_repro_data',
           METADATA_CATALOG 'metadata',
           DATA_INLINING_ROW_LIMIT 0
         );
memory D USE dl;
dl D CREATE TABLE IF NOT EXISTS t1(user_id VARCHAR, value INT);

-- Computed partition value is 0, 1, 2, or 3.
dl D ALTER TABLE t1 SET PARTITIONED BY (bucket(4, user_id));

-- Generated subdirectory is bucket=999
dl D COPY (
       SELECT 'alice'::VARCHAR AS user_id, 42::INT AS value, 999::INT AS bucket
     ) TO '/private/tmp/dl_bucket_invalid_external' (FORMAT parquet, PARTITION_BY(bucket));
dl D CALL ducklake_add_data_files(
       'dl',
       't1',
       '/private/tmp/dl_bucket_invalid_external/bucket=999/*.parquet',
       hive_partitioning => true
     );
┌──────────┐
│ filename │
│ varchar  │
└──────────┘
   0 rows 

-- Full scan does contain `alice`.
dl D SELECT 'full_scan' AS q, user_id, value FROM t1 ORDER BY ALL;
┌───────────┬─────────┬───────┐
│     q     │ user_id │ value │
│  varchar  │ varchar │ int32 │
├───────────┼─────────┼───────┤
│ full_scan │ alice   │    42 │
└───────────┴─────────┴───────┘

-- Query by `alice` returns empty result.
dl D SELECT 'filtered' AS q, user_id, value FROM t1 WHERE user_id = 'alice' ORDER BY ALL;
┌─────────┬─────────┬───────┐
│    q    │ user_id │ value │
│ varchar │ varchar │ int32 │
└─────────┴─────────┴───────┘
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [ ] Yes, I have

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1291

The bug is: 
- When `ducklake_add_data_files` imports Hive-partitioned Parquet files, transformed partition directory values were accepted as trusted metadata even when they were invalid for DuckLake’s partition spec
- Later, bucket partition pruning uses DuckLake’s partition metadata to choose which files to scan. A filter such as user_id = 'alice' is folded to Alice’s real bucket value, so the file recorded as bucket=999 is pruned away and the query can return incorrect results.
- This PR rejects invalid imported bucket partition values during `ducklake_add_data_files`

### DIFF
--- a/test/sql/add_files/add_file_partitioned.test
+++ b/test/sql/add_files/add_file_partitioned.test
@@ -192,3 +192,25 @@ FROM multi_partition_t WHERE day(dt) = 20 order by id;
 ----
 3	c	2024-04-20
 12	z	2024-04-20
+
+# Bucket transform partition values must be valid bucket ids for the configured bucket count.
+statement ok
+CREATE TABLE bucket_partition_t(user_id VARCHAR, value INTEGER);
+
+statement ok
+ALTER TABLE bucket_partition_t SET PARTITIONED BY (bucket(4, user_id));
+
+statement ok
+COPY (
+  SELECT 'alice'::VARCHAR AS user_id, 42::INT AS value, 999::INT AS bucket
+) TO '{DATA_PATH}/add_file_partitioned/invalid_bucket' (FORMAT parquet, PARTITION_BY(bucket));
+
+statement error
+CALL ducklake_add_data_files(
+  'ducklake',
+  'bucket_partition_t',
+  '{DATA_PATH}/add_file_partitioned/invalid_bucket/bucket=999/*.parquet',
+  hive_partitioning => true
+)
+----
+contains an invalid partition value for the table configuration
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -114,6 +114,21 @@ struct HivePartition {
 	optional_idx partition_key_index;
 };
 
+static bool IsValidTransformedHivePartitionValue(const HivePartition &hive_partition,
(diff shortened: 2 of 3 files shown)


test/sql/add_files/add_file_partitioned.test
```diff
@@ -192,3 +192,25 @@ FROM multi_partition_t WHERE day(dt) = 20 order by id;
 ----
 3	c	2024-04-20
 12	z	2024-04-20
+
+# Bucket transform partition values must be valid bucket ids for the configured bucket count.
```
pdet, line 196: Can we also add a test with a `NULL` value, including a re-add?

pdet (commented):
Thanks, just one comment.

test/sql/add_files/add_file_partitioned.test
dentiny: NULL handle and test added

PR #1297, opened 2026-07-11
Validate parquet field ids on data ingestion

### LINKED ISSUE #1296: BUG: adding parquet with dropped column succeeds
### What happens?

Hi team, when adding parquet files into DuckLake tables, even if a column has been dropped, the parquet file ingestion still succeeds.

### To Reproduce

```sql
# name: test/sql/add_files/add_files_field_id_drop_readd.test
# description: ducklake_add_data_files should reject stale Parquet field IDs after nested drop/re-add
# group: [add_files]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (
	DATA_PATH '{DATA_PATH}/ducklake_add_files_field_id_drop_readd',
	METADATA_CATALOG 'metadata',
	DATA_INLINING_ROW_LIMIT 10
)

statement ok
CREATE TABLE ducklake.test(s STRUCT(i INT, j INT));

statement ok
INSERT INTO ducklake.test VALUES ({'i': 1, 'j': 10});

statement ok

statement ok
SET VARIABLE old_files = (SELECT LIST(data_file) FROM ducklake_list_files('ducklake', 'test'));

# Byte-copy the DuckLake-written Parquet file so the old field IDs are preserved.
statement ok
COPY (
	SELECT content
	FROM read_blob(getvariable('old_files')[1])
) TO '{DATA_PATH}/ducklake_add_files_old_field_ids.parquet' (FORMAT BLOB);

query TI
SELECT name, field_id
FROM parquet_schema('{DATA_PATH}/ducklake_add_files_old_field_ids.parquet')
WHERE name IN ('s', 'i', 'j')
ORDER BY name
----
i	2
j	3
s	1

statement ok
ALTER TABLE ducklake.test DROP COLUMN s.j;

statement ok
ALTER TABLE ducklake.test ADD COLUMN s.j INT;

statement ok
INSERT INTO ducklake.test VALUES ({'i': 2, 'j': 20});

statement ok

query I rowsort
SELECT s FROM ducklake.test
----
{'i': 1, 'j': NULL}
{'i': 2, 'j': 20}

query TIT
SELECT column_name, column_id, CASE WHEN end_snapshot IS NULL THEN 'live' ELSE 'dropped' END
FROM metadata.ducklake_column
WHERE column_name = 'j'
ORDER BY column_id
----
j	3	dropped
j	4	live

statement error
CALL ducklake_add_data_files(
	'ducklake',
	'test',
	'{DATA_PATH}/ducklake_add_files_old_field_ids.parquet',
	ignore_extra_columns => true
);
----
field ID
```
gets error message
```sh
1. test/sql/add_files/add_files_field_id_drop_readd.test:77
Query unexpectedly succeeded! (test/sql/add_files/add_files_field_id_drop_readd.test:77)!
CALL ducklake_add_data_files(
        'ducklake',
        'test',
        'duckdb_unittest_tempdir/59371/ducklake_add_files_old_field_ids.parquet',
        ignore_extra_columns => true
);
filename
VARCHAR
[ Rows: 0]
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1296

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -150,6 +150,9 @@ private:
 	                                                    vector<unique_ptr<ParquetColumn>> &parquet_columns,
 	                                                    const vector<unique_ptr<DuckLakeFieldId>> &field_ids,
 	                                                    const string &prefix = string());
+	void CollectLiveFieldIds(const vector<unique_ptr<DuckLakeFieldId>> &field_ids, unordered_set<idx_t> &result);
+	void ValidateParquetFieldIds(const ParquetFileMetadata &file, const vector<unique_ptr<ParquetColumn>> &columns,
+	                             const unordered_set<idx_t> &live_field_ids, const string &prefix = string());
 	void MapColumnStats(ParquetFileMetadata &file_metadata, DuckLakeDataFile &result);
 	unique_ptr<DuckLakeNameMapEntry> MapHiveColumn(ParquetFileMetadata &file_metadata, const DuckLakeFieldId &field_id,
 	                                               const Value &hive_value);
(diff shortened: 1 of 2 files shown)


src/functions/ducklake_add_data_files.cpp
```diff
@@ -1164,6 +1167,36 @@ DuckLakeFileProcessor::MapColumns(ParquetFileMetadata &file_metadata,
...
+                                                    const unordered_set<idx_t> &live_field_ids, const string &prefix) {
+	for (auto &column : columns) {
+		const auto full_name = prefix.empty() ? column->name : StringUtil::Format("%s.%s", prefix, column->name);
+		if (!IsDuckLakeInternalColumn(column->name) && column->field_id.IsValid()) {
+			const auto source_field_id = column->field_id.GetIndex();
+			if (live_field_ids.find(source_field_id) == live_field_ids.end()) {
+				throw InvalidInputException(
+				    "Parquet field ID mismatch for column \"%s\" in file \"%s\": field ID %d is not a live field in "
+				    "table \"%s\"",
+				    full_name, file.filename, source_field_id, table.name.GetIdentifierName());
```
pdet, line 1193: What if the names match perfectly?
e.g.,

```sql
CREATE TABLE ducklake.t(i INTEGER, j INTEGER);
COPY (SELECT 84 i, 42 j) TO 'ext.parquet' (FIELD_IDS {i: 100, j: 101});
CALL ducklake_add_data_files('ducklake','t','ext.parquet');
```

pdet (commented):
Thanks for the PR, one comment.

src/functions/ducklake_add_data_files.cpp
dentiny: Yeah I feel like it's intended -- when adding parquet files to DuckLake, if field id specified, it should match perfectly with DuckLake metadata, which is used to guard against concurrent changes.
```sql
memory D ATTACH 'ducklake::memory:' AS ducklake (DATA_PATH '/tmp/ducklake_repro1');
memory D CREATE TABLE ducklake.t(i INTEGER, j INTEGER);
memory D COPY (SELECT 84 i, 42 j) TO 'ext.parquet' (FIELD_IDS {i: 100, j: 101});
memory D CALL ducklake_add_data_files('ducklake','t','ext.parquet');
Invalid Input Error:
Parquet field ID mismatch for column "i" in file "ext.parquet": field ID 100 is not a live field in table "t"
```
and it should work when field id unassigned, which I expect should be most of the case
```sql
memory D COPY (SELECT 84 i, 42 j) TO 'ext1.parquet';
memory D CALL ducklake_add_data_files('ducklake','t','ext1.parquet');
┌──────────┐
│ filename │
│ varchar  │
└──────────┘
   0 rows 
```

pdet:
Thanks!

PR #1301, opened 2026-07-11
Accept empty version on snapshot expiration

### LINKED ISSUE #1300: BUG: snapshot expiration throws parser exception on empty versions
### What happens?

See repro SQL, should be straightforward.

### To Reproduce

```sql
1. test/sql/compaction/expire_snapshots_empty_versions.test:24
Query unexpectedly failed (test/sql/compaction/expire_snapshots_empty_versions.test:24)
 (test/sql/compaction/expire_snapshots_empty_versions.test:24)!
SELECT count(*)
FROM ducklake_expire_snapshots('ducklake', versions => [], dry_run => true);
Actual result:
Parser Error: Failed to get snapshot information from DuckLake: syntax error at or near ")"

LINE 5: ...ta_ducklake"."main".ducklake_snapshot) AND snapshot_id IN ()
                                                                      ^

LINE 2: FROM ducklake_expire_snapshots('ducklake', versions => [], dry_run...
             ^
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/ducklake/issues/1300

Another solution I thought about is to throw invalid input exception on empty version list, but I don't think it's invalid to do so.

### DIFF
--- a/test/sql/compaction/expire_snapshots_empty_versions.test
+++ b/test/sql/compaction/expire_snapshots_empty_versions.test
@@ -1,0 +1,33 @@
+# name: test/sql/compaction/expire_snapshots_empty_versions.test
+# description: expiring an empty list of snapshot versions is a no-op
+# group: [compaction]
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION {TEST_DIR}/{UUID}.db
+
+test-env DATA_PATH {TEST_DIR}
+
+statement ok
+ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (
+    DATA_PATH '{DATA_PATH}/ducklake_expire_empty_versions'
+)
+
+statement ok
+CREATE TABLE ducklake.test(i INTEGER)
+
+statement ok
+INSERT INTO ducklake.test VALUES (42)
+
+query I
+SELECT count(*)
+FROM ducklake_expire_snapshots('ducklake', versions => [], dry_run => true)
+----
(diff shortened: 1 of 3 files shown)


test/sql/compaction/expire_snapshots_empty_versions.test
```diff
@@ -0,0 +1,33 @@
+# name: test/sql/compaction/expire_snapshots_empty_versions.test
```
pdet, line 1: Hm, since we are touching this, can we also add checks and sanitization for `NULL` values?

e.g.,
* `FROM ducklake_expire_snapshots('ducklake', versions => NULL);`
* `FROM ducklake_expire_snapshots('ducklake', versions => [<id>], dry_run => NULL);`
* FROM ducklake_expire_snapshots('ducklake', older_than => NULL, dry_run => true);
* [NULL] / [1, NULL, 2] for versions

pdet (commented):
Thanks for the PR, I only have one small - adjacent comment

test/sql/compaction/expire_snapshots_empty_versions.test
dentiny: Make sense, added more checks in the function params.

pdet:
Thanks!

Issue #683, opened 2026-01-09
Transactional DDL puts the Ducklake catalog in an unusable state


diegoimbert:
### What happens?

Some transactional DDL queries make the ducklake unusable with the following error on any query : `Catalog Error: Column with name id already exists!` 

Note that this does not happen if you remove the BEGIN and END. This also does not happen if the CREATE and ALTERs are in the same transaction.

I am using :
- Latest duckdb-rs client (1.4.3)
- PostgreSQL 16.11 as a catalog
- AWS S3 as storage (probably irrelevant as no row was inserted, S3 was never hit)

### To Reproduce

Run this:
```sql
ATTACH 'ducklake:postgres:conn_string' AS dl (DATA_PATH '...');
USE dl;

CREATE TABLE message (
  id INT NOT NULL,
  user_id INT NOT NULL,
  content TEXT,
  created_at TIMESTAMP NOT NULL
);

BEGIN;

ALTER TABLE message ALTER COLUMN user_id TYPE BIGINT;
ALTER TABLE message ALTER COLUMN user_id SET DEFAULT 123;

END;
```

Now, any attempt to attach to the ducklake and then use it will fail.
Minimal example:
```sql
ATTACH 'ducklake:postgres:conn_string' AS dl (DATA_PATH '...');
USE dl;
```

### OS:

Mac OS (M4, ARM 64)

### DuckDB Version:

1.4.3

### DuckLake Version:

f7e92c6

### DuckDB Client:

Rust

### Full Name:

Diego Imbert

### Affiliation:

Windmill

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #1306, opened 2026-07-13
Accept typed-literal and cast constant defaults in ADD COLUMN (#1290)

### LINKED ISSUE #1290: BUG: ADD COLUMN with a typed-literal or cast constant default is rejected as non-literal
### What happens?

`ALTER TABLE ... ADD COLUMN c <type> DEFAULT <literal>` on a DuckLake table throws a `NotImplementedException` when the default is a typed literal or a cast instead of a plain constant. All of these fail:

```sql
DEFAULT DATE '2020-01-01'
DEFAULT TIMESTAMP '2020-01-01 00:00:00'
DEFAULT '5'::INTEGER
DEFAULT 'abc'::VARCHAR
```

Each of these is a constant value, and regular DuckDB accepts all of them as a column default on a non-DuckLake table. A plain literal default such as `DEFAULT 5` or `DEFAULT 'abc'` also works on a DuckLake table. Only the typed-literal and `::TYPE` spellings are rejected. This matters because many clients and ORMs write defaults out in the `::TYPE` form, so they cannot add a column with a default at all. Nothing is corrupted. A valid statement is just refused.

### To Reproduce

```python
import duckdb
import tempfile

ws = tempfile.mkdtemp(prefix="ducklake-adddefault-")
db = duckdb.connect()
db.execute("INSTALL ducklake; LOAD ducklake")
db.execute(f"ATTACH 'ducklake:{ws}/c.ducklake' AS dl (DATA_PATH '{ws}/d')")
db.execute("CREATE TABLE dl.t (id INTEGER)")
db.execute("INSERT INTO dl.t VALUES (1), (2)")

db.execute("ALTER TABLE dl.t ADD COLUMN d DATE DEFAULT DATE '2020-01-01'")
print(db.execute("SELECT id, d FROM dl.t ORDER BY id").fetchall())
```

Output:

```
_duckdb.NotImplementedException: Not implemented Error: We cannot add a column with a non-literal default value. Add the column and then explicitly set the default for new values using "ALTER ... SET DEFAULT"
```

**Observed:** the `ALTER ... ADD COLUMN` fails.

**Expected:** the column is added and the existing rows are back-filled, giving `[(1, datetime.date(2020, 1, 1)), (2, datetime.date(2020, 1, 1))]`. That is what regular DuckDB returns for the same statement on a non-DuckLake table.

All four spellings fail the same way, each tested on a column of the matching type (DATE, TIMESTAMP, INTEGER, VARCHAR). The plain literal defaults work for comparison: `INTEGER DEFAULT 5` gives `[(1, 5), (2, 5)]` and `VARCHAR DEFAULT 'abc'` gives `[(1, 'abc'), (2, 'abc')]`. So the problem is limited to the typed-literal and cast spellings.

Also reproducible from the DuckDB CLI client:

```sql
ATTACH 'ducklake:/tmp/c.ducklake' AS dl (DATA_PATH '/tmp/d');
CREATE TABLE dl.t (id INTEGER);
INSERT INTO dl.t VALUES (1), (2);
ALTER TABLE dl.t ADD COLUMN d DATE DEFAULT DATE '2020-01-01';
```

```
Not implemented Error: We cannot add a column with a non-literal default value. Add the column and then explicitly set the default for new values using "ALTER ... SET DEFAULT"
```

`ExtractInitialValue` in `src/storage/ducklake_field_data.cpp` checks `initial_expr->GetExpressionType() != ExpressionType::VALUE_CONSTANT` (line 62 @ `b2f5f88f`). The parser represents a typed literal (`DATE '2020-01-01'`) and a cast (`'5'::INTEGER`) as an `OPERATOR_CAST` around a `VALUE_CONSTANT`, not as a bare `VALUE_CONSTANT`, so the check rejects them even though the value is constant. `DuckLakeFieldData::SetDefault` has the same `!= VALUE_CONSTANT` check on the add-column path (line 335 @ `b2f5f88f`).

One way to fix it: fold a constant default that is wrapped in a single cast naming the column's own type, by reading the inner `VALUE_CONSTANT` and casting it to the column type. It should still reject `TRY_CAST` (which returns NULL on failure rather than erroring), a cast whose type differs from the column (that can change the value, e.g. `'3.9'::DOUBLE` on an `INTEGER` column rounds to 4 while `'3.9'::INTEGER` errors), and nested casts like `'3.9'::DOUBLE::INTEGER`.

### OS:

### DuckDB Version:

v1.5.4 (stable release; also reproduced on a source build @ `0cd9e82950`, `v1.6.0-dev8546`)

### DuckLake Version:

Released extension `d318a545` on DuckDB v1.5.4 (also reproduced on ducklake `main` source build @ `b2f5f88f8df36b229898e4a5184bc86e5bfe9b4f`)

### DuckDB Client:

Python (also reproduced via the CLI client)

### Hardware:

Not performance-related.

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [X] Yes, I have

- [X] Yes, I have

GavinPizza:
Closes #1290.

`ALTER TABLE ... ADD COLUMN c <type> DEFAULT <literal>` on a DuckLake table was rejected with `NotImplementedException("... non-literal default value ...")` whenever the default was a typed literal (`DATE '2020-01-01'`, `TIMESTAMP '...'`) or a cast-decorated constant (`'5'::INTEGER`, `'abc'::VARCHAR`). A bare literal such as `DEFAULT 5` was accepted.

All of these spellings are compile-time constants and stock DuckDB accepts them on a non-DuckLake table, but the parser wraps the typed and cast forms in an `OPERATOR_CAST` over a `VALUE_CONSTANT` rather than a bare `VALUE_CONSTANT`, so the `!= VALUE_CONSTANT` guard refused them. Clients and ORMs that write defaults in the `::TYPE` form could not add a column with a default at all.

## Changes

- Add a `TryFoldConstantDefault` helper in `src/storage/ducklake_field_data.cpp` that folds a constant default optionally wrapped in a single cast decoration. It unwraps one cast layer whose target names the column's own type, reads the inner `VALUE_CONSTANT`, and casts the value to the column type.
- Route `ExtractInitialValue` and the add-column check in `DuckLakeFieldData::SetDefault` through the helper instead of the bare `!= VALUE_CONSTANT` gate.
- Keep the guard closed for defaults that are not a plain constant fold, which still raise the non-literal error: `TRY_CAST` (returns NULL on failure), a cast whose decoration names a different type than the column (can change the value, for example `'3.9'::DOUBLE` on an `INTEGER` column rounds to 4 while `'3.9'::INTEGER` errors), and nested casts such as `'3.9'::DOUBLE::INTEGER`.

## Testing

- New test `test/sql/issues/issue_add_column_cast_default.test` covers the four accepted spellings, back-fill of existing rows (both still inlined and flushed to parquet), a new insert that omits the column picking up the folded default, persistence across `ducklake_flush_inlined_data` and a DETACH/re-ATTACH catalog reload, and the four rejected negatives.
- `test/sql/default/add_column_with_default.test` gains a block pinning the persisted metadata of a cast default.
- Verified the new test fails on `main` before the change (the `DATE` line throws) and passes after.


src/storage/ducklake_field_data.cpp
```diff
@@ -54,20 +55,55 @@ static unique_ptr<ParsedExpression> ExtractDefaultExpression(optional_ptr<const
 	return default_expr->Copy();
 }
 
+static bool TryFoldConstantDefault(const ParsedExpression &expr, const LogicalType &type, Value &result) {
```
pdet: Cant we reuse functionality already  existing like the `ExpressionExecutor::TryEvaluateScalar`, you will need to get the binder, and whatnot but would be much cleaner.

pdet (commented):
Thanks for the PR, I left one comment

src/storage/ducklake_field_data.cpp
GavinPizza: Yep that is the correct call and much cleaner than what I did. I have updated my PR!

Thank you for taking time to look it over and pointing it out to me :)

GavinPizza: One Config Tests failure on this branch: `test/sql/transactions/test_ddl_after_dml.test`. On main this test skips itself by hitting the non-literal default error (whitelisted in skip_error_messages) that this PR fixes. With this fix it now runs far enough to hit a pre-existing bug: ADD COLUMN + RENAME in one transaction corrupts the catalog. Repro on current main:
```sql
ATTACH 'ducklake:x.db' AS dl; USE dl;
CREATE TABLE t2 (id INT);
BEGIN;
ALTER TABLE t2 ADD COLUMN c0 INT DEFAULT 7;
ALTER TABLE t2 RENAME TO t3;
COMMIT;
SELECT * FROM t3;    -- Catalog Error: Column with name id already exists!
```

Other instances of the multi-ALTER-per-transaction family I could find: #683 has the same symptom (fixed for the ALTER TYPE + SET DEFAULT pair in #714), and the still-open #740 is adjacent (ADD + RENAME COLUMN with inlining).

Since it's unrelated to the default folding here, I'd file it as a separate issue. For this PR, is adding the test to skip_tests in `attach_ducklake.json` with a reference to the new issue the right move?

GavinPizza: This turned out to be #1332, fixed by #1339 which is now in main, so no skip_tests was needed.

PR #1312, opened 2026-07-14
Route Top-N file stats through backend-specific CTEs

bill-ph:
Inspired by #1147 and extending the fix (filter pushdown to metadata manager instead of duckdb join which triggers metadata table scans) further.

Route Top-N `min_value`/`max_value` reads through the existing generated column-stats CTEs. Static and dynamic requirements for the same column are merged into one CTE, so PostgreSQL reuses the selective `postgres_query` path while other catalogs retain the attached metadata scan.

Adds regression coverage for same-column static/Top-N filters and PostgreSQL metadata SQL routing.

Tested `topn_file_pruning.test` and `filter_pushdown.test` with the default and PostgreSQL configurations.

### DIFF
--- a/src/common/ducklake_util.cpp
+++ b/src/common/ducklake_util.cpp
@@ -331,7 +331,28 @@ string DuckLakeUtil::JoinPath(FileSystem &fs, const string &a, const string &b) 
 }
 
 shared_ptr<DynamicFilterData> DuckLakeUtil::GetOptionalDynamicFilterData(const TableFilter &filter) {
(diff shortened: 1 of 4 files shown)


pdet:
Hi @bill-ph, thanks for the PR. Could you retarget it to `main`?

bill-ph:
done @pdet

bill-ph:
bump @pdet

pdet:
Thanks!

PR #1313, opened 2026-07-14
Cache the negative result in DuckLakeCatalog::GetTableStats

wmTJc9IK0Q:
A table with no stats at the current snapshot (an empty table, or one that did
not exist at that snapshot) returned nullptr without storing anything in the
object cache. Every subsequent call therefore missed the cache and re-ran
GetGlobalTableStats(), which is a full query against the metadata catalog.

DuckLakeTableEntry::GetStatistics() is called once per referenced column during
planning, so an empty table pays one metadata query per column, on every query,
forever. Measured on a 12-column table with an otherwise identical schema:

    EXPLAIN SELECT <12 cols> FROM <empty ducklake table>   ~100ms  (never warms)
    EXPLAIN SELECT <12 cols> FROM <same table, 1 row>      ~5ms

The cost is linear in the number of columns projected (~4.5ms/column) and never
warms up, because nothing is ever cached. Native DuckDB tables plan in ~0ms.

Caching the negative result fixes it: ~100ms -> ~6ms for the empty table.

This bites hardest on a fresh deployment, where every lake table is empty until
the first flush, and permanently for any signal a deployment does not use.


pdet:
Thanks for the PR, can you fix the format and keep comments to max one line (and only if they are absolutely necessary)

src/storage/ducklake_catalog.cpp
```diff
@@ -820,7 +824,16 @@ shared_ptr<DuckLakeTableStats> DuckLakeCatalog::GetTableStats(DuckLakeTransactio
 	}
 
 	if (!table_stats) {
-		// Table had no stats row for this snapshot (or did not exist at the snapshot)
+		// Table had no stats row for this snapshot (e.g. it is empty, or did not
```
dentiny: Is the comment really necessary?

wmTJc9IK0Q:
@pdet @dentiny - I cleaned up the comments

pdet:
Hi @wmTJc9IK0Q, seem that CI is failing. Can you have a look at that?

wmTJc9IK0Q:
Merged main and fixed the format failure. No functional change.

pdet:
Thanks!

PR #1316, opened 2026-07-17
Clear up unnecessary inlined tables

pdet:


pdet:
This got merged into v1.5 instead

PR #1317, opened 2026-07-18
Fix transform partition stats

### LINKED ISSUE #1298: BUG: query stats for missing column crashes
### What happens?

### To Reproduce

```sql
# name: test/sql/add_files/add_files_transform_partition_missing_column_stats.test
# description: transformed partition values must not become source-column stats when the source column is missing
# group: [add_files]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/ducklake_transform_missing_stats')

statement ok
CREATE TABLE ducklake.test(id INTEGER, dt DATE, payload VARCHAR)

statement ok
ALTER TABLE ducklake.test SET PARTITIONED BY (year(dt))

# Partitioning the staging output by year creates year=2024 in the path while
# leaving the partition key out of the Parquet file. The source column dt is
# intentionally absent as well.
statement ok
COPY (
    SELECT 1 AS id, 'a' AS payload, 2024 AS year
) TO '{DATA_PATH}/ducklake_transform_missing_stats_staging' (
    FORMAT parquet,
    PARTITION_BY (year)
)

statement ok
CALL ducklake_add_data_files(
    'ducklake',
    'test',
    '{DATA_PATH}/ducklake_transform_missing_stats_staging/year=2024/*.parquet',
    allow_missing => true
)

# The missing source column reads as NULL.
query IIII
SELECT COUNT(*), COUNT(dt), MIN(dt), MAX(dt) FROM ducklake.test
----
1	0	NULL	NULL
```
gets exception
```sh
1. test/sql/add_files/add_files_transform_partition_missing_column_stats.test:42
Query unexpectedly failed (test/sql/add_files/add_files_transform_partition_missing_column_stats.test:42)
 (test/sql/add_files/add_files_transform_partition_missing_column_stats.test:42)!
SELECT COUNT(*), COUNT(dt), MIN(dt), MAX(dt) FROM ducklake.test;
Actual result:
Invalid Input Error: Failed to cast value: invalid date field format: "2024", expected format is (YYYY-MM-DD)
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

oleg-milanov:
Closes [#1298](https://github.com/duckdb/ducklake/issues/1298)

Transformed partition values (e.g. `year=2024`) were being written as min/max stats for the source column when the source column was missing from the file.

The change is to only set min/max for `IDENTITY` transforms, for `YEAR/MONTH/DAY/HOUR` the source column reads as `NULL` (`BUCKET` is already ignored).

**Additional fix: identity partitions take precedence over transforms**

When a column is partitioned by both an identity and a transform (`PARTITIONED BY (dt, year(dt))`), both entries share the same field_index. `emplace` doesn't overwrite, so the transform entry (processed first) left the column with null_count = row_count and the pruner dropped the file for `IS NOT NULL` filters. Identity entries now overwrite via `insert_or_assign`, and transform entries are skipped if stats already exist.

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -1136,11 +1136,11 @@ void DuckLakeFileProcessor::MapColumnStats(ParquetFileMetadata &file_metadata, D
 		column_stats.has_num_values = true;
 		column_stats.num_values = file_metadata.row_count.GetIndex();
 		column_stats.has_null_count = true;
-		if (!hive_value.IsNull()) {
+		if (entry.transform.type == DuckLakeTransformType::IDENTITY && !hive_value.IsNull()) {
 			column_stats.min = column_stats.max = hive_value.ToString();
 			column_stats.has_min = column_stats.has_max = true;
 		} else {
-			// All rows in this file have NULL for this partition column
+			// for non-identity transforms the partition value is the transform output, not a value of the source column
 			column_stats.null_count = file_metadata.row_count.GetIndex();
 			column_stats.any_valid = false;
 		};
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -1130,13 +1130,19 @@ void DuckLakeFileProcessor::MapColumnStats(ParquetFileMetadata &file_metadata, D
 		auto &field_type = entry.field_type;
 		auto &hive_value = entry.hive_value;
 
+		// the identity entry must take precedence: it reflects the actual column data
+		const auto is_identity = entry.transform.type == DuckLakeTransformType::IDENTITY;
+		if (!is_identity && result.column_stats.find(field_index) != result.column_stats.end()) {
+			continue;
+		}
+
 		DuckLakeColumnStats column_stats(field_type);
 		// num_values and null_count both needed to write count
 		// metadata in DuckLakeColumnStatsInfo::FromColumnStats
 		column_stats.has_num_values = true;
 		column_stats.num_values = file_metadata.row_count.GetIndex();
 		column_stats.has_null_count = true;
-		if (entry.transform.type == DuckLakeTransformType::IDENTITY && !hive_value.IsNull()) {
+		if (is_identity && !hive_value.IsNull()) {
 			column_stats.min = column_stats.max = hive_value.ToString();
 			column_stats.has_min = column_stats.has_max = true;
 		} else {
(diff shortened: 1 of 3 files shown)


src/functions/ducklake_add_data_files.cpp
```diff
@@ -1130,11 +1130,11 @@ void DuckLakeFileProcessor::MapColumnStats(ParquetFileMetadata &file_metadata, D
 		column_stats.has_num_values = true;
 		column_stats.num_values = file_metadata.row_count.GetIndex();
 		column_stats.has_null_count = true;
-		if (!hive_value.IsNull()) {
+		if (entry.transform.type == DuckLakeTransformType::IDENTITY && !hive_value.IsNull()) {
```
pdet: Can you add the following as a test:

```sql
ATTACH 'ducklake:meta.db' AS ducklake (DATA_PATH 'data');
CREATE TABLE ducklake.test(id INTEGER, dt DATE);
ALTER TABLE ducklake.test SET PARTITIONED BY (dt, year(dt));
  
COPY (SELECT 1 AS id, DATE '2024-06-15' AS dt, 2024 AS year)
TO 'staging' (FORMAT parquet, PARTITION_BY (dt, year));

CALL ducklake_add_data_files('ducklake', 'test',
'staging/dt=2024-06-15/year=2024/*.parquet', allow_missing => true);

SELECT COUNT(*) FROM ducklake.test;                       -- 1
SELECT COUNT(*) FROM ducklake.test WHERE dt IS NOT NULL;  -- 1

```

pdet (commented):
Thanks for the PR, I have one comment

src/functions/ducklake_add_data_files.cpp
oleg-milanov: Thanks for the suggestion! I added the test. While adding it, it revealed a related but separate issue: when a column is partitioned by both an identity and a transform (e.g. PARTITIONED BY (dt, year(dt))), the transform entry was processed first and emplace left dt with null_count = row_count. That made the stats-based pruner drop the file for WHERE dt IS NOT NULL (returned 0).

I fixed it in the same MapColumnStats loop by letting identity entries take precedence over transform entries for the same source column (since the identity value is an actual value of the column). Does that approach sound reasonable to you, or would you prefer it handled differently? Happy to split it into its own PR if you prefer.

oleg-milanov:
CI note: all failing CI jobs fail at the same location - `ducklake_add_data_files.cpp` at the `file.filename` reference in `ValidateParquetFieldIds` - which is code this PR does not touch.

The same error is currently failing on main's CI (line 1230 there; 1240 in this PR only because of the lines added above it), confirming it's a pre-existing issue unrelated to these changes.

Update: all CI jobs pass now, after the fix in main

PR #1322, opened 2026-07-20
Replace debug-only assertions with runtime bounds checks in deletion vector parsing

arijitroy003:
## Summary

- `FromBlob()` in `ducklake_deletion_vector.cpp` uses `D_ASSERT` for all blob/data boundary validation when parsing deletion vectors from puffin files. These assertions compile out in release builds, allowing crafted inputs to cause out-of-bounds reads.
- Replace all six `D_ASSERT` calls with `InvalidInputException` throws that are active in all build configurations.
- Add underflow guards before each `vector_size` subtraction to prevent unsigned integer wraparound.
- Validate `amount_of_bitmaps` is non-negative before use in `reserve()` and the parsing loop.
- Check `bitmap_size` does not exceed remaining `vector_size` before deserialization.
- Use `NumericCast` for the `reserve()` call to safely convert `int64_t` to `idx_t`.

## Test plan

- [ ] Existing deletion vector round-trip tests continue to pass (valid blobs parse identically)
- [ ] Truncated blobs now throw `InvalidInputException` in release builds instead of reading out-of-bounds
- [ ] Negative bitmap count in crafted blob throws instead of causing undefined behavior in `reserve()`

### DIFF
--- a/src/storage/ducklake_deletion_vector.cpp
+++ b/src/storage/ducklake_deletion_vector.cpp
@@ -64,45 +64,76 @@ unique_ptr<DuckLakeDeletionVectorData> DuckLakeDeletionVectorData::FromBlob(data
 	auto vector_size = Load<uint32_t>(blob_start);
 	vector_size = BSwap(vector_size);
 	blob_start += sizeof(uint32_t);
-	D_ASSERT(blob_start < blob_end);
+	if (blob_start >= blob_end) {
+		throw InvalidInputException("Deletion vector blob truncated after vector_size field");
+	}
 
 	auto checksummed_data_start = blob_start;
 	auto memcmp_res = memcmp(DELETION_VECTOR_MAGIC, blob_start, 4);
 	blob_start += 4;
+	if (vector_size < 4) {
+		throw InvalidInputException("Deletion vector vector_size too small for magic bytes");
+	}
 	vector_size -= 4;
-	D_ASSERT(blob_start < blob_end);
+	if (blob_start >= blob_end) {
+		throw InvalidInputException("Deletion vector blob truncated after magic bytes");
+	}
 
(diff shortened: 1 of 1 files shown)


arijitroy003:

src/storage/ducklake_deletion_vector.cpp
```diff
@@ -64,45 +64,76 @@ unique_ptr<DuckLakeDeletionVectorData> DuckLakeDeletionVectorData::FromBlob(data
...
 	auto checksummed_data_length = blob_start - checksummed_data_start;
+	if (blob_start + sizeof(uint32_t) > blob_end) {
+		throw InvalidInputException("Deletion vector blob truncated before checksum");
+	}
 	auto stored_checksum = BSwap(Load<uint32_t>(blob_start));
 	blob_start += sizeof(uint32_t);
-	D_ASSERT(blob_start == blob_end);
+	if (blob_start != blob_end) {
+		throw InvalidInputException("Deletion vector blob has %lld unexpected trailing bytes",
+		                            static_cast<int64_t>(blob_end - blob_start));
```
dentiny:
```suggestion
NumericCast<int64_t>(blob_end - blob_start));
```
which checks against overflow and underflow

dentiny (commented):
Thanks! I just have one minor comment.

I have one followup items left on my own: I think it's be nice to have data corruption exception type, we're now using IO exception.

arijitroy003:
Applied the NumericCast suggestion — thanks for the review!

arijitroy003:
Fixed the build — added the missing `#include "duckdb/common/operator/numeric_cast.hpp"` for `NumericCast`. CI should pass now.

pdet:
Thanks!

PR #1325, opened 2026-07-20
Fix incorrect error messages and remove dead code

arijitroy003:
## Summary

- **Fix copy-paste error in view query error message**: The error handler for the `ducklake_view` query in `LoadCatalog` reported "Failed to get partition information" instead of "Failed to get view information", making it difficult to diagnose view-related failures.
- **Replace FIXME placeholder with descriptive error**: The `UpdateStats` default branch threw `"FIXME: unsupported nested type"` — replaced with a message that includes the actual type name for easier debugging.
- **Remove dead code**: Removed an unused `GetDatabase` forward declaration, two duplicate `#include` directives (`sql_identifier.hpp`, `bound_constant_expression.hpp`), and an unreachable `sqlite_scanner` check that was already handled by the `metadata_managers` map lookup above it.

## Test plan

- [ ] Existing test suite passes (no behavioral changes, only error messages and dead code removal)
- [ ] Verified the `sqlite_scanner` entry exists in the `metadata_managers` map, confirming the removed check was unreachable
- [ ] Verified the removed `GetDatabase` forward declaration has no callers in this file (line 146 calls a member function on the catalog object, not the free function)

### DIFF
--- a/src/storage/ducklake_inline_data.cpp
+++ b/src/storage/ducklake_inline_data.cpp
@@ -295,7 +295,7 @@ void UpdateStats(vector<DuckLakeBaseColumnStats> &stats, idx_t c, Vector &data, 
 			break;
 		}
 		default:
-			throw InternalException("FIXME: unsupported nested type");
+			throw InternalException("Unsupported nested type for inline data stats: %s", data.GetType().ToString());
 		}
 		return;
 	}
--- a/src/storage/ducklake_inline_data.cpp
+++ b/src/storage/ducklake_inline_data.cpp
@@ -295,7 +295,7 @@ void UpdateStats(vector<DuckLakeBaseColumnStats> &stats, idx_t c, Vector &data, 
 			break;
 		}
 		default:
-			throw InternalException("Unsupported nested type for inline data stats: %s", data.GetType().ToString());
+			throw NotImplementedException("FIXME: unsupported nested type");
 		}
 		return;
 	}
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -4,7 +4,6 @@
 #include "common/ducklake_util.hpp"
 #include "duckdb/planner/tableref/bound_at_clause.hpp"
 #include "duckdb/common/types/blob.hpp"
-#include "duckdb/common/sql_identifier.hpp"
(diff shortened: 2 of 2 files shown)


arijitroy003:

src/storage/ducklake_inline_data.cpp
```diff
@@ -295,7 +295,7 @@ void UpdateStats(vector<DuckLakeBaseColumnStats> &stats, idx_t c, Vector &data,
 			break;
 		}
 		default:
-			throw InternalException("FIXME: unsupported nested type");
```
dentiny, line 298: I think it's fine to leave the `FIXME` -- it's a well-known comment, a sidenote, I think a more proper exception type is `Unimplemented` instead of internal

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -65,9 +61,6 @@ unique_ptr<DuckLakeMetadataManager> DuckLakeMetadataManager::Create(DuckLakeTran
 		auto create = metadata_manager_iter->second;
 		return create(transaction);
 	}
-	if (catalog_type == "sqlite_scanner") {
```
dentiny, line 68: I'm not sure about the change, could you please elaborate?

src/storage/ducklake_inline_data.cpp
arijitroy003: Good call — restored the FIXME and changed the exception type to `NotImplementedException`. Pushed.

src/storage/ducklake_metadata_manager.cpp
arijitroy003: The removed code was:

```cpp
if (catalog_type == "sqlite_scanner") {
    return make_uniq<SQLiteMetadataManager>(transaction);
}
```

This check was unreachable because `sqlite_scanner` is already registered in the `metadata_managers` map a few lines above:

```cpp
{"sqlite", SQLiteMetadataManager::Create}, {"sqlite_scanner", SQLiteMetadataManager::Create}
```

The `Create` function does a map lookup first (line 59-62), so when `catalog_type == "sqlite_scanner"`, it matches the map entry and returns before ever reaching the explicit `if` check. Both paths produce the same result (`SQLiteMetadataManager`), but the `if` block is dead code since the map lookup always handles it first.

PR #1326, opened 2026-07-20
Search all secret storages when resolving DuckLake secrets by name

### LINKED ISSUE #1318: ATTACH 'ducklake:<secret_name>' cannot resolve secrets stored in extension-registered secret storages (e.g. postgres SECRET_STORAGE_TABLE)
### What happens

`ATTACH 'ducklake:<secret_name>'` fails with `Secret "<secret_name>" was not found` when the ducklake secret lives in an extension-registered secret storage — for example the postgres extension's `SECRET_STORAGE_TABLE` store — even though the secret is visible in `duckdb_secrets()` in the same session.

The cause is visible in [`DuckLakeSecret::GetSecret`](https://github.com/duckdb/ducklake/blob/2856687c875bbee90d523fe15627f8d8fd494622/src/storage/ducklake_secret.cpp#L52-L65): the by-name lookup queries only the two built-in storages, with a FIXME acknowledging the limitation:

```cpp
unique_ptr<SecretEntry> DuckLakeSecret::GetSecret(ClientContext &context, const string &secret_name) {
	auto &secret_manager = SecretManager::Get(context);
	auto transaction = CatalogTransaction::GetSystemCatalogTransaction(context);
	// FIXME: this should be adjusted once the `GetSecretByName` API supports this use case
	auto secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "memory");
	if (secret_entry) {
		return secret_entry;
	}
	secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "local_file");
	if (secret_entry) {
		return secret_entry;
	}
	return nullptr;
}
```

Any secret storage other than `"memory"` and `"local_file"` is never searched, so ducklake secrets managed centrally in a postgres-backed store cannot be used with the `ducklake:<name>` shorthand.

### Reproduction

```sql
-- Postgres-backed secret storage (postgres extension)
ATTACH 'host=... dbname=demo ...' AS demo (TYPE postgres, SECRET_STORAGE_TABLE 'public.duckdb_secrets');

CREATE OR REPLACE SECRET my_lake IN postgres_demo (
    TYPE ducklake,
    METADATA_PATH '...',
    DATA_PATH 's3://bucket/lake/'
);

SELECT name, type, storage FROM duckdb_secrets() WHERE name = 'my_lake';
-- ┌─────────┬──────────┬───────────────┐
-- │ my_lake │ ducklake │ postgres_demo │  ← secret is present and loadable

ATTACH 'ducklake:my_lake' AS lake;
-- Invalid Input Error: Secret "my_lake" was not found - if this was meant to be
-- a path to a DuckDB file, use duckdb:my_lake instead
```

The same secret works fine when created in the `memory` or `local_file` storages, confirming the storage list is the discriminator.

### Expected

The by-name lookup searches all registered secret storages (or `GetSecretByName` grows an all-storages form, per the FIXME), so `ATTACH 'ducklake:<name>'` resolves secrets regardless of which storage holds them. Scope-based lookups (e.g. s3 secrets resolving by path prefix) already work across storages; the by-name path is the outlier.

### Versions

- DuckDB v1.5.4 (Variegata)
- ducklake `d318a545` — reproduced from both `core` and `core_nightly`
- postgres_scanner `ac4f6c6` (core_nightly `9add648` — same behaviour)

### Related

#1270 — the name-charset check in `PathIsSecret()`. With `[A-Za-z0-9_]`-only names that barrier is passed, and this storage-list limitation is the next thing preventing centrally stored ducklake secrets from being attachable by name.

kinghuang:
## Summary

`ATTACH 'ducklake:<secret_name>'` failed with `Secret "<secret_name>" was not found` when the secret lived in an extension-registered secret storage (e.g. the postgres extension's `SECRET_STORAGE_TABLE` store), because `DuckLakeSecret::GetSecret` only queried the two built-in storages, `memory` and `local_file`.

The `GetSecretByName` API the old FIXME was waiting on has since gained support for this use case: when the storage argument is omitted, it searches **all** registered secret storages and raises `Ambiguity detected for secret name '…'` when the same name exists in more than one storage. This PR replaces the hard-coded two-storage lookup with that call.

Fixes #1318

## Behavior change

Previously, a secret name present in both `memory` and `local_file` silently resolved to the `memory` one. It now raises DuckDB's ambiguity error instead, consistent with `DROP SECRET` without a storage qualifier. This felt like the right semantics (silently picking one storage is a footgun), but happy to preserve the old precedence if preferred.

## History

The all-storages form of `GetSecretByName` is not new — it has iterated every registered storage since the multi-storage secret manager was introduced. What the old FIXME was effectively waiting on is duckdb/duckdb#17551 (merged May 2025, first released in DuckDB v1.4.0): before that change, an ambiguous secret name threw `InternalException`, which invalidates the database instance — not something a user should be able to trigger by having two same-named secrets. Since v1.4.0 it throws a regular user-facing `InvalidConfigurationException`, so the lookup is safe to use here. (Caveat: if DuckLake ever builds against pre-1.4 DuckDB, the ambiguity error is fatal there.)

## Tests

- New `test/sql/secrets/ducklake_secret_storages.test`: persistent (`local_file`) secret resolution by name, cross-storage ambiguity error, and re-resolution after dropping the duplicate. The ambiguity case fails without this fix (attach silently succeeds using the memory secret).
- A true extension-storage test would need a live Postgres plus `postgres_scanner`, so it isn't included; the new all-storages code path is exercised via the ambiguity case, which is only reachable through it.
- `./build/release/test/unittest "~[.]test/*"`: all 487 test cases pass (17138 assertions).
- `python3 duckdb/scripts/format.py --all --check --directories src`: passes.

### DIFF
--- a/src/storage/ducklake_secret.cpp
+++ b/src/storage/ducklake_secret.cpp
@@ -52,15 +52,7 @@ CreateSecretFunction DuckLakeSecret::GetFunction() {
 unique_ptr<SecretEntry> DuckLakeSecret::GetSecret(ClientContext &context, const string &secret_name) {
 	auto &secret_manager = SecretManager::Get(context);
 	auto transaction = CatalogTransaction::GetSystemCatalogTransaction(context);
-	// FIXME: this should be adjusted once the `GetSecretByName` API supports this use case
-	auto secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "memory");
-	if (secret_entry) {
-		return secret_entry;
-	}
-	secret_entry = secret_manager.GetSecretByName(transaction, secret_name, "local_file");
-	if (secret_entry) {
-		return secret_entry;
-	}
-	return nullptr;
+	// omitting the storage searches all registered secret storages, including extension-registered ones
+	return secret_manager.GetSecretByName(transaction, secret_name);
 }
 } // namespace duckdb
--- a/test/configs/minio.json
+++ b/test/configs/minio.json
@@ -14,6 +14,12 @@
     "json"
   ],   "skip_tests": [
    {
+      "reason": "on_init initializes the secret manager, so SET secret_directory (needed for the persistent secret storage) cannot be applied; the cross-storage lookup is catalog-agnostic and is covered under the default config",
+      "paths": [
+        "test/sql/secrets/ducklake_secret_storages.test"
+      ]
+    },
+   {
       "reason": "Extension Loading tests, not relevant to remote files.",
       "paths": [
         "test/sql/general/missing_parquet.test",
--- a/test/configs/minio.json
+++ b/test/configs/minio.json
@@ -14,7 +14,7 @@
     "json"
   ],   "skip_tests": [
    {
-      "reason": "on_init initializes the secret manager, so SET secret_directory (needed for the persistent secret storage) cannot be applied; the cross-storage lookup is catalog-agnostic and is covered under the default config",
+      "reason": "Secret storage resolution is independent of the catalog backend and is covered under the default config",
       "paths": [
         "test/sql/secrets/ducklake_secret_storages.test"
       ]
--- a/test/configs/postgres.json
+++ b/test/configs/postgres.json
@@ -19,6 +19,12 @@
   ],
   "skip_tests": [
     {
+      "reason": "on_init initializes the secret manager, so SET secret_directory (needed for the persistent secret storage) cannot be applied; the cross-storage lookup is catalog-agnostic and is covered under the default config",
+      "paths": [
+        "test/sql/secrets/ducklake_secret_storages.test"
+      ]
+    },
+    {
       "reason": "Tests only for DuckDB as a catalog",
       "paths": [
         "test/sql/general/missing_parquet.test",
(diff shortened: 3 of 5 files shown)


pdet:
Thanks, looks good to me.

pdet:
It seems CI is failing, could you have a look at that @kinghuang? Also, while you are at it, could you move this PR to main?

kinghuang:
Done. Rebased onto `main`.

The CI failure was from the new `ducklake_secret_storages.test`. It creates a persistent (`local_file`) secret, which requires `SET secret_directory`, but the `postgres` and `minio` configs initialize the secret manager in their `on_init`, so that `SET` throws. Since the cross-storage lookup is independent of the catalog backend, I've skipped the test under the `postgres`/`sqlite`/`minio` configs (matching how `ducklake_secrets.test` is already handled), and it still runs under the default, `deletion_vectors`, and `no_inline` configs.

pdet:
Hi @kinghuang , seems that CI is still failing

kinghuang:
Made a change to remove `SET secret_directory='...'`. Conflicts under the PostgreSQL config because its `on_init` initializes the secret manager.

@pdet, can you approval a new workflow run for testing?

kinghuang:
@pdet, I've rebased for #1352.

pdet:
Thanks!

PR #1327, opened 2026-07-20
Scope Postgres metadata attach to METADATA_SCHEMA

### LINKED ISSUE #1064: BUG: Postgres backend: ATTACH reflects the entire database regardless of METADATA_SCHEMA
### What happens?

When the metadata backend is PostgreSQL, every DuckLake `ATTACH` causes the `duckdb-postgres` extension to issue a bulk catalog reflection over `pg_class` / `pg_attribute` / `pg_namespace` / `pg_type` — see `src/storage/postgres_table_set.cpp` ([GetInitializeQuery](https://github.com/duckdb/duckdb-postgres/blob/main/src/storage/postgres_table_set.cpp#L22-L52)).

DuckLake already knows that it only cares about the schema named by `METADATA_SCHEMA` — every metadata read it issues is scoped to `{METADATA_CATALOG}.{METADATA_SCHEMA}.*`. But it doesn't forward this information to the underlying Postgres attach, so the reflection runs unscoped against the *whole* database. In a multi-tenant deployment where many tenants share one Postgres catalog (each tenant using a different `METADATA_SCHEMA`), every attach scans the catalog for every tenant's relations.

This problem is reinforced by [#1065](https://github.com/duckdb/ducklake/issues/1065): the leaked `ducklake_inlined_data_*` tables from that bug add tens of thousands of unrelated relations to `pg_class`, which is exactly the workload this issue's unscoped reflection then has to scan on every attach. In our prod cluster, ~46k orphan tables had accumulated across 83 tenants — every attach paid for them via this issue. Fixing one mitigates the other; fixing both is what eliminates the cost.

A user can work around this today by manually passing `META_SCHEMA '<same-value-as-METADATA_SCHEMA>'` so the `META_*` pass-through in `DuckLakeInitializer::GetAttachOptions()` forwards it to duckdb-postgres's `SCHEMA 'xxx'` option. But:
- This is surprising — you have to pass the same value twice.
- It's not documented on https://ducklake.select.
- The performance delta is very large, so the default matters.

**Proposed fix:** When the metadata backend is Postgres and `METADATA_SCHEMA` is set (either explicitly or by default), DuckLake should automatically include `SCHEMA '<metadata_schema>'` in the attach options it generates for duckdb-postgres, unless the user has already set it explicitly via `META_SCHEMA` or `metadata_parameters`.

Relevant upstream:
- duckdb-postgres [#221](https://github.com/duckdb/duckdb-postgres/issues/221) — same slow attach reported.
- duckdb-postgres [#259](https://github.com/duckdb/duckdb-postgres/pull/259) — introduced the `SCHEMA 'xxx'` attach option this issue asks DuckLake to opt into automatically.

### To Reproduce

Self-contained minimal repro. Demonstrates that an `ATTACH` against one DuckLake metadata schema reflects the entire database, including 100k tables in 100 unrelated schemas.

Requires `docker` and `duckdb >= 1.5.2`:

```bash
#!/usr/bin/env bash
set -euo pipefail

docker run -d --rm --name ducklake-1064 -p 55432:5432 \
    -e POSTGRES_PASSWORD=x -e POSTGRES_DB=lake postgres:17 >/dev/null
trap 'docker rm -f ducklake-1064 >/dev/null' EXIT
until docker exec ducklake-1064 pg_isready -U postgres -d lake >/dev/null 2>&1; do sleep 0.5; done

PSQL() { PGPASSWORD=x psql -h 127.0.0.1 -p 55432 -U postgres -d lake -v ON_ERROR_STOP=1 -q "$@"; }
{
    echo "CREATE SCHEMA target;"
    for i in {1..100}; do
        echo "CREATE SCHEMA other_${i};"
        for j in {1..1000}; do echo "CREATE TABLE other_${i}.junk_${j}(id bigint, a text, b text, c jsonb, d timestamp);"; done
    done
} | PSQL >/dev/null

mkdir -p /tmp/lake1064
DSN="postgres:host=127.0.0.1 port=55432 user=postgres password=x dbname=lake"
ATTACH() { duckdb -c "INSTALL ducklake;LOAD ducklake;INSTALL postgres;LOAD postgres;ATTACH 'ducklake:$DSN' AS l (DATA_PATH '/tmp/lake1064/', METADATA_SCHEMA 'target'$1);" >/dev/null; }

ATTACH ""        # warmup, discarded
time ATTACH ", META_SCHEMA 'target'"
time ATTACH ""
```

Observed on macOS arm64 / DuckDB 1.5.2 / Postgres 17, 100,001 user relations (1 target + 100×1000 unrelated):

```
with META_SCHEMA       real 0m0.218s
without META_SCHEMA    real 0m2.009s
```

~9× gap on local docker.

Also reproduced end-to-end against a real production DuckLake (DuckLake 1.0 metadata backend on Postgres 17.7 with ~214k relations across ~70 schemas), using a single DuckDB 1.5.2 `ATTACH` via the DuckDB CLI:

```
Scenario                           | Wall-clock of the ATTACH statement
-----------------------------------+-----------------------------------
Without META_SCHEMA                |  8.9 s
With  META_SCHEMA '<md-schema>'    |  3.5 s
Delta                              |  ~5.4 s saved per attach
```

And the same prod database, measured directly with `EXPLAIN (ANALYZE, BUFFERS)` against the reflection query:

```
Scenario                           | Exec time | Rows returned | Temp disk spill | Buffer hits
-----------------------------------+-----------+---------------+-----------------+------------
Unscoped (current default)         |  5,101 ms |     1,329,015 |    317 MB       |     55,467
With pg_namespace.nspname = <md-s> |     59 ms |         7,309 |    0 (in-mem)   |      7,559
Delta                              |  ~86×     |      ~182×    |   eliminated    |     ~7.3×
```

The 5.4 s wall-clock delta maps 1:1 to the 5.1 s reflection-query delta — confirming that this one query dominates the extra cost.

Unscoped plan: seq-scan `pg_attribute` (1.68 M rows), seq-scan `pg_class` (74 k after filter), seq-scan `pg_type` (148 k), external merge sort to disk.

Scoped plan: `pg_namespace_nspname_index` → `pg_class_relname_nsp_index` → `pg_attribute_relid_attnum_index`, in-memory sort, under 60 ms.

The relevant DuckLake code is `src/storage/ducklake_initializer.cpp` [`GetAttachOptions()`](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_initializer.cpp#L25-L59) and the unconditional `ATTACH OR REPLACE` in [`Initialize()`](https://github.com/duckdb/ducklake/blob/main/src/storage/ducklake_initializer.cpp#L61-L70). A plausible fix is roughly:

```cpp
// In GetAttachOptions(), after processing explicit metadata_parameters:
const bool is_postgres_backend = catalog.MetadataType() == "postgres";
const bool user_set_schema =
    options.metadata_parameters.count("schema") > 0;
if (is_postgres_backend && !user_set_schema && !options.metadata_schema.empty()) {
    attach_options.push_back("SCHEMA " + DuckLakeUtil::SQLLiteralToString(options.metadata_schema));
}
```

(Same treatment should apply if/when analogous options exist for other backends.)

### OS:

macOS 15 (arm64) for the local repro; production catalog is AWS Aurora PostgreSQL 17.7 on Linux.

### DuckDB Version:

v1.5.2 (Variegata) 8a5851971f (also reproduced on v1.5.0 3a3967aa81)

### DuckLake Version:

DuckLake extension commit `415a9ebd` (PR #966, "v1.0", 2026-04-09), metadata storage version 1.0. Also reproduced on the earlier `7ea15644` (PR #803, 2026-02-26) with storage version 0.4.

### DuckDB Client:

CLI (`duckdb`) for the repro; production usage is via the Python client.

### Hardware:

N/A — this is a Postgres catalog–side cost. The effect scales with total relation count in the attached Postgres database, not with client hardware.

### Full Name:

Marcel Steinbach

### Affiliation:

Ometria

- [x] Yes, I have

- [x] Yes, I have

kinghuang:
Fixes #1064

## Problem

When the metadata backend is Postgres, every DuckLake `ATTACH` triggers duckdb-postgres's bulk catalog reflection over the *entire* database, even though DuckLake only ever reads `{METADATA_CATALOG}.{METADATA_SCHEMA}.*`. On large or multi-tenant catalogs this is expensive: the repro in #1064 shows a ~9× attach slowdown with 100k unrelated relations locally, and ~5.4 s of extra wall-clock per attach on a production catalog (86× on the reflection query itself, with a 317 MB temp-disk spill).

duckdb-postgres has supported scoping the attach via the `SCHEMA` option since duckdb/duckdb-postgres#259, and users can already opt in manually by passing `META_SCHEMA '<same value as METADATA_SCHEMA>'` — but that requires passing the same value twice and is easy to miss.

## Fix

When the metadata backend is Postgres and the user explicitly set `METADATA_SCHEMA`, `DuckLakeInitializer::GetAttachOptions()` now automatically forwards it as the `SCHEMA` option of the underlying Postgres attach.

Notes on scope:

- **Only when `METADATA_SCHEMA` is explicit.** `GetAttachOptions()` runs before the metadata catalog is attached, and the default schema is only resolved afterwards via `GetDefaultSchemaName()`, so the default-schema case is left unscoped (unchanged behavior). Scoping the default case (duckdb-postgres resolves it to `public`) could be a follow-up if desirable.
- **An explicit `META_SCHEMA` always wins.** If `schema` is present in the metadata parameters (via `META_SCHEMA` or `metadata_parameters`), no option is added. In particular `META_SCHEMA ''` restores the previous unscoped behavior as an escape hatch.
- **New-lake creation works.** `MetadataExists()` already treats missing-schema catalog errors as "fresh DuckLake", and `CREATE SCHEMA IF NOT EXISTS` is not blocked by a scoped Postgres attach; the new test covers creating a lake whose metadata schema does not exist in Postgres yet.

## Testing

Added `test/sql/metadata/metadata_schema_scoped_attach.test` (gated on `postgres_scanner` + `DUCKLAKE_CI`, like the other Postgres-specific tests), covering:

- creating a new DuckLake in a `METADATA_SCHEMA` that does not yet exist in Postgres
- unrelated Postgres schemas are no longer reflected into the metadata catalog (this assertion fails without the fix)
- re-attaching an existing lake stays scoped
- explicit `META_SCHEMA` passthrough keeps working
- `META_SCHEMA ''` opts out of the automatic scoping

Ran the full suite with `test/configs/postgres.json` against Postgres 17: no regressions (the only local failures are missing-`httpfs` environment artifacts and the pre-existing `inlined_data_table_leak.test` failure, which reproduces identically without this change).

- [ ] CI green (DBMS Catalog Tests / Postgres job exercises the new test)

### DIFF
--- a/src/storage/ducklake_initializer.cpp
+++ b/src/storage/ducklake_initializer.cpp
@@ -12,6 +12,7 @@
 #include "storage/ducklake_catalog.hpp"
 #include "storage/ducklake_transaction.hpp"
 #include "storage/ducklake_schema_entry.hpp"
+#include "common/ducklake_util.hpp"
 #include "common/ducklake_version.hpp"
 #include "metadata_manager/ducklake_metadata_manager_v1_1.hpp"
 #include "metadata_manager/sqlite_metadata_manager.hpp"
@@ -46,6 +47,16 @@ string DuckLakeInitializer::GetAttachOptions() {
 	if (metadata_type.empty() || metadata_type == "duckdb") {
 		// this is duckdb, we always do latest storage
 		attach_options.push_back(StringUtil::Format("STORAGE_VERSION '%s'", "latest"));
+	}
+	// scope the underlying Postgres attach to the metadata schema - otherwise the postgres extension reflects
+	// every schema in the database on attach, which is very slow on large or multi-tenant catalogs.
+	// only done when the user explicitly set METADATA_SCHEMA - the default schema is not known until after attach.
+	// an explicit META_SCHEMA (or metadata_parameters 'schema') takes precedence over this.
+	bool is_postgres = metadata_type == "postgres" || metadata_type == "postgres_scanner";
+	bool user_set_schema = options.metadata_parameters.find("schema") != options.metadata_parameters.end();
+	if (is_postgres && !user_set_schema && !options.metadata_schema.empty()) {
+		attach_options.push_back("SCHEMA " +
+		                         DuckLakeUtil::SQLLiteralToString(options.metadata_schema.GetIdentifierName()));
 	}
 	if (options.hide_metadata_catalog) {
 		attach_options.push_back("HIDDEN true");
--- a/test/sql/metadata/metadata_schema_scoped_attach.test
+++ b/test/sql/metadata/metadata_schema_scoped_attach.test
@@ -8,8 +8,6 @@ require parquet
 
 require postgres_scanner
 
-require-env DUCKLAKE_CI
-
 test-env DATA_PATH {TEST_DIR}
 
 # set up an unrelated schema in the metadata database - it should not be reflected by DuckLake attaches
--- a/test/sql/metadata/metadata_schema_scoped_attach.test
+++ b/test/sql/metadata/metadata_schema_scoped_attach.test
@@ -1,0 +1,86 @@
+# name: test/sql/metadata/metadata_schema_scoped_attach.test
+# description: Postgres metadata attach is scoped to METADATA_SCHEMA (issue #1064)
+# group: [metadata]
+
+require ducklake
+
+require parquet
+
+require postgres_scanner
+
+require-env DUCKLAKE_CI
+
+test-env DATA_PATH {TEST_DIR}
+
+# set up an unrelated schema in the metadata database - it should not be reflected by DuckLake attaches
+statement ok
+ATTACH 'postgres:dbname=ducklakedb' AS pg_setup
+
+statement ok
+CALL postgres_execute('pg_setup', 'DROP SCHEMA IF EXISTS junk_1064 CASCADE; DROP SCHEMA IF EXISTS scoped_meta_1064 CASCADE; CREATE SCHEMA junk_1064; CREATE TABLE junk_1064.junk_tbl(i INTEGER);')
+
+statement ok
+DETACH pg_setup
+
+# create a new DuckLake in a METADATA_SCHEMA that does not exist yet in Postgres
+statement ok
+ATTACH 'ducklake:postgres:dbname=ducklakedb' AS lake (DATA_PATH '{DATA_PATH}/scoped_files', METADATA_SCHEMA 'scoped_meta_1064', METADATA_CATALOG 'pgmeta')
+
+statement ok
+CREATE TABLE lake.tbl(i INTEGER);
+
+statement ok
+INSERT INTO lake.tbl VALUES (42);
+
+query I
+SELECT * FROM lake.tbl
+----
+42
+
+# the underlying Postgres attach is scoped to the metadata schema - unrelated schemas are not reflected
+query I
+SELECT COUNT(*) FROM duckdb_schemas() WHERE database_name='pgmeta' AND schema_name='junk_1064'
+----
+0
+
+query I
+SELECT COUNT(*) FROM duckdb_schemas() WHERE database_name='pgmeta' AND schema_name='scoped_meta_1064'
+----
+1
+
+statement ok
+DETACH lake
+
+# re-attaching an existing DuckLake is scoped as well
+statement ok
+ATTACH 'ducklake:postgres:dbname=ducklakedb' AS lake2 (DATA_PATH '{DATA_PATH}/scoped_files', METADATA_SCHEMA 'scoped_meta_1064', METADATA_CATALOG 'pgmeta2')
+
+query I
+SELECT * FROM lake2.tbl
+----
+42
+
+query I
+SELECT COUNT(*) FROM duckdb_schemas() WHERE database_name='pgmeta2' AND schema_name='junk_1064'
+----
+0
+
+statement ok
+DETACH lake2
+
+# an explicit META_SCHEMA passthrough keeps working
+statement ok
+ATTACH 'ducklake:postgres:dbname=ducklakedb' AS lake3 (DATA_PATH '{DATA_PATH}/scoped_files', METADATA_SCHEMA 'scoped_meta_1064', METADATA_CATALOG 'pgmeta3', META_SCHEMA 'scoped_meta_1064')
+
+query I
+SELECT * FROM lake3.tbl
+----
+42
+
+query I
+SELECT COUNT(*) FROM duckdb_schemas() WHERE database_name='pgmeta3' AND schema_name='junk_1064'
+----
+0
+
+statement ok
+DETACH lake3
--- a/test/configs/quack.json
+++ b/test/configs/quack.json
@@ -34,7 +34,8 @@
     {
       "reason": "Test only for Postgres as catalog",
       "paths": [
-        "test/sql/metadata/ducklake_settings_postgres.test"
+        "test/sql/metadata/ducklake_settings_postgres.test",
+        "test/sql/metadata/metadata_schema_scoped_attach.test"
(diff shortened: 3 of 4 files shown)


kinghuang:
The two failing Windows build jobs are a pre-existing base-branch issue, not caused by this PR.

Both die in `duckdb/third_party/fmt/include/fmt/format.h` (`'stdext': is not a class or namespace name` / `checked_array_iterator` — an MSVC STL extension removed in newer toolchains). This reproduces on the current `v1.5-variegata` tip ([run 29743345177](https://github.com/duckdb/ducklake/actions/runs/29743345177), commit `d8a1881e`, the parent this PR is rebased onto) with no DuckLake changes involved.

All test suites pass, including the Postgres catalog tests that cover this change.

pdet:
Hi @kinghuang, thanks for the PR, can you retarget to `main`?

kinghuang:
Retargeted to `main`. @pdet, can you approve the workflow runs for tests?

kinghuang:
It looks like the new failing runs are from changes in `main`. See https://github.com/duckdb/ducklake/actions/runs/30466138293. The `duckdb-version` pin might need a bump.

> ParquetFileMetadata has no member named 'filename'

I'll rebase and check again after `main` passes.

kinghuang:
@pdet, I've rebased for #1352.

kinghuang:
@pdet, listed the test to be PostgreSQL only. Can you approve the workflow run again?

test/sql/metadata/metadata_schema_scoped_attach.test
```diff
@@ -0,0 +1,86 @@
...
+# description: Postgres metadata attach is scoped to METADATA_SCHEMA (issue #1064)
+# group: [metadata]
+
+require ducklake
+
+require parquet
+
+require postgres_scanner
+
+require-env DUCKLAKE_CI
```
pdet: Why do we need this variable? Is this set in any of our CI?
Maybe we should remove this and skip this test where appropriate as it is a postgres scanner only

pdet (commented):
Hi, @kinghuang, on a second look I have one comment.

test/sql/metadata/metadata_schema_scoped_attach.test
kinghuang: Ah, it's just copied from `ducklake_settings_postgres.test`. I agree, it doesn't serve any purpose here. I'll drop that.

Claude says it's a vestigial stopgap from when PostgreSQL CI was disabled? I'm not really sure.

src/storage/ducklake_initializer.cpp
```diff
@@ -47,6 +48,16 @@ string DuckLakeInitializer::GetAttachOptions() {
 		// this is duckdb, we always do latest storage
 		attach_options.push_back(StringUtil::Format("STORAGE_VERSION '%s'", "latest"));
 	}
+	// scope the underlying Postgres attach to the metadata schema - otherwise the postgres extension reflects
+	// every schema in the database on attach, which is very slow on large or multi-tenant catalogs.
+	// only done when the user explicitly set METADATA_SCHEMA - the default schema is not known until after attach.
+	// an explicit META_SCHEMA (or metadata_parameters 'schema') takes precedence over this.
```
pdet: Can we keep comments max to one line?

PR #1328, opened 2026-07-20
Fix rewrites after partition evolution

jdctinuiti:
Fixes repeated delete rewrites after a table's partition spec changes by resolving output paths from the data file's historical partition spec.

Tests: full DuckLake suite, deletion-vector config, and Quack-backed regression.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -575,7 +575,28 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
 	auto &columns = table.GetColumns();
 	string data_path;
 	if (partition_id.IsValid()) {
-		data_path = DuckLakePartitionUtils::BuildHivePartitionPath(table, partition_values, catalog.Separator());
+		auto partition_table = &table;
+		auto partition_data = partition_table->GetPartitionData();
+		if (!partition_data || partition_data->partition_id != partition_id.GetIndex()) {
+			auto partition_snapshot_id = source_files[0].partition_snapshot_id;
+			auto partition_schema_version = source_files[0].partition_schema_version;
+			if (!partition_snapshot_id.IsValid() || !partition_schema_version.IsValid()) {
+				throw InternalException("DuckLakeCompactor: failed to find partition schema");
+			}
+			DuckLakeSnapshot partition_snapshot(partition_snapshot_id.GetIndex(), partition_schema_version.GetIndex(),
+			                                    0, 0);
+			auto partition_entry = catalog.GetSchemaForSnapshot(transaction, partition_snapshot).GetEntryById(table_id);
+			if (!partition_entry) {
+				throw InternalException("DuckLakeCompactor: failed to find table entry for partition schema");
+			}
+			partition_table = &partition_entry->Cast<DuckLakeTableEntry>();
+		}
+		partition_data = partition_table->GetPartitionData();
+		if (!partition_data || partition_data->partition_id != partition_id.GetIndex()) {
+			throw InternalException("DuckLakeCompactor: failed to find partition spec");
+		}
+		data_path =
+		    DuckLakePartitionUtils::BuildHivePartitionPath(*partition_table, partition_values, catalog.Separator());
 	}
 
 	bool write_row_id = false;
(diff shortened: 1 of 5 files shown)


test/sql/rewrite_data_files/test_rewrite_after_partition_evolution.test
```diff
@@ -0,0 +1,130 @@
+# name: test/sql/rewrite_data_files/test_rewrite_after_partition_evolution.test
```
pdet, line 1: Can we add something like the following as a test?
```sql
      
 ATTACH 'ducklake:/tmp/t3diag/meta.db' AS ducklake (
        DATA_PATH '/tmp/t3diag/data',
        METADATA_CATALOG 'metadata',
        DATA_INLINING_ROW_LIMIT 0
);
    
CREATE TABLE ducklake.t(source VARCHAR, id VARCHAR, v INTEGER);
    
BEGIN;
ALTER TABLE ducklake.t SET PARTITIONED BY (source);
INSERT INTO ducklake.t VALUES ('old', 'new', 1), ('old', 'new', 2), ('old', 'new', 3);
ALTER TABLE ducklake.t SET PARTITIONED BY (id);
COMMIT;

DELETE FROM ducklake.t WHERE v = 1;
CALL ducklake_rewrite_data_files('ducklake', 't', delete_threshold => 0);

```

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -575,7 +575,28 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
...
+		auto partition_data = partition_table->GetPartitionData();
+		if (!partition_data || partition_data->partition_id != partition_id.GetIndex()) {
+			auto partition_snapshot_id = source_files[0].partition_snapshot_id;
+			auto partition_schema_version = source_files[0].partition_schema_version;
+			if (!partition_snapshot_id.IsValid() || !partition_schema_version.IsValid()) {
+				throw InternalException("DuckLakeCompactor: failed to find partition schema");
+			}
+			DuckLakeSnapshot partition_snapshot(partition_snapshot_id.GetIndex(), partition_schema_version.GetIndex(),
+			                                    0, 0);
+			auto partition_entry = catalog.GetSchemaForSnapshot(transaction, partition_snapshot).GetEntryById(table_id);
```
pdet: Should this just be ` catalog.GetEntryById` ?

src/functions/ducklake_compaction_functions.cpp
```diff
@@ -575,7 +575,28 @@ DuckLakeCompactor::GenerateCompactionCommand(vector<DuckLakeCompactionFileEntry>
...
+			if (!partition_entry) {
+				throw InternalException("DuckLakeCompactor: failed to find table entry for partition schema");
+			}
+			partition_table = &partition_entry->Cast<DuckLakeTableEntry>();
+		}
+		partition_data = partition_table->GetPartitionData();
+		if (!partition_data || partition_data->partition_id != partition_id.GetIndex()) {
+			throw InternalException("DuckLakeCompactor: failed to find partition spec");
+		}
```
pdet: Can this be a helper function? maybe `ResolvePartitionSpecTable`

pdet (commented):
Thanks for the PR, I've added a few commetns

test/sql/rewrite_data_files/test_rewrite_after_partition_evolution.test
jdctinuiti: Added this same-transaction sequence to the existing same-cardinality regression. It exposed an intermediate partition-spec metadata gap, so the fix now preserves that spec and asserts every file partition ID remains resolvable.

src/functions/ducklake_compaction_functions.cpp
jdctinuiti: Yes, this now uses catalog.GetEntryById(transaction, partition_snapshot, table_id).

src/functions/ducklake_compaction_functions.cpp
jdctinuiti: Done!

jdctinuiti:
Hey @pdet thank you for the review! Do you think you could also review https://github.com/duckdb/ducklake/pull/1261 and https://github.com/duckdb/ducklake/pull/1260? They are part of the changes you suggested i reintroduce in smaller PRs.

pdet:
Thanks!

Issue #1225, opened 2026-06-06
Reading a VARCHAR column from a Postgres-backed inlined table crashes once the inlined chunk exceeds ~2 GB (linux_arm64)


giomattoni:
### What happens?

On a DuckLake catalog backed by Postgres, reading a VARCHAR column from a
table whose `ducklake_inlined_data_<tid>_<sv>` chunk holds more than ~2 GB
of cell bytes crashes with:

```
INTERNAL Error: Attempted to access index 0 within vector of size 0
  duckdb::PostgresMetadataManager::TransformInlinedData+0x794
  duckdb::DuckLakeInlinedDataReader::TryInitializeScan+0xa44
  duckdb::MultiFileFunction<ParquetMultiFileInfo>::TryInitializeNextBatch+0x738
```
Other reads (count(*), non-VARCHAR columns, the same VARCHAR column on smaller
tables) all succeed. The session is left FATAL'd and must be reconnected.

Reproduces on the linux_arm64 build of the extension. The same script run
against the osx_arm64 build does NOT reproduce.

Related (same error string, different code path): #281.
Related (other Postgres-inlined-table bugs): #1123, #1089, #1155, #1219.

### To Reproduce

Start an empty Postgres on `localhost:55432`:

```bash
docker run --rm -d -p 55432:5432 -e POSTGRES_DB=repro -e POSTGRES_USER=repro -e POSTGRES_PASSWORD=repro postgres:17-alpine
```

`Dockerfile`:

```dockerfile
FROM --platform=linux/arm64 python:3.12-slim
RUN pip install --no-cache-dir 'duckdb==1.5.2' psycopg2-binary
COPY repro.py /repro.py
ENTRYPOINT ["python", "-u", "/repro.py"]
```

`repro.py`:

```python
import duckdb, psycopg2, os, shutil

N_ROWS = int(os.environ.get("N_ROWS", "70"))
DSN  = "host=host.docker.internal port=55432 dbname=repro user=repro password=repro"
DATA = "/tmp/dldata/"

pg = psycopg2.connect(host="host.docker.internal", port=55432,
                      dbname="repro", user="repro", password="repro")
pg.autocommit = True
pg.cursor().execute("DROP SCHEMA IF EXISTS public CASCADE; CREATE SCHEMA public")
pg.close()
shutil.rmtree(DATA, ignore_errors=True); os.makedirs(DATA, exist_ok=True)

c = duckdb.connect()
[...]

PR #1330, opened 2026-07-21
Surface errored inlined-data reads as catchable errors

GavinPizza:
`PostgresMetadataManager::TransformInlinedData` runs its `BLOB->VARCHAR` reinterpret type-compare loop over `result.types[i]` before it checks `result.HasError()`. When the inlined-data read comes back as an errored or empty `QueryResult`, `result.types` is empty. In a release build the guarding `D_ASSERT` is compiled out, so the loop indexes `result.types[0]` on an empty vector and throws a fatal `InternalException` ("Attempted to access index 0 within vector of size 0"). That invalidates the shared DuckDB instance instead of surfacing the real read failure.

This reorders the checks so an internal read failure is reported as a catchable error:

- hoist the existing `result.HasError()` check to the top, so the real read error is thrown before any type comparison;
- add an explicit column-count guard that throws a catchable `InvalidInputException` when the result has fewer columns than expected.

The happy path is unchanged. This is related to #1225 (Postgres inlined VARCHAR read crashes once the chunk exceeds ~2 GB) and #1089 (long column names crash on flush with the same error string). Neither reproduces from plain SQL, so the PR does not claim to close them. It turns their fatal, instance-killing `InternalException` into a catchable error that names the actual failure.

### Testing

`test/sql/issues/inlined_reader_empty_read.test` is Postgres-backed, gated on `DUCKLAKE_CI` plus `postgres_scanner`, and run by the `Catalogs.yml` job. It round-trips the healthy reinterpret path, then fault-injects the errored read by dropping the inlined-data table out from under the reader. Without the fix it goes red (`INTERNAL Error: Attempted to access index 0 within vector of size 0`, instance invalidated); with the fix it is green (catchable error, and the following `SELECT 1` still succeeds). Reproduce it against a release build, since `D_ASSERT` masks the bug in debug.


GavinPizza:
Fixed the Format Check failure. No logic changes as it was just a clang-format nit on the new error message.

pdet:
Thanks, LGTM!

Issue #1307, opened 2026-07-13
FLOAT->DOUBLE column retype corrupts min/max stats, causing wrong zone-map pruning


GavinPizza:
### What happens?

If I widen a `FLOAT` column to `DOUBLE` with `ALTER TABLE ... ALTER COLUMN c TYPE DOUBLE`, queries that filter on that column start returning wrong results for the rows that were written before the retype. A filter that should match those rows prunes them instead, with no error.

The reason is the min/max stats. DuckLake stores them as strings, and after the retype the stored string is read back as a `DOUBLE`. For floats that isn't the same value: a `FLOAT` max of `0.1` is stored as `"0.1"`, but `"0.1"` read as a `DOUBLE` gives the double nearest `0.1`, while the `FLOAT` `0.1` widened to `DOUBLE` is the larger `0.10000000149011612`. So the recorded max (`0.1`) is now smaller than a value that's really in the column, and the zone map skips the rows above it.

A plain `FLOAT` column (no retype) is fine, and integer/decimal widenings are fine too, because their string form reads back the same at the wider type. It's specifically the `FLOAT`/`REAL` -> `DOUBLE` widening that breaks.

### To Reproduce

Stock DuckDB plus the `ducklake` extension. Every row is `0.1::FLOAT`, which reads back as `0.10000000149011612` once the column is `DOUBLE`, so `v > 0.1` should match all of them.

```python
import duckdb
import tempfile

d = tempfile.mkdtemp()
con = duckdb.connect()
con.execute("INSTALL ducklake; LOAD ducklake;")
con.execute(f"ATTACH 'ducklake:{d}/f.ducklake' AS dl (DATA_PATH '{d}/data')")

con.execute("CREATE TABLE dl.t (v FLOAT)")
con.execute("INSERT INTO dl.t SELECT 0.1::FLOAT FROM range(2048)")
con.execute("ALTER TABLE dl.t ALTER COLUMN v TYPE DOUBLE")

# every row reads back as 0.10000000149011612 > 0.1, so this should be 2048
print(con.execute("SELECT count(*) FROM dl.t WHERE v > 0.1").fetchone())

# a forced scan proves such values are really there
print(con.execute("SELECT max(v + 0.0) FROM dl.t").fetchone())
```

Output:

```
(0,)
(0.10000000149011612,)
```

[...]

PR #1331, opened 2026-07-21
Cast column min/max across ALTER COLUMN TYPE instead of discarding

### LINKED ISSUE #1308: Same-transaction ALTER COLUMN TYPE widening + INSERT drops column min/max stats
### What happens?

If I widen a column's type with `ALTER TABLE ... ALTER COLUMN c TYPE <wider>` (say `INTEGER` to `BIGINT`) and then insert into the table, the table-level min/max stats for that column come back NULL, even though the column still has plenty of non-NULL data. A plain INSERT of the same type keeps the stats fine, so it's the retype that breaks them. Doing the retype and the insert in one transaction is what triggers it: the merge that builds the global stats gets the old min/max at the pre-ALTER type and the new file's min/max at the post-ALTER type, decides they're different types and can't be compared, and just throws the min/max away.

Those stats are the column's zone map, so once they read NULL nothing can prune on that column. That's the symptom I can reproduce every time, and it's what the script below shows. I think there's a worse version tied to the cache-key gap in the root-cause section, where the min gets left at a stale non-NULL value instead of NULL and could throw off the decode of older row groups into wrong values with no error, but I haven't been able to reproduce that reliably so I'm not leaning on it here.

You can see it straight in the metadata: after the steps below, `min_value` and `max_value` for the column are NULL even though the data clearly runs from 1000 to 5063.

### To Reproduce

One connection, stock DuckDB with the `ducklake` extension. The retype and the insert go in the same transaction, which is what makes it fire: the seed stats get read at the pre-ALTER type while the new file is written at the post-ALTER type.

```python
import duckdb
import tempfile

d = tempfile.mkdtemp()
con = duckdb.connect()
con.execute(
    f"ATTACH 'ducklake:{d}/my.ducklake' AS dl "
    f"(DATA_PATH '{d}/data', METADATA_CATALOG 'm')"
)
con.execute("CREATE TABLE dl.t (id INTEGER, v INTEGER)")

# batch 1: pre-ALTER rows, v in [1000, 1063]
con.execute("INSERT INTO dl.t SELECT i, 1000 + (i % 64) FROM range(4096) t(i)")

# stats are correct here -> [('1000', '1063')]
print(con.execute(
    "SELECT min_value, max_value FROM m.ducklake_table_column_stats "
    "WHERE table_id = 1 AND column_id = 2"
).fetchall())

# retype + insert in the SAME transaction
con.execute("BEGIN")
con.execute("ALTER TABLE dl.t ALTER COLUMN v TYPE BIGINT")   # widening, catalog only
con.execute("INSERT INTO dl.t SELECT 100000 + i, 5000 + (i % 64) FROM range(4096) t(i)")
con.execute("COMMIT")

# bug: min/max are now NULL, though the data spans 1000..5063
print(con.execute(
    "SELECT min_value, max_value FROM m.ducklake_table_column_stats "
    "WHERE table_id = 1 AND column_id = 2"
).fetchall())
```

Output:

```
[('1000', '1063')]
[(None, None)]
```

**Observed:** after the commit, `min_value` and `max_value` are both NULL.

**Expected:** `[('1000', '5063')]`. The old min/max should just be widened to BIGINT and merged with the new file, the same as a normal same-type INSERT.

If you don't want to rely on the exact NULL, the simpler check is that the recorded `min_value`/`max_value` should always bound the table's real `min(v)`/`max(v)`. After the retype they're NULL and bound nothing.

`DuckLakeColumnStats::MergeStats` in `src/storage/ducklake_stats.cpp` throws the min/max out whenever the type changes instead of casting them:

```cpp
if (types_differ) {
    // min/max from different types cannot be compared - invalidate them
    has_min = false;   // line 148
    has_max = false;   // line 149
}
```

So the min/max it was carrying across the retype get dropped and nothing rebuilds them. DuckLake only allows widening promotions, so for the integer/decimal/temporal widenings the old values are representable in the new type (`INTEGER 1000` is a fine `BIGINT 1000`) and there's no reason to drop them. Casting them across instead keeps the table-level min/max valid. (The `FLOAT`/`REAL` → `DOUBLE` widening has a separate string-reload subtlety tracked in #1307.)

There's a related spot on the separate-commit path worth flagging: `StatsCacheKey` (`src/storage/ducklake_catalog.cpp:1083`) keys the stats cache on `(next_file_id, table_id)` and leaves out `schema_version`. A catalog-only `ALTER COLUMN ... TYPE` bumps `schema_version` but not `next_file_id`, so a later read can be handed a stale entry that's still at the old type (call site at `ducklake_catalog.cpp:804`). The same-transaction repro above hits the bug without needing the cache at all, but it's the same underlying type mismatch so I'm noting it here.

*(Line numbers are on `34db89b2`, current `main`.)*

### OS:

### DuckDB Version:

### DuckLake Version:

### DuckDB Client:

Python

### Hardware:

Not performance-related.

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [X] Yes, I have

- [X] Yes, I have

GavinPizza:
Closes #1308

A catalog-only `ALTER TABLE … ALTER COLUMN c TYPE <wider>` (a widening promotion, e.g. `INTEGER` to `BIGINT`) followed by an `INSERT` corrupts the table-level min/max stats for that column. When the retype and the insert commit together, the global-stats merge sees the accumulator's min/max at the pre-ALTER type and the new file's min/max at the post-ALTER type, decides the types are incomparable, and **discards** the table-level min/max, leaving `ducklake_table_column_stats.min_value`/`max_value` `NULL` even though the column has non-NULL data over a known range.

Those stats are a zone map. Dropping them silently loses predicate pruning, and on the paths where the merge produces a *value-wrong* (rather than merely `NULL`) min, DuckDB's compressed materialization can use it as an exact re-base constant and mis-decode older row groups, a silent wrong read of pre-ALTER rows with no error raised.

The root cause is in `DuckLakeColumnStats::MergeStats`: on any type change it set `has_min = has_max = false`. But DuckLake only permits **widening** promotions, so the existing min/max are always representable in the new type, so discarding them is never required. This PR instead **casts** the existing min/max to the new type (`DefaultTryCastAs`) and keeps them when the cast is lossless, invalidating only the rare value that genuinely cannot be represented. The min/max then stay valid and comparable across the retype, exactly as a same-type INSERT would leave them.

Adds a regression test (`test/sql/alter/alter_column_type_stats.test`) that pins the table-level stats after a widening retype+insert and checks they still bound the data. It is **red on `main`** (stats come back `NULL`) and green with this change.

Scope: this is value-preserving for the integer/decimal/temporal widenings whose string-encoded stats read back unchanged at the wider type. The `FLOAT`/`REAL` to `DOUBLE` case is intentionally out of scope: its persisted string stats are reinterpreted at the current type on reload (a separate, pre-existing bug that reproduces on `main` independent of this change), tracked in #1307.


src/storage/ducklake_stats.cpp
```diff
@@ -99,8 +99,26 @@ DuckLakeColumnStats DuckLakeColumnStats::FromGlobalStats(const LogicalType &type
 	return stats;
 }
 
+// Cast a string-encoded stat to a new column type after ALTER COLUMN ... TYPE, so a same-transaction
+// retype+INSERT keeps min/max instead of discarding them. Callers must skip FLOAT sources: a float's
+// stored string is not bit-faithful when reparsed at a wider type (#1307), giving a value-wrong bound.
+static bool ReconcileStatToType(string &value, const LogicalType &target) {
```
pdet: Can we add the following as test? I belieev we are not covering TZ

```sql
CREATE TABLE t (id INTEGER, v TIMESTAMP);
INSERT INTO t SELECT i, TIMESTAMP '2020-06-15 12:00:00' + INTERVAL (i) SECOND FROM range(4096) t(i);
BEGIN;
ALTER TABLE t ALTER COLUMN v TYPE TIMESTAMPTZ;
INSERT INTO t SELECT 100000+i, TIMESTAMPTZ '2019-01-01 00:00:00+00' + INTERVAL (i) SECOND FROM range(4096) t(i);
COMMIT;
```

pdet (commented):
Thanks for the PR, I had one comment, and also can you keep comments max one line and self contained (i.e., not referencing something external), also only have comments if they are actually needed to understand the code/test.

src/storage/ducklake_stats.cpp
GavinPizza: Thank you, there was no TZ coverage here at all and adding it found a real bug where the stat was pinned to UTC and pruned live rows. This is now fixed along with the comments.

PR #1333, opened 2026-07-22
Resolve inlined-data-table membership per snapshot

### LINKED ISSUE #1305: BUG: stale inlined-data-table membership raises "ducklake_inlined_data_... does not exist" on a separately-attached reader after a concurrent flush
### What happens?

A DuckLake reader intermittently fails with:

```
Catalog Error: ... ducklake_inlined_data_<id>_<ver> does not exist
```

after another connection flushes inlined data, and a plain retry keeps failing. Only a fresh `ATTACH`/reconnect heals it.

This shows up when the reader is attached separately from the connection that flushed, two connections or processes over a shared metadata backend (Postgres, DuckDB-server). A reader on the same connection recovers on its own; a separately-attached one keeps failing on every retry until it re-`ATTACH`es. Root cause below.

### To Reproduce

Deterministic on `main` with data inlining on. Two DuckLake catalogs over one shared metadata store (two `ATTACH`es sharing `METADATA_CATALOG 'm'`) stand in for two independent connections: they share one metadata handle but are distinct catalog instances with distinct `instance_id`s, so the writer's cache invalidation never reaches the reader.

```python
import duckdb, tempfile
ws = tempfile.mkdtemp(prefix="ducklake-membership-")
con = duckdb.connect()

# writer and reader are two catalogs over the same metadata store 'm'
con.execute(f"ATTACH 'ducklake:{ws}/c.ducklake' AS dl_w (DATA_PATH '{ws}/d', DATA_INLINING_ROW_LIMIT 100, METADATA_CATALOG 'm')")
con.execute(f"ATTACH 'ducklake:{ws}/c.ducklake' AS dl_r (DATA_PATH '{ws}/d', DATA_INLINING_ROW_LIMIT 100, METADATA_CATALOG 'm')")

con.execute("CREATE TABLE dl_w.t (id INTEGER, v INTEGER)")

# seed two inline generations across a schema_version bump, so the flush has a
# superseded inlined table to drop: an ADD COLUMN bumps schema_version, so the
# second insert lands in a new inlined table and the first becomes superseded
con.execute("INSERT INTO dl_w.t VALUES (1, 10)")
con.execute("ALTER TABLE dl_w.t ADD COLUMN w INTEGER DEFAULT 0")
con.execute("INSERT INTO dl_w.t VALUES (2, 20, 200)")

# the reader (a separate catalog instance) caches membership {gen_1, gen_2}
print(con.execute("SELECT sum(v) FROM dl_r.t").fetchall())   # [(30,)]

# the writer flushes: both generations are materialized to parquet and the emptied,
# superseded gen_1 inlined table is dropped on an ordinary commit (no schema_version bump)
con.execute("CALL ducklake_flush_inlined_data('dl_w')")

# the reader's next read still uses its stale, un-invalidated membership
# BUG: Catalog Error: ... ducklake_inlined_data_1_1 does not exist
print(con.execute("SELECT sum(v) FROM dl_r.t").fetchall())
```

**Observed:** the second reader read fails with `ducklake_inlined_data_1_1 does not exist`. `DETACH dl_r` + re-`ATTACH` (a new `instance_id`) heals it - the reported reconnect-only behavior.

**Expected:** the reader recovers on its next transaction without a reconnect.

Membership is loaded by an unfiltered subquery in `BuildCatalogForSnapshot` (`src/storage/ducklake_metadata_manager.cpp`, `... FROM {METADATA_CATALOG}.ducklake_inlined_data_tables WHERE table_id = tbl.table_id`, around line 722) and baked into the `DuckLakeTableEntry` held by the schema cache entry, which is keyed on `schema_version` (`SchemaCacheKey` / `GetSchemaCacheEntry` in `src/storage/ducklake_catalog.cpp`). `ducklake_inlined_data_tables` has no `begin_snapshot`/`end_snapshot`, and `DropEmptySupersededInlinedTables` (`src/storage/ducklake_transaction_state.cpp` around line 1748) drops the superseded table without bumping `schema_version`, so a reader re-hits the stale `schema_version`-keyed entry.

The same-instance mitigation is the `InvalidateSchemaCache` sweep at the end of `DropEmptySupersededInlinedTables` (around line 1799, duplicated in `src/storage/ducklake_metadata_manager.cpp` around line 5376). Its cache key embeds a per-`ATTACH` `instance_id`, so it cannot reach a reader on a different catalog instance.

*(Line numbers on `34db89b2`, current `main`.)*

### OS:

### DuckDB Version:

### DuckLake Version:

### DuckDB Client:

Python

### Hardware:

Not performance-related.

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [X] Yes, I have

- [X] Yes, I have

### LINKED ISSUE #1413: Reader fails permanently with "ducklake_inlined_data_1_1 does not exist" after a concurrent ducklake_flush_inlined_data
### What happens?

`ducklake_flush_inlined_data` publishes its new snapshot before it removes the superseded inlined-data table. A reader that resolves inlined-table membership in that gap records a table name that is about to disappear, and then fails on every subsequent statement with `relation "ducklake_inlined_data_<id>_1" does not exist`.

On a PostgreSQL metadata catalog the gap runs to a few milliseconds per flush. Polling an unstalled flush over 15 runs, the new snapshot became visible 5.2-7.5 ms before the registry row was deleted, and 7.8-11.6 ms before the physical table was dropped. That is enough: a reader doing nothing but `SELECT count(*)` in a loop lands in it without any help.

In `DuckLakeTransactionState::Commit` (`src/storage/ducklake_transaction_state.cpp:1953-1961`) the snapshot is committed at `commit_connection()` and the cleanup runs afterwards:

```cpp
context.commit_connection();
...
if (flushed_inlined && !context.skip_drop_empty_inlined) {
    DropEmptySupersededInlinedTables(context);
}
```

Membership comes from a correlated subquery in `BuildCatalogForSnapshot` (`src/storage/ducklake_metadata_manager.cpp:760-764`). It is a bare `WHERE inlined_data_tables.table_id = tbl.table_id`. The two sibling subqueries beside it both filter on `{SNAPSHOT_ID}` against `begin_snapshot`/`end_snapshot`. This one cannot, because `ducklake_inlined_data_tables` has no such columns (`:276-277`). The catalog it builds is then cached per schema version, so what a reader sees depends on when it first built the catalog for that version.

The closing edge is the registry `DELETE`, not the `DROP`. `drops_sql` emits `DELETE FROM ducklake_inlined_data_tables ...` followed by `DROP TABLE ...`, and on PostgreSQL those do not commit atomically. Stalling the `DROP` with an event trigger reproduces nothing, because membership is already clean by then.

### To Reproduce

Two processes against one PostgreSQL metadata catalog. Writer:

```sql
ATTACH 'ducklake:postgres:dbname=dlwindow host=/var/run/postgresql' AS lake (DATA_PATH '/tmp/f/', DATA_INLINING_ROW_LIMIT 100);
CREATE TABLE lake.t(a INTEGER);
INSERT INTO lake.t VALUES (1);
ALTER TABLE lake.t ADD COLUMN b INTEGER;   -- bumps schema_version, so generation 1 is superseded
INSERT INTO lake.t VALUES (2,2);
CALL ducklake_flush_inlined_data('lake');
```

Reader, in a separate process, looping `SELECT count(*) FROM lake.t` across the flush. Three runs of 8000 reads each on `main` @ 3d8e24a8 gave 7560, 7605 and 7613 failed reads, every one of them:

```
Invalid Error: Failed to read inlined data from DuckLake: ... COPY (SELECT ... FROM "public"."ducklake_inlined_data_1_1") ...:
ERROR:  relation "public.ducklake_inlined_data_1_1" does not exist
```

The failures begin a few hundred reads in and continue to the end of the run without ever clearing.

To make it deterministic instead of racing, hold the window open from a third session. `EXCLUSIVE` blocks the writer's post-commit `DELETE` but not the reader's `SELECT` of the same table, and table ownership is enough, so no superuser is involved:

```sql
BEGIN; LOCK TABLE ducklake_inlined_data_tables IN EXCLUSIVE MODE; SELECT pg_sleep(20); COMMIT;
```

With the writer parked there, a fresh reader process reads once inside the window, waits for the lock to release, and reads again. The second read fails, as does every read after it, in three runs out of three.

To rule out the reader just sitting on an old snapshot: commit an unrelated `INSERT` from a third process after the flush. It advances `snapshot_id` without bumping `schema_version`, and on `main` the reader never sees it.

### OS:

Linux x86_64 (Debian 12)

### DuckDB Version:

v1.6.0-dev12414

### DuckLake Version:

main @ 3d8e24a8

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Gavin Brand

### Affiliation:

_No response_

- [X] Yes, I have

- [X] Yes, I have

GavinPizza:
Closes #1305

Inlined-table membership is a field on the table entry, so it is frozen inside the schema cache entry, which the `ObjectCache` keys on `schema_version`. A flush can drop a superseded inlined table on a commit that advances `snapshot_id` without bumping `schema_version`, so a reader that re-hits that entry gets a table name the flush already dropped and fails with `ducklake_inlined_data_<id>_<ver> does not exist`. The existing mitigation (`InvalidateSchemaCache` after a drop) is keyed on a per-`ATTACH` `instance_id`, so it only heals readers on the same instance. A reader on a separate attach or process, which is the normal case with a shared metadata backend, keeps failing until it reconnects.

This resolves membership per snapshot instead of per `schema_version`. A new `DuckLakeCatalog::GetInlinedDataTables(transaction, table)` reads the live `ducklake_inlined_data_tables` state. Where `snapshot_id` advances between statements, the result is cached under a `snapshot_id` key. Where it cannot, the result is memoized for the statement only. The four read sites in `ducklake_multi_file_list.cpp` and the cardinality path `GetNetInlinedRowCount` now go through it. The per-instance sweep stays because the commit-time conflict check and the two write paths still read the entry's copy; routing those through the helper too is a follow-up. The read path is unchanged, measured on a read-only loop and on an insert-heavy one, and readers that fall through to the memo pay one extra metadata query per statement.

Adds regression tests under `test/sql/issues/`. The cross-instance test attaches one shared metadata store twice (distinct `instance_id`s, standing in for two processes) and is red on `main`, green here. The pinned-reader, open-transaction and shared-attach tests are red on the first commit of this branch and green on the second. The shared-attach one also passes on `main`: the first commit regressed that case and the second takes it back.

This does not close #1413, the pre-existing window between the commit and the drop, though resolving per snapshot does narrow it: on `main` such a reader never recovers, while here it ages out on the next snapshot.


src/storage/ducklake_catalog.cpp
```diff
@@ -384,6 +393,25 @@ void DuckLakeCatalog::PinSchemaForQuery(DuckLakeTransaction &transaction, shared
...
+		return table.GetInlinedDataTables();
+	}
+	auto table_id = table.GetTableId();
+	auto &cache = GetObjectCacheInstance();
+	// Key on the transaction's committed snapshot, not any per-read time-travel target.
+	auto key = InlinedDataTablesCacheKey(transaction.GetSnapshot().snapshot_id, table_id);
+	auto cached = cache.Get<DuckLakeInlinedDataTablesCacheEntry>(key);
+	if (cached) {
+		return cached->inlined_data_tables;
+	}
```
pdet, line 409: Is the cache being invalidated for a superseeded table drop?

e.g. (on postgres/sqlite),
```sql
 statement ok con_w
  ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS dl (DATA_PATH '{DATA_PATH}/m1', DATA_INLINING_ROW_LIMIT 100)

  statement ok con_w
  CREATE TABLE dl.t (id INTEGER, v INTEGER)

  statement ok con_w
  INSERT INTO dl.t VALUES (1, 10)

  ALTER TABLE dl.t ADD COLUMN w INTEGER DEFAULT 0

  statement ok con_w
  INSERT INTO dl.t VALUES (2, 20, 200)

  statement ok con_r
  BEGIN TRANSACTION

  query I con_r
  SELECT sum(v) FROM dl.t
  ----
  30

  statement ok con_w
  CALL ducklake_flush_inlined_data('dl')

  query I con_r
  SELECT sum(v) FROM dl.t
  ----
  30

```

src/storage/ducklake_catalog.cpp
```diff
@@ -384,6 +393,25 @@ void DuckLakeCatalog::PinSchemaForQuery(DuckLakeTransaction &transaction, shared
...
+	auto inlined_data_tables = transaction.GetMetadataManager().GetInlinedDataTablesForTable(table_id);
+	cache.Put(std::move(key), make_shared_ptr<DuckLakeInlinedDataTablesCacheEntry>(inlined_data_tables));
```
pdet, line 411: Should we do one key per table?
For example, what happens in here?
```sql

statement ok
CREATE TABLE dl.t (i INTEGER)

loop i 0 1000

statement ok
INSERT INTO dl.t VALUES (${i})

statement ok
SELECT count(*) FROM dl.t

endloop
```

pdet (commented):
Thanks for the PR, I had two comments.

src/storage/ducklake_catalog.cpp
GavinPizza: It is not invalidated because the drop runs after `commit_connection()` rather than as part of the commit, so there's a window where the new snapshot is visible but the old inlined table hasn't gone yet. A transaction starting in that window caches a table that's about to be dropped.

Within a single transaction it is fine, because `GetConnection` opens its own metadata connection and holds a transaction on it, so on Postgres that's REPEATABLE READ for the life of the DuckLake transaction.

I ran your script on all three catalogs: DuckDB passes, SQLite fails at the flush with "database is locked", and Postgres blocks on the DROP. I have not managed to build a deterministic repro for the stale read itself.

src/storage/ducklake_catalog.cpp
GavinPizza: I believe you mean the snapshot_id as the key is already per table. It does churn and leave 1000 entries that are never read again but this is in line with other functions in the codebase such as the stats cache StatsCacheKey right next to it, which is keyed the same way: the instance, an advancing id (next_file_id), then the table id.

I ran your loop on main and observed it is already growing the object cache from 2 KB to 330 KB. My changes are adding an additional 129 KB on top of that which I think is acceptable compared to the alternative of memoizing on the transaction instead of the object cache. That alternative removes the churn entirely and both membership tests pass, but it costs about 10% on read-only workloads because every auto-commit statement gets a fresh transaction. I can switch to that instead if that is the direction you prefer.

GavinPizza: The key cannot lose `snapshot_id` because without it one entry per table would outlive the flush that drops the table it names, which is the bug this PR is fixing.

The churn is unchanged by the second commit I pushed, since your loop is auto-commit and still takes the cache. What I did not say last time is that those entries are evictable by the buffer pool, because `GetEstimatedCacheMemory` returns a real size, so the growth is bounded by the LRU rather than by the key.

On the question I left open, I ended up not routing every reader through a transaction memo. Only readers whose `snapshot_id` cannot advance, meaning one inside an explicit transaction or one on a pinned attach, now skip the cache. Your loop shows no measurable difference against the first commit, and the object cache ends at the same size. The 10 percent I quoted was a read-only cost, and gating is what keeps it off the readers that can use the cache safely.

src/storage/ducklake_catalog.cpp
GavinPizza: Nothing deletes `InlinedDataTablesCacheKey`. Only `StatsCacheKey` and `SchemaCacheKey` have an invalidate path (`ducklake_catalog.cpp:1199,1203`), so it is not the post-commit ordering alone that leaves the entry stale. The entry is meant to stop being consulted once `snapshot_id` advances, and the second commit takes the readers whose `snapshot_id` cannot advance out of that cache entirely.

Your script as written passes on `main`, and that is the per-instance sweep working: one `ATTACH` means one `instance_id`, so `InvalidateSchemaCache` after the drop does reach the reader. Give the reader its own `ATTACH` of the same store and `main` fails at your second read, the one expecting 30, with `no such table: ducklake_inlined_data_1_1`. That read is inside the transaction, so "within a single transaction it is fine" last time only held for a metadata catalog with transaction-lifetime MVCC.

That form is on the branch as `issue_inlined_membership_open_transaction.test`, and its header covers the two conditions it needs, since SQLite wants WAL and a DuckDB metadata catalog will not show this at all. Your single-attach form and a pinned attach are tested too. The single-attach one passes on `main` and is there because the first commit regressed it.

That window I described is real and measured, but it is a separate, pre-existing bug rather than this one. Filed as #1413.

jdctinuiti:
Another reproduction of #1305, with two separate processes against one `quack_serve` metadata server and no open transaction in either. With `DATA_INLINING_ROW_LIMIT 100`: create a table, insert two rows, `ALTER TABLE ... ADD COLUMN`, insert two more rows, and read the table from a reader process. After a writer process runs `ducklake_flush_inlined_data`, every later read in the reader fails with `Table with name ducklake_inlined_data_1_1 does not exist!` until it re-attaches, while the writer keeps reading normally.

PR #1336, opened 2026-07-22
Tolerate missing mtime in orphaned file cleanup

### LINKED ISSUE #1042: `CHECKPOINT` fails with `KeyError: 'mtime'` when using GCS-buckets
### Description

Calling `CHECKPOINT` on a DuckLake catalog fails with the following error:

```
Failed to perform CHECKPOINT; in DuckLake: Failed to get files scheduled for deletion from DuckLake: KeyError: 'mtime'
```
GCS does not always expose mtime — specifically, composite objects in GCS don't support this in the same way as regular objects, and when mtime is unavailable, tools relying on it have no fallback.

### Steps to reproduce

```python
import duckdb
import gcsfs

conn = duckdb.connect()

# Register GCS filesystem
fs = gcsfs.GCSFileSystem()
conn.register_filesystem(fs)

# Load extensions
for ext in ("ducklake", "postgres"):
    conn.sql(f"INSTALL {ext}")
    conn.sql(f"LOAD {ext}")

# Attach DuckLake catalog backed by PostgreSQL + GCS
conn.sql("""
    ATTACH 'ducklake:postgres:
        dbname=mydb
        user=myuser
        host=localhost
    ' AS my_catalog
    (DATA_PATH 'gs://my-bucket/data/',
     METADATA_SCHEMA my_schema,
     DATA_INLINING_ROW_LIMIT 300,
     AUTOMATIC_MIGRATION TRUE);
""")
conn.sql("USE my_catalog")

# Create a table and insert some data
conn.sql("CREATE TABLE test (id INTEGER, name VARCHAR)")
conn.sql("INSERT INTO test VALUES (1, 'hello'), (2, 'world')")

# Triggers the bug
conn.execute("CHECKPOINT")
```

### OS:

Linux

### DuckDB Version:

1.5.2

### DuckLake Version:

1.0

### DuckDB Client:

Python 3.13

### Hardware:

_No response_

### Full Name:

Vidar Strandseter

### Affiliation:

Statistics Norway

No - I cannot share the data sets because they are confidential

- [x] Yes, I have

- [x] Yes, I have

u70b3:
## Problem

On GCS with a registered `gcsfs` filesystem, `CHECKPOINT` can abort with:

```
Failed to perform CHECKPOINT; in DuckLake: Failed to get files scheduled
for deletion from DuckLake: KeyError: 'mtime'
```

Closes #1042.

## Root cause

`CHECKPOINT` calls `ducklake_delete_orphaned_files` without arguments, so orphan enumeration runs in `older_than` mode and filters `read_blob(...).last_modified` against the configured retention period.

For a registered fsspec filesystem this calls `FileSystem::GetLastModifiedTime`. `gcsfs.modified()` indexes `info()["mtime"]`, but `mtime` is absent for some GCS object metadata (including the composite-object case from #1042) and synthesized directory entries. The resulting Python `KeyError: 'mtime'` aborts the query and therefore the checkpoint.

## Fix

`GetOrphanFilesForCleanup` now selects the enumeration strategy by cleanup mode:

- `cleanup_all` uses `glob`, which lists paths without requiring per-object modification metadata.
- `older_than` keeps `read_blob`, because the timestamp comparison genuinely requires `last_modified`. The narrowly scoped missing-mtime failure retains all orphan candidates for that cleanup round instead of aborting `CHECKPOINT`. Other authentication, network, and storage errors still surface.

Retaining files whose age cannot be established is the conservative direction: it cannot over-delete data, and those files can still be reclaimed with `cleanup_all`.

The compatibility fallback requires both `ExceptionType::INVALID` and the current gcsfs `KeyError: 'mtime'` signature, and uses `RawMessage()` rather than the formatted error. It is intentionally temporary. The Python filesystem boundary currently lets the Python exception reach DuckDB as `ExceptionType::INVALID`, so DuckLake has no provider-independent error code to inspect. duckdb/duckdb-python#584 tracks translating unsupported modification metadata to `ExceptionType::NOT_IMPLEMENTED`, which DuckDB core already converts to a `NULL` `last_modified` value. Once available in supported DuckDB versions, this fallback can be removed.

No DuckLake metadata schema or specification change is required.

### CI regression test

`test/sql/remove_orphans/puffin_orphan_minio.test`, gated by `S3_TEST_SERVER_AVAILABLE`, covers the `cleanup_all -> glob` path on MinIO:

- creates a live Puffin deletion vector;
- copies it to an unreferenced orphan path;
- verifies dry-run discovery and deletion of the orphan;
- verifies the referenced deletion vector and table semantics remain intact.

The MinIO test does not claim to reproduce the Python/gcsfs missing-mtime exception: native httpfs/S3 supplies modification metadata.

### Python + gcsfs end-to-end verification

The actual failing path was separately verified with DuckDB 1.5.5, gcsfs 2026.7.0 registered through `register_filesystem`, and an in-process mock of the GCS JSON API whose object resources can omit `updated`:

| operation | released extension | with this fix |
|---|---|---|
| `read_blob` with `last_modified` filter | `KeyError: 'mtime'` | unchanged core behavior; tolerated by DuckLake |
| `CHECKPOINT` / default `older_than` cleanup | aborts with `KeyError: 'mtime'` | completes and retains candidates |
| explicit `older_than` cleanup | aborts with `KeyError: 'mtime'` | completes and retains candidates |
| `cleanup_all` | completes | deletes the mtime-less orphan through `glob` |

The mock reproduces the gcsfs-visible contract rather than using a real GCS account; a real-bucket smoke test is still welcome.

### Latest local verification

- formatting check: passed;
- `ducklake_metadata_manager.cpp` target object: compiled successfully against the checked-out DuckDB headers;
- the previous full PR CI matrix, including MinIO and all distribution builds, passed.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -4864,6 +4864,14 @@ FROM {METADATA_CATALOG}.ducklake_files_scheduled_for_deletion f
 	return StringUtil::Replace(query, "{SEPARATOR}", separator);
 }
 
+//! Returns true for the specific "backend cannot supply per-object mtime" failure we tolerate
+//! during older_than orphan cleanup (ducklake#1042): a filesystem without mtime support (e.g.
+//! gcsfs reading a GCS composite object) surfaces this exact KeyError from its info() dictionary.
+//! Any other enumeration error (auth, network, ...) must still surface loudly.
+static bool IsMissingMTimeError(const ErrorData &error) {
+	return StringUtil::Contains(error.Message(), "KeyError: 'mtime'");
+}
+
 vector<DuckLakeFileForCleanup> DuckLakeMetadataManager::GetOrphanFilesForCleanup(const string &filter,
                                                                                  const string &separator) {
 	auto known_files_query = GetKnownFilesForCleanupQuery(separator);
@@ -4902,18 +4910,61 @@ vector<DuckLakeFileForCleanup> DuckLakeMetadataManager::GetOrphanFilesForCleanup
 		}
 		appender.Close();
 
-		auto query = StringUtil::Format(R"(SELECT filename
+		// Enumerate orphan candidate files. When no timestamp filter is requested
+		// (cleanup_all mode), enumerate with `glob` rather than `read_blob`: glob
+		// only lists entries and never touches per-object metadata, so this mode
+		// does not depend on the storage backend supplying `mtime`.
+		//
+		// When a timestamp filter is requested (older_than mode - this includes
+		// CHECKPOINT, which runs ducklake_delete_orphaned_files with the default
+		// 'delete_older_than' setting), the `last_modified` comparison requires
+		// per-object mtime. Some backends cannot supply it for every object (e.g.
+		// gcsfs raises KeyError: 'mtime' for GCS composite objects, see
+		// ducklake#1042), which aborts the query - and with it the whole
+		// CHECKPOINT. That specific capability failure is tolerated by degrading
+		// to "retain all" for this round: skipping cleanup of orphans whose age
+		// cannot be determined is always safe (never over-delete), and they can
+		// still be removed via cleanup_all, which does not need mtime. Any other
+		// enumeration error (auth, network, ...) is rethrown - silently
+		// suppressing those would disable orphan cleanup without any signal.
+		unique_ptr<QueryResult> res;
+		if (filter.empty()) {
+			auto query = StringUtil::Format(R"(SELECT file
+FROM glob({DATA_PATH} || '**') files
+WHERE (suffix(file, '.parquet') OR suffix(file, '.puffin'))
+AND NOT EXISTS (
+	SELECT 1 FROM %s known_files WHERE known_files.full_path = REPLACE(files.file, '\', '/')
+))",
+			                                temp_table_identifier);
+			SubstituteCatalogPlaceholders(query);
+			res = transaction.ExecuteRaw(query);
+			if (res->HasError()) {
+				res->GetErrorObject().Throw("Failed to get files scheduled for deletion from DuckLake: ");
+			}
+		} else {
+			auto query = StringUtil::Format(R"(SELECT filename
 FROM read_blob({DATA_PATH} || '**') files
 WHERE (suffix(filename, '.parquet') OR suffix(filename, '.puffin'))
 AND NOT EXISTS (
 	SELECT 1 FROM %s known_files WHERE known_files.full_path = REPLACE(files.filename, '\', '/')
 )
 %s)",
-		                                temp_table_identifier, filter);
-		SubstituteCatalogPlaceholders(query);
-		auto res = transaction.ExecuteRaw(query);
-		if (res->HasError()) {
-			res->GetErrorObject().Throw("Failed to get files scheduled for deletion from DuckLake: ");
+			                                temp_table_identifier, filter);
+			SubstituteCatalogPlaceholders(query);
+			res = transaction.ExecuteRaw(query);
+			if (res->HasError()) {
+				auto &error = res->GetErrorObject();
+				if (IsMissingMTimeError(error)) {
+					// mtime-dependent enumeration failed (e.g. objects without mtime
+					// on the storage backend). Retain all orphans this round;
+					// CHECKPOINT must stay usable. The aged-out orphans will be
+					// picked up on a later run once the backend can supply mtime,
+					// or via cleanup_all.
+					drop_temp_table();
+					return {};
+				}
+				error.Throw("Failed to get files scheduled for deletion from DuckLake: ");
+			}
 		}
 
 		vector<DuckLakeFileForCleanup> result;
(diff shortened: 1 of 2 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -4866,6 +4864,14 @@ FROM {METADATA_CATALOG}.ducklake_files_scheduled_for_deletion f
 	return StringUtil::Replace(query, "{SEPARATOR}", separator);
 }
 
+//! Returns true for the specific "backend cannot supply per-object mtime" failure we tolerate
+//! during older_than orphan cleanup (ducklake#1042): a filesystem without mtime support (e.g.
+//! gcsfs reading a GCS composite object) surfaces this exact KeyError from its info() dictionary.
+//! Any other enumeration error (auth, network, ...) must still surface loudly.
```
dentiny, line 5071: just curious, we don't have a better way to these errors? for example, error code? I'm worried we will have to add more and more similar util functions

u70b3: Good point. I traced the exception boundary, and there is no usable provider-specific error code at the DuckLake layer today: `PythonFilesystem::GetLastModifiedTime()` forwards the fsspec Python exception, and by the time the query returns an `ErrorData` it has `ExceptionType::INVALID`.

I agree that matching messages in downstream extensions should not become the general pattern. I opened duckdb/duckdb-python#584 to translate unsupported modification metadata at the Python filesystem boundary into `ExceptionType::NOT_IMPLEMENTED`. DuckDB core already turns that type into a `NULL` `last_modified`, so `older_than` would then skip only the affected entries and this helper could be removed.

For compatibility with current released versions, I kept the narrow fallback here but tightened it to require `ExceptionType::INVALID` and to inspect `RawMessage()` rather than the formatted message. I also added a TODO linked to the upstream issue. All unrelated error types still propagate.

carlopi (commented):
I have been asked to take a look, but this sounds more like a duckdb/duckdb side issue? Like possibly having read_blob provide time only when required?

I can see this working, but it does add surface, and it's not really tight, since any filter would bring back the error.

Maybe one path would avoid duplicating the SQL logic, by restructuring the two queries so to avoid future reworking having to duplicate logic.

u70b3:
Thanks for taking a look!

SQL duplication — fixed in cc3aea08: both modes now build from a single shared query skeleton (`GetOrphanCandidateQuery`) that differs only in the enumeration source (`glob`/`file` vs `read_blob`/`filename`) and the trailing timestamp filter, with one shared execution path after that.

On the layering question — agreed the durable fix belongs upstream, and that chain is already in motion: duckdb/duckdb-python#584 (now picked up) translates the fsspec `KeyError: 'mtime'` into `ExceptionType::NOT_IMPLEMENTED` at the Python filesystem boundary, and core's `read_blob` already maps a `NOT_IMPLEMENTED` `GetLastModifiedTime` to a NULL `last_modified` (`direct_file_reader.cpp`), so `older_than` would then degrade to per-row NULLs instead of aborting. One note on "provide time only when required": the fetch is already projection-driven in core — the reason it still fires here is that the `older_than` filter genuinely requires the column, so the `NOT_IMPLEMENTED` → NULL mapping is the piece that actually unblocks this path.

On the remaining string-matching fallback: the only reason it exists is compatibility. Released DuckDB versions will never get the upstream translation, and the extension keeps supporting those versions — for gcsfs users on them, `CHECKPOINT` aborts today without it. The match is fenced to `ExceptionType::INVALID` + this exact signature, all other errors still throw, the degradation direction is retain-all (never over-delete), and the helper carries a TODO linking the upstream issue for removal once the translation has propagated to supported versions.

If you'd rather not carry the fallback at all, the alternative is to keep only the `cleanup_all` → `glob` change here (strict improvement, no message matching) and leave `older_than`/`CHECKPOINT` broken for gcsfs users on current releases until the upstream fix propagates. Happy to go either way — which do you prefer?

u70b3:
Friendly bump — this has been quiet for about three weeks, so a quick summary of where things stand:

- The SQL-duplication feedback is addressed in d43efd86: both cleanup modes now build from a single shared orphan-candidate query skeleton, differing only in the enumeration source (`glob` vs `read_blob`) and the trailing timestamp filter.
- On the layering question: the durable fix is in motion upstream — duckdb/duckdb-python#584 translates the fsspec missing-mtime failure to `ExceptionType::NOT_IMPLEMENTED`, which core already degrades to a NULL `last_modified`. The string-matching fallback here only covers released DuckDB versions that will never receive that translation; it is narrowly fenced and fails toward retaining files, never over-deleting. Still happy to drop it and keep only the `cleanup_all` → `glob` change if that trade-off is preferred.
- CI is green except for 4 jobs that were cancelled (not failed) on the last run — re-running them would need someone with write access.

@carlopi @dentiny could you take another look when you have a moment? @pdet pinging you as an active maintainer in case this fell off the radar — this unbreaks `CHECKPOINT` for gcsfs users (#1042).

carlopi:
Hey @u70b3, thanks a ton for the PR, but after discussion with maintainer, but I feel this is still not the appropriate place for a fix, and this should be solved at FileSystem level, that means on the python layer.

Sorry it took a while to come back on this.

PR #1344, opened 2026-07-25
Materialize the file list before the scan fans out across threads

wmTJc9IK0Q:
DuckLakeMultiFileList populates itself lazily, so the first thing every scan task does is call GetFiles() and block on file_lock while one of them runs the metadata query. With DuckDB 2.0's multi-file read-ahead the waiting tasks do not just block once - they repeatedly park and wake trying to produce read-ahead jobs, which turns a sub-millisecond scan into ~20ms of kernel time.

Materialize the list in InitializeGlobalState instead, where we are still single-threaded and hold the pruned list. Same metadata queries, but the scan tasks now find the list ready and significantly reduce context switching and waiting.

On a single-partition scan (one 20k-row file, threads=16):

```
  sys time     4.1s -> 0.52s   (200 queries)
  ctx switches 240k -> 41k
  total CPU    27.5ms -> 8ms per query
```

Measured no change in metadata query counts for count(*), LIMIT 1, fully-pruned, WHERE false, single-partition and full scans, and no regression on 2000-file, 20M-row or 326-row-group scans.


pdet:
Thanks, LGTM!

pdet:
@wmTJc9IK0Q seems like the formatter is failing, could you have a look?

wmTJc9IK0Q:
@pdet done - also confirmed the test failure was just flaky

pdet:
Hi @wmTJc9IK0Q, it seems that CI is failing, and there are merge conflicts, could you have a look?

wmTJc9IK0Q:
Merged main and fixed the quack `table_changes` failures.

Materializing the file list in `InitializeGlobalState` means a query with several DuckLake scans (`table_changes` expands to four) now runs its metadata queries concurrently. The quack metadata manager recovers from a failed statement with `ROLLBACK; BEGIN TRANSACTION`, and a statement from another thread landing between the failure and the rollback failed with "Current transaction is aborted". `Query()` now holds a lock across the statement and its recovery.

Full quack suite and the standard test suite pass locally, and CI is green.

pdet:
Thanks!

PR #1345, opened 2026-07-25
Cache the begin snapshot of a committed schema version

wmTJc9IK0Q:
Every scan resolves the begin snapshot of each of the table's schema versions, which costs one metadata query per schema version. The backing row in ducklake_schema_versions is written once when the schema version is created and is never updated, so the mapping is permanent and can be cached on the catalog.

Only rows observed by a transaction that has not written anything are cached: a schema version created by the current transaction can still be rolled back, and the GetBeginSnapshotForTable fallback taken when no row exists yet is not stable either.

Takes a partition-filtered scan of a two-schema-version table from 6 metadata queries down to 4, worth ~10% of wall and CPU on a loop of 200 such scans (0.96s -> 0.87s wall, 1.10s -> 0.98s user).

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -290,6 +290,12 @@ public:
 	//! Cache the result of an inlined deletion table existence check
 	void CacheInlinedDeletionTableResult(TableIndex table_id, DuckLakeSnapshot snapshot, bool exists);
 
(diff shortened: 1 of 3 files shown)


pdet:
Thanks, LGTM

pdet:
@wmTJc9IK0Q seems the formatter CI is failing, could you have a look?

wmTJc9IK0Q:
@pdet fixed, thanks

PR #1346, opened 2026-07-26
Make a repeated identical SET SORTED BY a no-op

wmTJc9IK0Q:
ALTER TABLE ... SET SORTED BY unconditionally builds a new DuckLakeSort and
returns a new catalog entry, so re-issuing the same clause writes a fresh
sort_id, one ducklake_sort_expression row per key, and a new snapshot every
time — even though nothing about the table changed.

That is a real cost for any application that declares its schema on startup.
Re-running a setup script over an existing catalog grows ducklake_snapshot and
ducklake_sort_expression forever: 13 sorted tables meant ~13 snapshots and ~29
metadata rows per process start, none of which describe a change.

SET PARTITIONED BY already avoids this: AlterTable compares the requested
DuckLakePartition against the current one via PartitionFieldsMatch and returns
nullptr when they are equal. This mirrors that for sort orders, comparing
sort_key_index, expression, dialect, sort_direction and null_order.

Comparing null_order needs one normalization first. The parser leaves
null_order as ORDER_DEFAULT when the clause does not spell it out, while the
value read back from the catalog is always NULLS_FIRST or NULLS_LAST, so
`SET SORTED BY (a ASC)` would never compare equal to itself. Resolve it at
build time the same way the metadata writer and the ORDER BY builder already
do, which also makes the stored sort_field independent of how the clause was
written.

RESET SORTED BY gets the same treatment: clearing a sort order that is already
clear now returns nullptr instead of writing a snapshot.


Alex-Monahan:
I thought this was already working, but perhaps not! Can you share a reproduction of the problem?

wmTJc9IK0Q:
Sure — here is a standalone one. No build needed for the first run: `duckdb` v1.5.2 with `INSTALL ducklake` reproduces it, and so does a source build of `main`.

The shape is an application that declares its schema on every start, so the same DDL runs again over an existing catalog. `CREATE TABLE IF NOT EXISTS` and `SET PARTITIONED BY` absorb that; `SET SORTED BY` does not.

```sql
LOAD ducklake;
ATTACH 'ducklake:x.ducklake' AS lk (DATA_PATH 'x_files/');

-- boot 1
CREATE TABLE IF NOT EXISTS lk.t (id INTEGER, ts TIMESTAMP);
ALTER TABLE lk.t SET PARTITIONED BY (id);
ALTER TABLE lk.t SET SORTED BY (ts);

-- boot 2 — byte-identical script, nothing about the table has changed
CREATE TABLE IF NOT EXISTS lk.t (id INTEGER, ts TIMESTAMP);
ALTER TABLE lk.t SET PARTITIONED BY (id);
ALTER TABLE lk.t SET SORTED BY (ts);

-- boot 3
CREATE TABLE IF NOT EXISTS lk.t (id INTEGER, ts TIMESTAMP);
ALTER TABLE lk.t SET PARTITIONED BY (id);
ALTER TABLE lk.t SET SORTED BY (ts);

SELECT snapshot_id, changes FROM ducklake_snapshots('lk') ORDER BY snapshot_id;
SELECT
    (SELECT count(*) FROM __ducklake_metadata_lk.ducklake_partition_info)  AS partition_info_rows,
    (SELECT count(*) FROM __ducklake_metadata_lk.ducklake_sort_info)       AS sort_info_rows,
    (SELECT count(*) FROM __ducklake_metadata_lk.ducklake_sort_expression) AS sort_expression_rows;
```

### On `main`

`fd4d1866` ("Make identical partition changes a no-op") already handles the partition side, so the contrast is clean: three `SET PARTITIONED BY` produce **one** snapshot, three `SET SORTED BY` produce **three**.

```
┌─────────────┬───────────────────────────┐
│ snapshot_id │          changes          │
├─────────────┼───────────────────────────┤
│ 0           │ {schemas_created=[main]}  │
│ 1           │ {tables_created=[main.t]} │
│ 2           │ {tables_altered=[1]}      │   <- boot 1, SET PARTITIONED BY
│ 3           │ {tables_altered=[1]}      │   <- boot 1, SET SORTED BY
│ 4           │ {tables_altered=[1]}      │   <- boot 2, SET SORTED BY
│ 5           │ {tables_altered=[1]}      │   <- boot 3, SET SORTED BY
└─────────────┴───────────────────────────┘
┌─────────────────────┬────────────────┬──────────────────────┐
│ partition_info_rows │ sort_info_rows │ sort_expression_rows │
│                   1 │              3 │                    3 │
└─────────────────────┴────────────────┴──────────────────────┘
```

Each redundant statement retires the previous `sort_id` and writes a new one plus a fresh row per sort key, so the tail of `ducklake_sort_info` is a chain of generations that all say the same thing:

```
┌─────────┬──────────┬────────────────┬──────────────┐
│ sort_id │ table_id │ begin_snapshot │ end_snapshot │
├─────────┼──────────┼────────────────┼──────────────┤
│       3 │        1 │              3 │            4 │
│       4 │        1 │              4 │            5 │
│       5 │        1 │              5 │         NULL │
└─────────┴──────────┴────────────────┴──────────────┘
```

### With this PR

```
┌─────────────┬───────────────────────────┐
│ snapshot_id │          changes          │
├─────────────┼───────────────────────────┤
│ 0           │ {schemas_created=[main]}  │
│ 1           │ {tables_created=[main.t]} │
│ 2           │ {tables_altered=[1]}      │   <- boot 1, SET PARTITIONED BY
│ 3           │ {tables_altered=[1]}      │   <- boot 1, SET SORTED BY
└─────────────┴───────────────────────────┘
┌─────────────────────┬────────────────┬──────────────────────┐
│ partition_info_rows │ sort_info_rows │ sort_expression_rows │
│                   1 │              1 │                    1 │
└─────────────────────┴────────────────┴──────────────────────┘
```

(Before/after are the same source tree with only `src/storage/ducklake_table_entry.cpp` reverted to the parent commit, so nothing else differs.)

### RESET

Same story for clearing a sort order that is already clear:

```sql
ATTACH 'ducklake:y.ducklake' AS ly (DATA_PATH 'y_files/');
CREATE TABLE ly.u (id INTEGER);
ALTER TABLE ly.u RESET SORTED BY;
ALTER TABLE ly.u RESET SORTED BY;
SELECT snapshot_id, changes FROM ducklake_snapshots('ly') ORDER BY snapshot_id;
```

`main` produces two `tables_altered` snapshots on a table that never had a sort order; with the PR it produces none.

### On the released extension

Worth flagging in case you tested there: `INSTALL ducklake` on DuckDB v1.5.2 gives `415a9ebd`, which predates `fd4d1866`, so the same script yields **six** `tables_altered` snapshots — the partition side churns too. The sort side is distinguishable even there, because the duplicated metadata rows only happen for sorts: `partition_info_rows = 1` but `sort_info_rows = sort_expression_rows = 3`.

### On the `null_order` normalization

It is load-bearing rather than cosmetic: the parser leaves `null_order = ORDER_DEFAULT` when the clause does not spell it out, while the value read back from the catalog is always `NULLS_FIRST`/`NULLS_LAST`, so without resolving it at build time `SET SORTED BY (ts)` never compares equal to itself after a reattach. It is not over-eager either — equivalent spellings collapse, genuine changes do not:

```
after SET SORTED BY (ts)                  -> snapshot 2
after SET SORTED BY (ts ASC)              -> snapshot 2
after SET SORTED BY (ts ASC NULLS LAST)   -> snapshot 2
after SET SORTED BY (ts DESC)             -> snapshot 3
after SET SORTED BY (ts DESC NULLS FIRST) -> snapshot 4
```

`test/sql/sorted_table/set_sorted_by_idempotent.test` in this PR covers the same ground as an sqllogictest (snapshot counts, metadata row counts, direction/null-order changes still taking effect, and `RESET`).

pdet:
@Alex-Monahan can you review this when you have the time? Many thanks!

wmTJc9IK0Q:
Merged main and fixed the format failure. Also skipped `set_sorted_by_idempotent.test` on the quack catalog - its `sort_metadata()` macro reads three metadata tables in one query, which hits the same "multiple streaming scans" limitation the other `sorted_table` metadata tests are already skipped for.

PR #1353, opened 2026-08-01
Reject data files that violate NOT NULL constraints in add_data_files

### LINKED ISSUE #1295: BUG: able to import parquet file which breaks DuckLake table constraint
### What happens?

Hi team, I found when we adding data files (parquet files) to DuckLake tables, it doesn't fail when parquet file breaks table constraint.

### To Reproduce

```sql
# name: test/sql/add_files/add_files_not_null_constraints.test
# description: ducklake_add_data_files should reject files that violate NOT NULL constraints
# group: [add_files]

require ducklake

require parquet

test-env DATA_PATH {TEST_DIR}

statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (
  DATA_PATH '{DATA_PATH}/ducklake_add_files_not_null_constraints',
  DATA_INLINING_ROW_LIMIT 0
)

statement ok
COPY (
  SELECT * FROM (VALUES (1, 10), (2, NULL)) v(id, x)
) TO '{DATA_PATH}/add_files_not_null_scalar.parquet' (FORMAT parquet)

statement ok
CREATE TABLE ducklake.scalar_not_null(id INTEGER, x INTEGER NOT NULL)

statement error
CALL ducklake_add_data_files('ducklake', 'scalar_not_null', '{DATA_PATH}/add_files_not_null_scalar.parquet')
----
NOT NULL constraint
```
gets error message
```sh
1. test/sql/add_files/add_files_not_null_constraints.test:27
Query unexpectedly succeeded! (test/sql/add_files/add_files_not_null_constraints.test:27)!
CALL ducklake_add_data_files('ducklake', 'scalar_not_null', 'duckdb_unittest_tempdir/71382/add_files_not_null_scalar.parquet');
filename
VARCHAR
[ Rows: 0]
```

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

Yes

- [x] Yes, I have

- [x] Yes, I have

allandefraga:
Fixes #1295

## Problem

`ducklake_add_data_files` registers a Parquet file without checking the table's NOT NULL
constraints, so a column declared `NOT NULL` starts returning NULL values after the import:

```sql
COPY (SELECT * FROM (VALUES (1, 10), (2, NULL)) v(id, x)) TO 'f.parquet';
CREATE TABLE ducklake.tbl(id INTEGER, x INTEGER NOT NULL);
CALL ducklake_add_data_files('ducklake', 'tbl', 'f.parquet');  -- succeeds today
SELECT * FROM ducklake.tbl;                                     -- x IS NULL

### DIFF
--- a/src/functions/ducklake_add_data_files.cpp
+++ b/src/functions/ducklake_add_data_files.cpp
@@ -147,6 +147,7 @@ private:
 	                                                    const vector<unique_ptr<DuckLakeFieldId>> &field_ids,
 	                                                    const string &prefix = string());
 	void MapColumnStats(ParquetFileMetadata &file_metadata, DuckLakeDataFile &result);
+	void CheckNotNullConstraints(const ParquetFileMetadata &file, const DuckLakeDataFile &data_file);
 	unique_ptr<DuckLakeNameMapEntry> MapHiveColumn(ParquetFileMetadata &file_metadata, const DuckLakeFieldId &field_id,
 	                                               const Value &hive_value);
 	void DetermineMapping(ParquetFileMetadata &file);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/functions/ducklake_add_data_files.cpp [hunk 1]
```diff
@@ -1086,6 +1087,50 @@ void DuckLakeFileProcessor::MapColumnStats(ParquetFileMetadata &file_metadata, D
...
+	}
+	for (auto &field_id : table.GetFieldData().GetFieldIds()) {
+		if (!not_null_fields.count(field_id->Name())) {
+			continue;
+		}
+		auto field_index = field_id->GetFieldIndex();
+		auto stats_entry = data_file.column_stats.find(field_index);
+		if (stats_entry != data_file.column_stats.end()) {
+			auto &stats = stats_entry->second;
+			if (stats.has_null_count && stats.null_count > 0) {
```
dentiny, line 1112: Parquet doesn't require `null_count` to be filled, so I don't think we have an easy/cheap and ensured way to check null constraint.

allandefraga: you are right, and as written it was best effort - no null_count meant the file went through unchecked

Pushed 025aca9
when a NOT NULL column has no null count in the footer, the values are counted instead,
for those columns only and in a single pass over the file.

```sql
SELECT COUNT(*) FILTER (WHERE "x" IS NULL) FROM read_parquet('file.parquet')
```

So the cheap path stays free - a file that carries statistics never gets read - and the check is now
ensured either way.

the read is only paid for files whose writer omitted the statistics, and only for the
columns that actually have the constraint.


Also,

On how often the read actually happens - every writer I measured fills null_count:

```
| writer | null_count |
| --- | --- |
| DuckLake `INSERT` | present (0 / 1 / 1 for the three columns) |
| DuckLake after `ducklake_merge_adjacent_files` | present on every file |
| DuckDB `COPY ... TO ... (FORMAT parquet)` | present |
| pyarrow, default arguments | present |
| pyarrow, `write_statistics=False` | absent - the only way I could produce such a file |
```

So files we write ourselves never take the read path, including after compaction. It only applies to
imported files whose writer chose to skip statistics.

PR #1354, opened 2026-08-01
Do not delete data files that are still referenced by another table

allandefraga:
## Problem

A Parquet file can be referenced by more than one `ducklake_data_file` entry — most easily by calling
`ducklake_add_data_files` with the same external file for two tables. Cleanup does not account for that:
`ducklake_expire_snapshots` collects the data file entries that are no longer reachable and schedules
**their paths** for physical deletion, without checking whether a surviving entry still points at the same
file. The file is then deleted from disk by `ducklake_cleanup_old_files` while it is still live for the
other table, which loses the data:

```sql
COPY (SELECT 42 AS id) TO 'shared.parquet';
CREATE TABLE lake.t1(id INTEGER);
CREATE TABLE lake.t2(id INTEGER);
CALL ducklake_add_data_files('lake', 't1', 'shared.parquet');
CALL ducklake_add_data_files('lake', 't2', 'shared.parquet');

DROP TABLE lake.t1;
CALL ducklake_expire_snapshots('lake', older_than => NOW());
CALL ducklake_cleanup_old_files('lake', cleanup_all => true);

SELECT id FROM lake.t2;  -- IO Error: No files found that match the pattern "shared.parquet"
```

## Fix

The metadata entries are still removed exactly as before — only the physical deletion is skipped for a
path that another entry, which is not itself being cleaned up, still points at. `GetStillReferencedFiles`
does that check in one query over the entries being cleaned up. Relative paths are resolved against the
table they belong to, so those only match within the same table; absolute paths match across tables.

The last reference behaves as it always did: once `t2` is dropped and expired, the file is scheduled and
deleted.

## Test

`test/sql/remove_orphans/cleanup_shared_data_file.test` covers both halves: the file survives while `t2`
still references it (and `t2` stays readable), and it is deleted once the last reference goes away. Red
before the change — the file is deleted and the read fails — green after.

The full suite passes on this branch: 16207 assertions in 457 test cases (16 skipped for missing
httpfs / postgres_scanner / sqlite_scanner / spatial).

## Note

Delete files are scheduled through a separate path and are not covered here; I have not found a way to
share one delete file between tables, but if you know of one I am happy to extend the same check to it.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -3088,7 +3088,7 @@ ORDER BY row_id, begin_snapshot;)",
 }
 
 string DuckLakeMetadataManager::ReadInlinedDataAggregatesSql(const string &inlined_table_name,
-                                                              const string &select_list) {
+                                                             const string &select_list) {
 	return StringUtil::Format(R"(
 SELECT %s
 FROM {METADATA_CATALOG}.%s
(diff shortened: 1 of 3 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -4879,12 +4522,49 @@ ORDER BY snapshot_id
...
+	// the same file can be registered more than once - e.g. ducklake_add_data_files called with the same external
+	// file for two tables. Relative paths are resolved against the table they belong to, so those only match within
+	// the same table.
+	auto query = StringUtil::Format(R"(
+SELECT c.data_file_id
+FROM {METADATA_CATALOG}.ducklake_data_file c
+WHERE c.data_file_id IN (%s) AND EXISTS(
+    SELECT 1
+    FROM {METADATA_CATALOG}.ducklake_data_file o
+    WHERE o.path = c.path AND o.path_is_relative = c.path_is_relative
```
dentiny: What if there's mixed usage for relative path and absolute path?

dentiny (commented):
I'm curious in which case and why do we have multiple tables, instead of multiple versions on one table, referencing the same file? I think this change is more like a feature request instead of a simple fix.

src/storage/ducklake_metadata_manager.cpp
allandefraga: you're right, comparing the stored path was wrong. Same file, two entries:

```
data_file_id | table_id | path | path_is_relative
4 | 1 | outside.parquet | false
5 | 2 | /tmp/dl4/outside.parquet | false
```

Fixed in **53ac438**: both sides are resolved with `FromRelativePath` and then canonicalized before
comparing, so the stored form doesnt matter. the query only pre-filters candidates by file name, to
avoid scanning the whole table.

allandefraga:
@dentiny 

about your comment

only `ducklake_add_data_files` can produce it, since it registers a file in place instead of writing one.
nothing stops adding the same file to two table, or even twice to the same table (that just doubles
the rows). I found it while importing an existing lake, where the same parquet ends up registered in a
second table during a migration.

PR #1355, opened 2026-08-01
feat: Allow reading arbitrary comment keys

borchero:
# Motivation

External tools might want to set arbitrary metadata on columns & tables. Currently, the ducklake extension only allows for the `"comment"` key. Afaict, this restriction is not strictly necessary even if no other key can currently be modified with this extension.

### DIFF
--- a/src/storage/ducklake_catalog.cpp
+++ b/src/storage/ducklake_catalog.cpp
@@ -548,12 +548,16 @@ unique_ptr<DuckLakeCatalogSet> DuckLakeCatalog::LoadSchemaForSnapshot(DuckLakeTr
 				not_null_columns.insert(col_info.name);
 			}
 			ColumnDefinition column(Identifier(std::move(col_info.name)), field_id->Type());
+			InsertionOrderPreservingMap<string> column_tags;
 			for (auto &tag : col_info.tags) {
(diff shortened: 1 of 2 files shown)


dentiny (commented):
Your code doesn't compile.
```sh
/Users/hjiang/Desktop/ducklake/src/storage/ducklake_catalog.cpp:555:13: error: 'tags' is a private member of 'duckdb::ColumnDefinition'
  555 |                                         column.tags[tag.key] = tag.value;
      |                                                ^
/Users/hjiang/Desktop/ducklake/duckdb/src/include/duckdb/parser/column_definition.hpp:113:38: note: declared private here
  113 |         InsertionOrderPreservingMap<string> tags;
      |                                             ^
```

borchero:
Sorry @dentiny, I opened this PR prematurely, before I had my local dev setup running 😅 should be fixed now 🙏🏼

pdet:
Hi @borchero thanks for the PR, can you check the failing CI?

I think that opening the tags is a good idea, but it does worry me a bit wrt ducklake reserved words, maybe we should change comments to `ducklake_comments`, and reserve anything with a `ducklake_` prefix in the spec. This will require some backwards compatibility and version checking.

PR #1357, opened 2026-08-03
Respect preserve_insertion_order when planning DuckLake writes

### LINKED ISSUE #1356: CREATE TABLE AS into a DuckLake table does not preserve input order
### What happens?

Writing into a DuckLake table with `CREATE TABLE AS` drops the input row order, even with `preserve_insertion_order` at its default of `true`. The same statement against a regular DuckDB table keeps the order, which is what made me look at the extension rather than core.

`SET threads = 1` before the write fixes it. Adding `row_number() OVER ()` to the projection also avoids it, though that changes the plan as well, so I wouldn't read much into that on its own. `read_parquet` and `read_json_auto` go wrong the same way. A `range()` source stays ordered in my testing and I don't know why. `INSERT INTO ... SELECT` is affected the same way (an earlier version of this issue said otherwise, that was me accidentally testing a patched build).

The cause looks like `DuckLakeInsert::PlanCopyForInsert` in `src/storage/ducklake_insert.cpp`, which sets `physical_copy.parallel = true` unconditionally and never looks at `preserve_insertion_order`. Core picks the copy mode through the copy function's `execution_mode` callback in `plan_copy_to_file.cpp`, and parquet's `ParquetWriteExecutionMode` only returns `PARALLEL_COPY_TO_FILE` when the planner has decided order doesn't need preserving.

I'd guess PhysicalCopyToFile is built directly because PhysicalBatchCopyToFile doesn't support per_thread_output, file_size_bytes or partition_output, which DuckLake sets. So an order preserving write through the existing operator has to fall back to a single threaded one.

Deriving `physical_copy.parallel` from that callback fixes CTAS in my testing and leaves the parallel path alone when `preserve_insertion_order = false`. I'll open a PR and link it here. The cost is real, so you may prefer a batch aware approach that keeps the parallelism.

This was easy to miss because the first row group usually came out intact in my runs. On a smaller input, or if you only check the first rows, it looks fine.

### To Reproduce

```sql
COPY (SELECT i AS id, 'row' || i AS name FROM range(8000000) t(i))
  TO 'f.csv' (FORMAT csv, HEADER true);

INSTALL ducklake;
LOAD ducklake;
ATTACH 'ducklake:meta.ducklake' AS lake (DATA_PATH 'lakedata/');
USE lake;

SET preserve_insertion_order = true;
CREATE TABLE t AS SELECT * FROM read_csv('f.csv');

-- read back single threaded, so this reports how the rows were stored
SET threads = 1;
SELECT count(*) AS rows_out_of_input_order
FROM (SELECT id, row_number() OVER () - 1 AS pos FROM t)
WHERE id <> pos;
```
```
┌─────────────────────────┐
│ rows_out_of_input_order │
│          int64          │
├─────────────────────────┤
│                 8000000 │
└─────────────────────────┘
```

The same CSV and the same setting into a plain duckdb database file, in a separate session so the ducklake catalog is not attached (`duckdb native.db`):

```sql
SET preserve_insertion_order = true;
CREATE TABLE t AS SELECT * FROM read_csv('f.csv');
SELECT count(*) AS rows_out_of_input_order FROM t WHERE rowid <> id;
```

```
0
```

Variants, all against the DuckLake table with `preserve_insertion_order = true` and 8000000 rows:

```
read_csv, default threads (10)            8000000 out of order
read_csv, threads = 1                     0
read_parquet, default threads             6389760 out of order
read_parquet, threads = 1                 0
read_json_auto, default threads           7991122 out of order
range() source, default threads           0
read_csv with row_number() OVER ()        0
INSERT INTO ... SELECT read_csv           8000000 out of order
```

Also reproduced on a source build of ducklake main (7c0eafc8) against duckdb c97bd8b9, where the same script reports 7877120 out of order.

### OS:

macOS 26.2 (arm64)

### DuckDB Version:

v1.5.4 (08e34c447b), also v1.5.2 (8a5851971f); plus a source build of duckdb c97bd8b9

### DuckLake Version:

d318a545 and 415a9ebd; also main at 7c0eafc8

### DuckDB Client:

CLI

### Hardware:

Apple M4, 10 cores, 16 GB. Relevant here because threads defaults to 10 and thread count changes the outcome.

### Full Name:

Ayush Kaithwas

### Affiliation:

Polyteia GmbH

- [x] Yes, I have

- [x] Yes, I have

AyushKaithwas:
Fixes #1356.

PlanCopyForInsert hardcoded PhysicalCopyToFile::parallel = true, so DuckLake writes always took the fully parallel copy path. That path interleaves row groups from multiple threads and discards the input order, which is exactly the behaviour parquet's ParquetWriteExecutionMode reserves for preserve_insertion_order = false. As a result CREATE TABLE AS lost the input order even with the setting at its default of true, silently, and only once the reader went parallel.

Derive the flag from the copy function's execution_mode callback instead, the same way core does in plan_copy_to_file.cpp. supports_batch_index is passed as false because the options set here (per_thread_output, file_size_bytes, partition_output) are not supported by PhysicalBatchCopyToFile, so an order-preserving write falls back to REGULAR_COPY_TO_FILE and runs single threaded. Keeping order and write parallelism together would need batch index support in the partitioning and rotating writer, which is a larger change.

Repro, 8,000,000 row CSV with id ascending in file order, default threads:

```sql
SET preserve_insertion_order = true;
CREATE TABLE t AS SELECT * FROM read_csv('f.csv');
```

Reading back single threaded, 7,877,120 of 8,000,000 rows were not stored in input order before this change, and 0 after. A native DuckDB table stores 0 out of order with the same statement, which is the asymmetry that pointed here. With preserve_insertion_order = false the parallel path is unchanged.

Tested on a source build of main (7c0eafc8) against the pinned duckdb (c97bd8b9), macOS 26.2 arm64, 10 cores.

I have not added a test. The disorder only appears above roughly 500,000 rows, so a fixture large enough to catch it means generating a multi-megabyte CSV, which seemed too slow for the suite. Happy to add one if you want it, or to put it somewhere other than sqllogictest.

### DIFF
--- a/src/storage/ducklake_insert.cpp
+++ b/src/storage/ducklake_insert.cpp
@@ -718,7 +718,14 @@ PhysicalOperator &DuckLakeInsert::PlanCopyForInsert(ClientContext &context, Phys
 	physical_copy.partition_columns = std::move(copy_options.partition_columns);
 	physical_copy.names = copy_options.names;
 	physical_copy.expected_types = std::move(copy_options.expected_types);
-	physical_copy.parallel = true;
+	// a fully parallel copy interleaves row groups and loses the input order
+	const bool preserve_insertion_order = plan ? PhysicalPlanGenerator::PreserveInsertionOrder(context, *plan) : false;
+	// no batch index: PhysicalBatchCopyToFile does not support the options set above
+	auto execution_mode = CopyFunctionExecutionMode::PARALLEL_COPY_TO_FILE;
+	if (physical_copy.function.execution_mode) {
+		execution_mode = physical_copy.function.execution_mode(preserve_insertion_order, false);
+	}
+	physical_copy.parallel = execution_mode == CopyFunctionExecutionMode::PARALLEL_COPY_TO_FILE;
 	physical_copy.hive_file_pattern =
 	    copy_input.catalog.UseHiveFilePattern(!is_encrypted, copy_input.schema_id, copy_input.table_id);
 	if (plan) {


Alex-Monahan:
I don't think that we want to remove parallelism for all sorted inserts. That will impact performance substantially.

A few notes on this behavior: I believe it only happens when the insert is larger than the target file size. Each individual row group is sorted within itself, but the row groups are scattered across files, so file-level min/max statistics in the DuckLake catalog overlap heavily. Even within individual files, the order is not maintained. However, the parquet footer statistics are still just as effective. So the only consequence is that more footers are read than are necessary.

AyushKaithwas:
Agreed, with the default being true this serializes basically every DuckLake write, which is too expensive. The PR was mostly to show where the order gets lost, happy to close it or rework it.

The consequence for us isn't stats overlap or extra footer reads though. This came from CSV ingestion, where the file's row order is the data and there's no column to ORDER BY afterwards to recover it. The CSV reader docs say result order matches the file when the setting is on, and below roughly 500k rows the write does keep it, so small tests pass and bigger ingests scramble with no signal.

On the target file size: on an unpatched build of main, an 8M row insert wrote a single 68MB parquet file (target_file_size at the 512MiB default) and all 8,000,000 rows were out of input order, so it happens well below the target size. The interleaving is at row group granularity inside the one file, contiguous runs of 122,880 rows.

Also, the issue says INSERT INTO ... SELECT came out ordered. That's wrong, I had accidentally tested a patched binary. On a clean build INSERT is affected the same way (8,000,000 and 7,877,120 out of order in two runs). I've corrected the issue.

Would batch index support in the writer be a reasonable direction? That keeps both the parallelism and the order. I'm willing to work on it, but I'd need a pointer on how batch indexes should fit with the partitioning and file rotation. And if the current behavior is intended, a sentence in the docs saying DuckLake writes don't preserve insertion order would prevent the surprise, since nothing signals it today.

AyushKaithwas:
Hey, just wanted to check up on this.

Alex-Monahan:
As a workaround, could you first insert from your CSV file into a temp table and apply a row number identifier at that time? Then you could insert into DuckLake from that temp table. For example:

```sql
create or replace temp table csv_with_order as
    from read_csv()
    select *, row_number() over () as row_id
;
insert into ducklake_table
    select * from csv_with_order
```

PR #1359, opened 2026-08-04
Pin schema cache entries for the transaction lifetime

### LINKED ISSUE #1248: Concurrent transaction commits fail with "Failed to commit DuckLake transaction"
### What happens?

Since the DuckLake version bundled with DuckDB **1.5.3**, running **concurrent write transactions** that each `CREATE OR REPLACE TABLE ... AS SELECT` then `COMMIT` against a DuckLake catalog reliably fails. The identical workload is **100% clean on the DuckLake bundled with DuckDB 1.5.2** and remains broken on **1.5.4**.

The failure is a commit-time error:

```
TransactionException: Failed to commit: Failed to commit DuckLake transaction.
Failed to get path for schema with id <garbage-id> - schema not found in metadata catalog
```

where `<garbage-id>` is an obviously bogus value that **changes from run to run** (e.g. observed ids  were `281471621436160`, `8313474949019626339` -- which btw is little-endian ascii for `cklake_s`), the signature of a dereferenced/freed pointer. Under heavier concurrency or memory pressure the same defect also manifests as a hard process crash: `SIGSEGV` on Linux aarch64, or a managed `Allocation failure` on Linux x86_64 (jemalloc).

It is a **race**, so the number of concurrent transactions needed to trigger it is probabilistic, not fixed: as few as **2** concurrent transactions can trigger it, and it becomes 100% reliable at **16+** (8 worker threads, `memory_limit=2GB`). It is not memory-bound; it reproduces with trivial `range(1000)` tables well under the memory limit.

The regression was introduced somewhere in the DuckLake commit range bundled between DuckDB 1.5.2 (`415a9ebd`, good) and 1.5.3 (`e6a3bd0a`, bad). The likely culprit is **PR #1020 "Fix unbounded memory consumption on schema cache entries"** (commit `a8ea57577c`), which made the per-catalog schema cache evictable, consistent with the dangling-pointer schema id in the error.

### To Reproduce

**Minimal reproducer (self-contained, no external data, synthetic `range()` only).** A single shared in-memory connection, one short-lived cursor per task, one transaction per task, all committing `CREATE OR REPLACE TABLE` into the same DuckLake schema. No source catalog or input data is needed.

```python
import duckdb, concurrent.futures, tempfile, os

con = duckdb.connect(":memory:", config={"threads": "8", "memory_limit": "2GB"})
d = tempfile.mkdtemp()
con.execute(f"ATTACH 'ducklake:{os.path.join(d, 'catalog.ducklake')}' AS lake (DATA_PATH '{d}')")

def work(i):
    cur = con.cursor()
    cur.execute("BEGIN")
    cur.execute(f"CREATE OR REPLACE TABLE lake.main.t{i} AS SELECT r AS id, r % 100 AS grp FROM range(1000) t(r)")
    cur.execute("COMMIT")
    cur.close()

with concurrent.futures.ThreadPoolExecutor(max_workers=8) as ex:
    list(ex.map(work, range(40)))
print("ALL_DONE")
```

Output on DuckDB **1.5.3** (DuckLake `e6a3bd0a`):

```
_duckdb.TransactionException: TransactionContext Error: Failed to commit: Failed to commit DuckLake transaction.
Failed to get path for schema with id 281470346892432 - schema not found in metadata catalog
```

Output on DuckDB **1.5.2** (DuckLake `415a9ebd`):

```
ALL_DONE
```

**Failure rate vs. number of concurrent transactions.** Same script with `range(40)` replaced by `range(N)`. Each cell is failures out of 5 runs (a "failure" is an exception or crash before `ALL_DONE`). 8 worker threads, `memory_limit=2GB`:

| N (transactions) | DuckDB 1.5.2 | DuckDB 1.5.3 |
|---|---|---|
| 2 | 0/5 | 1/5 |
| 3 | 0/5 | 1/5 |
| 4 | 0/5 | 2/5 |
| 5 | 0/5 | 3/5 |
| 8 | 0/5 | 4/5 |
| 12 | 0/5 | 4/5 |
| 16 | 0/5 | 5/5 |
| 24 | 0/5 | 5/5 |
| 40 | 0/5 | 5/5 |

### OS:

Linux aarch64 (`python:3.12-slim` container); also reproduced on Linux x86_64

### DuckDB Version:

1.5.3 and 1.5.4 (bad); 1.5.2 (good)

### DuckLake Version:

DuckLake ships no version tags/releases, so the bundled commit SHA is the identifier. Bad = `e6a3bd0a8554b74d97cbc7e8acc3e2c9f01a0385` (bundled in DuckDB 1.5.3); good = `415a9ebdbd73db50a8c6ba703eb733ed16bcf33a` (bundled in 1.5.2). Likely introduced by PR #1020 commit `a8ea57577c`.

### DuckDB Client:

Python

### Hardware:

Reproduces at `memory_limit=2GB` with trivial `range(1000)` tables; not memory-bound.

### Full Name:

Marcel Steinbach

### Affiliation:

Ometria

- [X] Yes, I have

- [X] Yes, I have

utay:
Addresses #1248.

DuckLake catalog entries keep bare references to their parent schemas and may read them again at commit time. Because those schemas live in evictable cache entries, the entries must stay alive for the full transaction.

`main` already pins schema cache entries on `DuckLakeTransaction`. This PR removes the now-redundant query-scoped pin.

### DIFF
--- a/src/include/storage/ducklake_catalog.hpp
+++ b/src/include/storage/ducklake_catalog.hpp
@@ -78,10 +78,9 @@ struct DuckLakeSchemaCacheEntry : public ObjectCacheEntry {
 	optional_idx GetEstimatedCacheMemory() const override;
 };
 
-//! Query-scoped pin for DuckLake schema cache entries, which guarantee memory safety before transaction finishes.
-class DuckLakeSchemaPinState : public ClientContextState {
+//! Holds pins on DuckLake schema cache entries, keeping them alive while they are still referenced.
+class DuckLakeSchemaPinState {
 public:
-	void QueryEnd(ClientContext &context) override;
 	void Pin(shared_ptr<DuckLakeSchemaCacheEntry> entry);
 	//! Clear all pinned schema cache entries for this pin state.
 	void Clear();
@@ -312,12 +311,9 @@ private:
 	//! Look up (or load) the ObjectCache entry for a given snapshot.
 	shared_ptr<DuckLakeSchemaCacheEntry> GetSchemaCacheEntry(DuckLakeTransaction &transaction,
 	                                                         DuckLakeSnapshot snapshot);
-	//! Pin a schema cache entry for the duration of the current query to ensure safe memory access.
-	void PinSchemaForQuery(DuckLakeTransaction &transaction, shared_ptr<DuckLakeSchemaCacheEntry> entry);
 	void LoadNameMaps(DuckLakeTransaction &transaction);
 	string StatsCacheKey(idx_t next_file_id, TableIndex table_id) const;
 	string SchemaCacheKey(idx_t schema_version) const;
-	string SchemaPinStateKey() const;
 	ObjectCache &GetObjectCacheInstance();
 
 private:
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -203,10 +203,14 @@ public:
 	unique_ptr<QueryResult> ExecuteRaw(string query);
 	Connection &GetConnection();
 
+	//! Keep a schema cache entry alive for as long as this transaction lives. Transaction-local catalog entries hold
+	//! bare references into the cached catalog set, and those references are read again at commit time, so the entry
+	//! must not be evicted from the ObjectCache in between.
+	void PinSchemaCacheEntry(shared_ptr<DuckLakeSchemaCacheEntry> entry);
+
 	DuckLakeSnapshot GetSnapshot();
 	DuckLakeSnapshot GetSnapshot(optional_ptr<BoundAtClause> at_clause,
 	                             SnapshotBound bound = SnapshotBound::UPPER_BOUND);
-	void PinSchemaCacheEntry(shared_ptr<DuckLakeSchemaCacheEntry> entry);
 
 	static DuckLakeTransaction &Get(ClientContext &context, Catalog &catalog);
 
--- a/src/storage/ducklake_catalog.cpp
+++ b/src/storage/ducklake_catalog.cpp
@@ -166,10 +166,6 @@ optional_idx DuckLakeSchemaCacheEntry::GetEstimatedCacheMemory() const {
 	return EstimateCatalogSetMemory(catalog_set);
 }
 
-void DuckLakeSchemaPinState::QueryEnd(ClientContext &context) {
-	Clear();
-}
-
 void DuckLakeSchemaPinState::Clear() {
 	lock_guard<mutex> guard(lock);
 	pins.clear();
@@ -369,22 +365,9 @@ shared_ptr<DuckLakeSchemaCacheEntry> DuckLakeCatalog::GetSchemaCacheEntry(DuckLa
 
 DuckLakeCatalogSet &DuckLakeCatalog::GetSchemaForSnapshot(DuckLakeTransaction &transaction, DuckLakeSnapshot snapshot) {
 	auto entry = GetSchemaCacheEntry(transaction, snapshot);
-	transaction.PinSchemaCacheEntry(entry);
-	PinSchemaForQuery(transaction, entry);
-	return entry->catalog_set;
-}
-
-void DuckLakeCatalog::PinSchemaForQuery(DuckLakeTransaction &transaction, shared_ptr<DuckLakeSchemaCacheEntry> entry) {
-	if (!entry) {
-		return;
-	}
-	auto context_ref = transaction.context.lock();
-	if (!context_ref) {
-		return;
-	}
-	auto &registered = *context_ref->registered_state;
-	auto pin_state = registered.GetOrCreate<DuckLakeSchemaPinState>(SchemaPinStateKey());
-	pin_state->Pin(std::move(entry));
+	auto &catalog_set = entry->catalog_set;
+	transaction.PinSchemaCacheEntry(std::move(entry));
+	return catalog_set;
 }
 
 static unique_ptr<DuckLakeFieldId> TransformColumnType(DuckLakeColumnInfo &col) {
@@ -1136,10 +1119,6 @@ void DuckLakeCatalog::InvalidateSchemaCache(idx_t schema_version) {
 void DuckLakeCatalog::InvalidateNameMapCache(MappingIndex mapping_id) {
 	lock_guard<mutex> guard(name_maps_lock);
 	name_maps.Remove(mapping_id);
-}
-
-string DuckLakeCatalog::SchemaPinStateKey() const {
-	return StringUtil::Format("ducklake_schema_pin:%s:%s:%s", GetName(), MetadataPath(), instance_id);
 }
 
 ObjectCache &DuckLakeCatalog::GetObjectCacheInstance() {
--- a/test/sql/issues/issue_1248_concurrent_commit_schema_id.test
+++ b/test/sql/issues/issue_1248_concurrent_commit_schema_id.test
@@ -1,0 +1,48 @@
+# name: test/sql/issues/issue_1248_concurrent_commit_schema_id.test
+# description: https://github.com/duckdb/ducklake/issues/1248 - concurrent transactions must commit their new tables into the right schema
+# group: [issues]
+
+require notwindows
+
+require ducklake
+
+require parquet
+
+test-env DUCKLAKE_CONNECTION __TEST_DIR__/{UUID}.db
+
+test-env DATA_PATH __TEST_DIR__
+
+statement ok
+ATTACH 'ducklake:${DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '${DATA_PATH}/issue_1248')
+
+statement ok
+SET memory_limit='2GB'
+
+# Committing used to pick up a bogus schema id when another transaction committed in the meantime.
+concurrentloop threadid 0 8
+
+loop i 0 5
+
+statement ok
+BEGIN
+
+statement ok
+CREATE OR REPLACE TABLE ducklake.t_${threadid}_${i} AS SELECT r AS id, r % 100 AS grp FROM range(1000) t(r)
+
+statement ok
+COMMIT
+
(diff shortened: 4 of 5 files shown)


utay:
Hi @dentiny! Invoking you on this one, as you wrote the initial optimization and are probably familiar with this code :)

dentiny (approved):
Thanks! But I think the sql test comment should be cleaned up a bit

utay:
> Thanks! But I think the sql test comment should be cleaned up a bit

Done!

pdet:
Hi @utay, thanks for the PR.

Could you retarget it to `main`?

utay:
> Hi @utay, thanks for the PR.
> 
> Could you retarget it to `main`?

Hi @pdet, done, but tests are failing now; I will have a look

utay:
@pdet @dentiny tests should be fixed now; PTAL

pdet:
Thanks!

ukclivecox:
A ducklake invalidation on postgres is happening for us and a contributor added our test: https://github.com/duckdb/ducklake/pull/1457 suggesting its ok in 2.0 dev now

However running locally across 20 attempts of up to 30 seconds, the original empty-vector/database-invalidation error did indeed not recur but 3 attempts segfaulted, all while executing CREATE OR REPLACE TABLE, with the same stack:

  SchemaCatalogEntry::GetEntry
  DuckLakeSchemaEntry::HandleCreateConflict
  DuckLakeSchemaEntry::CreateTableExtended

  The other 17 attempts passed their final health query but still logged transaction/catalog errors. This resembles #1248, although the test we ran was built with #1359 and #1255.

Just posting here in case of use to validate if this PR does fix things.

staticlibs:
@ukclivecox 

Hi, thanks for additional details! Can you clarify, which exact versions of DuckDB, DuckLake and Postgres scanner you used to reproduce the segfault?

ukclivecox:
- DuckDB: v2.0.0-alpha42526, commit ae9f515999
  - DuckLake extension: e8924cfc9c
  - Postgres scanner extension: bc6aab54de
  - PostgreSQL server: 16.15

  DuckDB was the official prebuilt Linux x86_64 shared library. Both extensions
  were downloaded from
  https://extensions.duckdb.org/v2.0.0-alpha42526/linux_amd64/.

  The unchanged C++ ADBC reproducer segfaulted in 3 of 20 attempts, each
  configured for 30 seconds.

staticlibs:
@ukclivecox 

Thanks for these details! 

> Postgres scanner extension: bc6aab54de

This is the only difference with my env (where it is not reproducible) I can see, and this is somewhat a suspect, as this version of Postgres scanner does not include duckdb/duckdb-postgres#568 change. Will try it without that fix.

staticlibs:
@ukclivecox

Confirming that I indeed can successfully reproduce the problem on a build without duckdb/duckdb-postgres#568:

```
ERROR: FATAL Error: Failed to query most recent snapshot for DuckLake: Failed: database has been invalidated because of a previous fatal error. The database must be restarted prior to being used again.
Original error: "Attempted to access index 0 within vector of size 0"
Statement execution failed
SQL: UPDATE table_0 SET revision=5 WHERE item_id='item-1'
ERROR: INTERNAL Error: Attempted to access index 0 within vector of size 0

Stack Trace:

libduckdb.so(duckdb::Exception::ToJSON(std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > const&, duckdb::ExceptionType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x2dc) [0x7f037186d6ec]
libduckdb.so(duckdb::Exception::ToJSON(duckdb::ExceptionType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x58) [0x7f037186e798]
libduckdb.so(duckdb::Exception::Exception(duckdb::ExceptionType, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x11) [0x7f037186e821]
libduckdb.so(duckdb::InternalException::InternalException(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)+0x11) [0x7f0371872e91]
libduckdb.so(duckdb::InternalException::InternalException<unsigned long&, unsigned long&>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long&, unsigned long&)+0x167) [0x7f0370c66fb7]
ducklake.duckdb_extension(+0x16c04) [0x7b0362216c04]
ducklake.duckdb_extension(_ZN6duckdb25DuckLakeInlinedDataReader17TryInitializeScanERNS_13ClientContextERNS_24GlobalTableFunctionStateERNS_23LocalTableFunctionStateE+0x10c9) [0x7b03625a03c9]
libduckdb.so(duckdb::MultiFileFunction<duckdb::ParquetMultiFileInfo>::TryInitializeNextBatch(duckdb::ClientContext&, duckdb::MultiFileBindData const&, duckdb::MultiFileLocalState&, duckdb::MultiFileGlobalState&)+0xa10) [0x7f0372986c20]
libduckdb.so(duckdb::MultiFileFunction<duckdb::ParquetMultiFileInfo>::MultiFileInitLocal(duckdb::ExecutionContext&, duckdb::TableFunctionInitInput&, duckdb::GlobalTableFunctionState*)+0x1ea) [0x7f037298789a]
ducklake.duckdb_extension(+0xed5764) [0x7b03630d5764]
libduckdb.so(duckdb::PipelineExecutor::PipelineExecutor(duckdb::ClientContext&, duckdb::Pipeline&)+0x390) [0x7f0371d74270]
libduckdb.so(duckdb::PipelineTask::ExecuteTask(duckdb::TaskExecutionMode)+0x1ba) [0x7f0371d7a2ca]
libduckdb.so(duckdb::ExecutorTask::Execute(duckdb::TaskExecutionMode)+0xd3) [0x7f0371d70f63]
libduckdb.so(duckdb::Executor::ExecuteTask(bool)+0x75) [0x7f0371d7bd95]
libduckdb.so(duckdb::ClientContext::ExecuteTaskInternal(duckdb::ClientContextLock&, duckdb::BaseQueryResult&, bool)+0x50) [0x7f0371d5a060]
libduckdb.so(duckdb::PendingQueryResult::ExecuteInternal(duckdb::ClientContextLock&)+0x3c) [0x7f0371d5a27c]
libduckdb.so(duckdb::ClientContext::Execute(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, duckdb::shared_ptr<duckdb::PreparedStatementData, true>&, duckdb::PendingQueryParameters const&)+0x2d6) [0x7f0371d5c396]
libduckdb.so(duckdb::PreparedStatement::Execute(std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, duckdb::BoundParameterData, duckdb::CaseInsensitiveStringHashFunction, duckdb::CaseInsensitiveStringEquality, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, duckdb::BoundParameterData> > >&, bool)+0x29c) [0x7f0371d6277c]
libduckdb.so(duckdb_execute_prepared_streaming+0x67) [0x7f0371c2a707]
libduckdb.so(duckdb_adbc::StatementExecuteQuery(AdbcStatement*, ArrowArrayStream*, long*, AdbcError*)+0x560) [0x7f037141db80]
./repro() [0x4024fa]
./repro() [0x402bcf]
/lib64/libstdc++.so.6(+0x4e1e4) [0x7f037044e1e4]
/lib64/libc.so.6(+0x71ac4) [0x7f037027fac4]
/lib64/libc.so.6(+0xf44ac) [0x7f03703024ac]
```

I wonder if you can try the `318dabb` version of Postgres scanner from core_nightly repo with DuckDB 1.5.5 in your env:

```sql
FORCE INSTALL postgres FROM core_nightly;
```

I will look into updating the Postgres scanner version in 2.0-alpha central builds tomorrow.

ukclivecox:
Tested the requested `postgres_scanner` version `318dabb`, installed using `FORCE INSTALL postgres FROM core_nightly`, with DuckDB **1.5.5** (`d8cdaa33fd`), DuckLake **`d8a1881e`**, and PostgreSQL **16.15** on Linux x86_64. I verified the scanner revision and loaded that exact extension file in the unchanged standalone C++ ADBC reproducer.

The original failure still reproduced on **attempt 2** of a configured 20 × 30-second run:

```text
Attempted to access index 0 within vector of size 0
database has been invalidated because of a previous fatal error
```

The final health query also failed. Attempt 1 completed without invalidation but logged metadata and transaction errors. No segfault or timeout occurred in these two attempts; the harness stopped after reproducing the invalidation.

Only the scanner binary changed from our original 1.5.5 toolchain; the DuckDB library, DuckLake extension and reproducer were identical. 
[ducklake-1.5.5-postgres-318dabb-failure.log](https://github.com/user-attachments/files/32522585/ducklake-1.5.5-postgres-318dabb-failure.log)

staticlibs:
@ukclivecox 

Thanks for checking it! This confirms that duckdb/duckdb-postgres#568 is unrelated.

Just FYI, I filed #1469 that should fix the problem in 1.5. In environment where I can reproduce the problem reliably (faster machine with more CPU cores) in 1-2 minutes, with #1469 applied I was unable to reproduce it in a 20 min run.

The fix is specific to `v1.5-variegata` branch, I will try it more on the `main` branch and will follow up.

Issue #1311, opened 2026-07-14
PostgreSQL snapshot conflict retry depends on English lc_messages


kkrakow1:
## Description

DuckLake does not retry concurrent snapshot commits when the PostgreSQL metadata catalog returns localized error messages.

With PostgreSQL `lc_messages` set to, for example, `Polish_Poland.1250`, concurrent writers can fail with:

```text
TransactionContext Error: Failed to commit: Failed to commit DuckLake transaction.
Failed to flush changes into DuckLake: Failed to execute query
"INSERT INTO "public".ducklake_snapshot VALUES (...); ...":

BŁĄD: podwójna wartość klucza narusza ograniczenie unikalności
"ducklake_snapshot_pkey"
DETAIL: Klucz (snapshot_id)=(...) już istnieje.
```

Increasing `ducklake_max_retry_count` does not help.

Notably, the error does not contain:

```text
Exceeded the maximum retry count
```

This suggests that the error bypasses the retry loop rather than exhausting the configured number of retries.

After changing PostgreSQL `lc_messages` from Polish to English and reconnecting all writers, the same concurrency test succeeds and DuckLake retries the conflicting snapshot commits correctly.

## Suspected root cause

`DuckLakeTransaction::RetryOnError` appears to classify retryable errors by searching the human-readable error message for English phrases:

```cpp
bool DuckLakeTransaction::RetryOnError(const string &original_message) {
    auto message = StringUtil::Lower(original_message);

    if (StringUtil::Contains(message, "primary key") ||
        StringUtil::Contains(message, "unique")) {
        return true;
    }
    if (StringUtil::Contains(message, "conflict")) {
        return true;
    }
    if (StringUtil::Contains(message, "concurrent")) {
        return true;
    }
    return false;
}
```

A localized PostgreSQL duplicate-key error may contain none of these English phrases. As a result, a collision on `ducklake_snapshot_pkey` is not classified as retryable and bypasses DuckLake's retry path.

The constraint name `ducklake_snapshot_pkey` remains present in both the localized and English error messages.

Ideally, retry classification should use structured database error information, such as PostgreSQL SQLSTATE `23505` for a unique-constraint violation. If structured error information is not available at this layer, checking the stable DuckLake constraint name would still avoid depending on the PostgreSQL message language.

## Expected behavior

Retry handling should not depend on the human-readable language of PostgreSQL error messages.

[...]

Issue #1283, opened 2026-07-07
Concurrent CREATE OR REPLACE TABLE fails with "Exceeded the maximum retry count" — duplicate key on ducklake_snapshot_pkey (Postgres catalog, 1.5.4)


suresh-summation:
### What happens?

With a **Postgres catalog**, concurrent `CREATE OR REPLACE TABLE … AS SELECT …` statements issued from multiple connections consistently fail to commit with:

```
TransactionContext Error: Failed to commit: Failed to commit DuckLake transaction.
Exceeded the maximum retry count of 10 set by the ducklake_max_retry_count setting.
. Consider increasing the value with: e.g., "SET ducklake_max_retry_count = 100;"
Failed to flush changes into DuckLake: Failed to execute query
"INSERT INTO "public".ducklake_snapshot VALUES (232, NOW(), 232, 233, 232);
 ... UPDATE "public".ducklake_table SET end_snapshot = 232 ... ;
 INSERT INTO "public".ducklake_table VALUES (232, '019f…', 232, NULL, 0, '_tmp_w4', '_tmp_w4/', true);
 INSERT INTO "public".ducklake_column ... ;
 INSERT INTO "public".ducklake_data_file ... ;
 INSERT INTO "public".ducklake_snapshot_changes VALUES (232, 'dropped_table:52,created_table:"main"."_tmp_w4",inserted_into_table:232', ...);":
ERROR:  duplicate key value violates unique constraint "ducklake_snapshot_pkey"
DETAIL:  Key (snapshot_id)=(232) already exists.
```

Notes / observations:

- The collision is on the **`ducklake_snapshot` primary key** (`snapshot_id`) — an auto-incremented catalog id. Each concurrent committer independently picks the *same* next `snapshot_id`, so one wins and the rest hit the PK violation, retry, and after `ducklake_max_retry_count` (default **10**) attempts give up.
- The failing statements create **different tables** (`_tmp_w1`, `_tmp_w2`, …), so there is **no logical conflict** between the transactions — it is purely a race on the shared snapshot-id sequence.
- **Raising `ducklake_max_retry_count` makes it disappear.** At the load below, `=10` fails reliably; `=30` gives **0 failures**, and `=30` also survives a much heavier 50-writer / 2000-commit run with 0 failures. So this is retry-budget exhaustion under contention, not a hard deadlock.
[...]

PR #1361, opened 2026-08-04
fix: make transient & typed conflicts retryable

hello-world-bfree:
Part of #1094 resolution

## Problem

Currently some retries rely on string-matching, even the ducklake `TransactionException` and some transient background failures like deadlocks, dropped connections, etc., are not considered retryable

## Fix

Added a typed `RetryOnError` overload that classifies a transaction by type. It also includes the transient pg exceptions. current callers remain unchanged. also added `ducklake_would_retry()` for testability.

### DIFF
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -16,6 +16,7 @@
 namespace duckdb {
 
 ScalarFunction DuckLakeMurmur3Function();
+ScalarFunctionSet DuckLakeWouldRetryFunction();
 
 static void LoadInternal(ExtensionLoader &loader) {
 	loader.SetDescription("Adds support for DuckLake, SQL as a Lakehouse Format");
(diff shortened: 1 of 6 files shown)


hello-world-bfree:
Two open issues are user-facing reports of what this PR addresses:

- #1311  
  - Concurrent writers on a Polish-locale server fail with a duplicate-key error that never re-enters the retry loop, and `ducklake_max_retry_count` does not help. 
  - `RetryOnError` matching English substrings is the cause.
- #1283 

#1311 isn't fully fixed by this PR though. I added `"could not serialize"`, `"deadlock"`, and the connection-drop strings but they're all still English. `ExceptionType::TRANSACTION` only covers DuckLake's own conflicts; backend errors come through untyped.

The robust fix is to classify on SQLSTATE (`40001` serialization failure, `40P01` deadlock, `23505` unique violation) rather than me ssage text. In the libpq docs, `PG_DIAG_SQLSTATE` "is not localizable, and is always present."

In duckdb-postgres every error path captures `PQresultErrorMessage(result)` only. That throws `std::runtime_error` with no type or code to key off of.

Confirmed locally postgres through the path ducklake commits take 

```
Invalid Error: Failed to execute query "BEGIN TRANSACTION ISOLATION LEVEL REPEATABLE READ;
INSERT INTO public.sqlstate_probe VALUES (1)": ERROR:  duplicate key value violates unique constraint "sqlstate_probe_pkey"
DETAIL:  Key (id)=(1) already exists.
```

PR #1363, opened 2026-08-06
Add newer_than option to ducklake_merge_adjacent_files

zhming0:
Implements the proposal in https://github.com/duckdb/ducklake/discussions/1362.

## What

Adds a `newer_than => TIMESTAMPTZ` named parameter to `ducklake_merge_adjacent_files`:

```sql
CALL ducklake_merge_adjacent_files('my_catalog', newer_than => now() - INTERVAL 1 DAY);
```

Only files written at or after the given timestamp are considered compaction candidates. The name mirrors the existing `older_than` option of `ducklake_expire_snapshots` / `ducklake_cleanup_old_files`.

## Why

For append-heavy time-series tables compacted on a schedule, every run currently evaluates metadata for all live files, so cost grows with total table history even though only recently written files can be new merge candidates. With `newer_than`, a cron job only pays for what changed since the last run.

## How

Since data files are immutable, the filter is on file creation time. The timestamp is mapped to the first snapshot with `snapshot_time >= newer_than`, and a `begin_snapshot >= <snapshot_id>` predicate is pushed into the catalog query in `GetFilesForCompaction` (alongside the existing file-size predicates), so historic files are never read from the metadata catalog. The boundary is inclusive, so scheduled jobs never silently skip boundary files. The timestamp handling reuses the same pattern as `ducklake_expire_snapshots(older_than => ...)`, so it works across catalog backends.

### DIFF
--- a/src/functions/ducklake_compaction_functions.cpp
+++ b/src/functions/ducklake_compaction_functions.cpp
@@ -292,6 +292,7 @@ void DuckLakeCompactor::GenerateCompactions(DuckLakeTableEntry &table,
 	filter_options.min_file_size = options.min_file_size;
 	filter_options.max_file_size = options.max_file_size;
 	filter_options.target_file_size = target_file_size;
+	filter_options.newer_than = options.newer_than;
 	// FIXME: pass in the sort_data so that list of files is approximately sorted in the same way
 	// (sorted by the min/max metadata)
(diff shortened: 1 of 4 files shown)


dentiny (approved):

pdet:
Thanks!

PR #1369, opened 2026-08-08
Fix snapshot_id filters on deletion scans

oleg-milanov:
Filters pushed down on the virtual `snapshot_id` column of `ducklake_table_deletions/ducklake_table_changes` were applied to raw delete-file columns before per-row snapshot ids are gathered in `FinalizeChunk` - equality filters were silently dropped, and range filters saw uninitialized values.

This defers the filter at scan init (`GetFiltersToUse`) and applies it after gather, including when `snapshot_id` is filtered but not projected.

I am adding a regression test covering projected/unprojected equality and range filters, multi-file scans, and the ducklake_table_changes path.


pdet:
Hi @oleg-milanov thanks for the PR! Could you add the following as a test?

```sql
 ATTACH 'ducklake:/tmp/gap/gap.db' AS dl (DATA_PATH '/tmp/gap/data', DATA_INLINING_ROW_LIMIT 10);
  USE dl;
  CREATE TABLE d (k INTEGER, v INTEGER);                 -- snapshot 1
  INSERT INTO d SELECT i, i*10 FROM range(1, 101) t(i);  -- snapshot 2: one parquet file
  DELETE FROM d WHERE k = 2;                             -- snapshot 3: inlined delete                                                                                                                                                                                                                                                                             
  DELETE FROM d WHERE k = 4;                             -- snapshot 4: inlined delete
  DELETE FROM d WHERE k = 6;                             -- snapshot 5: inlined delete

  -- control: per-row snapshots are correct when nothing filters on rowid
  SELECT rowid, snapshot_id, k FROM ducklake_table_deletions('dl','main','d',1,5) ORDER BY rowid;
  -- 1 3 2 / 3 4 4 / 5 5 6

  -- rowid projected + snapshot filter: correct
  SELECT rowid, k FROM ducklake_table_deletions('dl','main','d',1,5) WHERE snapshot_id = 5 AND rowid = 5;
  -- 5 6

  -- GAP 1: rowid filter-only, no snapshot filter -> file constant instead of the row's snapshot
  SELECT snapshot_id, k FROM ducklake_table_deletions('dl','main','d',1,5) WHERE rowid = 5;
  -- expected: 5 6      actual: 3 6

  -- GAP 2: rowid filter-only + snapshot filter -> the deferred filter sees the constant -> row lost
  SELECT k FROM ducklake_table_deletions('dl','main','d',1,5) WHERE snapshot_id = 5 AND rowid = 5;
  -- expected: 6        actual: 0 rows

```

oleg-milanov:
Thanks @pdet for the review! Added the test and fixed the two gaps it exposed (rowid filter-only now keeps per-row snapshots).

PR #1378, opened 2026-08-11
Only apply global statistics to current-snapshot table scans

### LINKED ISSUE #1377: Historical, change-feed, and snapshot-pinned scans use current column statistics and return wrong results
### What happens?

DuckLake's global column statistics are not versioned: `DuckLakeMetadataManager::GetGlobalTableStats` reads `ducklake_table_stats` / `ducklake_table_column_stats` with no snapshot predicate.
`DuckLakeStatistics` then hands those current statistics to any scan without checking which snapshot the scan is reading.

Two distinct wrong results follow:
1. rows disappear: a predicate is eliminated against current statistics. `WHERE i IS NULL` returns nothing on a snapshot that does contain a `NULL`.
2. `count(column)` counts NULLs: once current statistics record a column as having no NULLs, `count(i)` is rewritten to `count(*)`, which then counts rows the older snapshot stored as `NULL`.

### To Reproduce

```sql
ATTACH 'ducklake:/tmp/dl/meta.db' AS lake (DATA_PATH '/tmp/dl/data', DATA_INLINING_ROW_LIMIT 0);

CREATE TABLE lake.history(i INTEGER);
INSERT INTO lake.history VALUES (NULL), (1);                                    -- snapshot 2
DELETE FROM lake.history WHERE i IS NULL;                                       -- snapshot 3
CALL ducklake_rewrite_data_files('lake', 'history', delete_threshold => 0);     -- snapshot 4

-- 1. a row that exists in snapshot 2 is eliminated by the predicate
SELECT count(*) FROM lake.history AT (VERSION => 2) WHERE i IS NULL;

-- 2. count(i) counts the NULL row
SELECT count(i), count(*) FROM lake.history AT (VERSION => 2);
SELECT count(i), count(*) FROM ducklake_table_insertions('lake', 'main', 'history', 2, 2);

DETACH lake;
ATTACH 'ducklake:/tmp/dl/meta.db' AS lake (SNAPSHOT_VERSION 2);
SELECT count(i), count(*) FROM lake.history;
```

| query | observed | expected |
|---|---|---|
| `count(*) ... AT (VERSION => 2) WHERE i IS NULL` | 0 | 1 |
| `count(i), count(*) ... AT (VERSION => 2)` | 2, 2 | 1, 2 |
| `count(i), count(*)` over `ducklake_table_insertions(... 2, 2)` | 2, 2 | 1, 2 |
| `count(i), count(*)` on a catalog pinned at `SNAPSHOT_VERSION 2` | 2, 2 | 1, 2 |

Confirmed by disabling the optimizer:

```sql
SET disabled_optimizers='statistics_propagation';
SELECT count(*) FROM lake.history WHERE i IS NULL;   -- returns 1, the correct answer
```

### OS:

macOS 26.6.1

### DuckDB Version:

duckdb v1.6.0-dev11186

### DuckLake Version:

ducklake main @ c6501ad9e5a574c549a913cd77465d5f269d7c24

### DuckDB Client:

CLI

### Full Name:

Florian Valeye

### Affiliation:

altertable-ai

- [X] Yes, I have

- [X] Yes, I have

fvaleye:
Fixes #1377

## Problem

DuckLake's global column statistics are not versioned. `DuckLakeStatistics` exposes them to every scan, including historical, change-feed, and snapshot-pinned scans whose rows differ from the current table. This can eliminate matching predicates or rewrite `count(column)` to `count(*)`, returning incorrect results.

## Fix

Only expose global column statistics when all three conditions hold:
- the scan is a regular table scan
- the scan snapshot matches the transaction snapshot
- the catalog is not snapshot-pinned

`DuckLakeGetPartitionStats` already applied the same scan-kind and snapshot checks inline. 
They now live in `DuckLakeFunctionInfo::CanUseGlobalStats()` and both call sites use it. Other scans return `nullptr`, so the optimizer falls back to reading the selected data.

### DIFF
--- a/src/include/storage/ducklake_multi_file_list.hpp
+++ b/src/include/storage/ducklake_multi_file_list.hpp
@@ -45,9 +45,6 @@ public:
 	unique_ptr<NodeStatistics> GetCardinality(ClientContext &context) const override;
 	DuckLakeTableEntry &GetTable();
 	unique_ptr<MultiFileList> Copy() const override;
-	bool HasTransactionLocalData() const {
-		return !transaction_local_files.empty() || transaction_local_data;
-	}
 	vector<DuckLakeFileListExtendedEntry> GetFilesExtended() const;
 	const vector<DuckLakeFileListEntry> &GetFiles() const;
 	const DuckLakeFileListEntry &GetFileEntry(idx_t file_idx) const;
@@ -55,6 +52,7 @@ public:
 		return filter_info.get();
 	}
 
+	bool CanUseGlobalStats() const;
 	bool IsDeleteScan() const;
 	const DuckLakeDeleteScanEntry &GetDeleteScanEntry(idx_t file_idx);
 
--- a/src/include/storage/ducklake_scan.hpp
+++ b/src/include/storage/ducklake_scan.hpp
@@ -57,6 +57,7 @@ struct DuckLakeFunctionInfo : public TableFunctionInfo {
 	unique_ptr<DuckLakeSnapshot> start_snapshot;
 
 	shared_ptr<DuckLakeTransaction> GetTransaction();
+	bool CanUseGlobalStats();
 };
 
 } // namespace duckdb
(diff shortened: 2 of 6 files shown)


pdet:
Thanks, this looks great!

PR #1380, opened 2026-08-11
Added clearer error message for missing secrets

okruitho:
This PR gives the end user better feedback on missing secrets when loading ducklake.


pdet:
Thanks!

PR #1384, opened 2026-08-13
Fix main CI by dropping leaked scoped metadata schemas in the postgres test config

GavinPizza:
Fix the failing Postgres job by dropping both leftover schemas in `on_init`. `metadata_schema_scoped_attach.test` leaves `scoped_meta_1064` and `junk_1064` in `ducklakedb`, and the `data_inlining` cleanup tests count inlined tables across all schemas.


GavinPizza:
Superseded. The same two DROP SCHEMA IF EXISTS statements landed on main in 565dc109 (via #1392), in the opposite order, so this now conflicts rather than no-ops. Closing.

PR #1391, opened 2026-08-14
Reclaim inlined rows that no surviving snapshot can read

### LINKED ISSUE #1390: Dead inlined rows are never reclaimed by expire_snapshots + cleanup_old_files
### What happens?

Deleted rows in an inlined data table never go away. `ducklake_expire_snapshots` and `ducklake_cleanup_old_files` both leave them in place, even after every snapshot that could read them has been expired. The only thing that removes them is `ducklake_flush_inlined_data`, and that writes the table to parquet.

This looks like the case #1067 raised, which was closed as fixed by #1145. #1237 points at #1321 for the same area. Both work on whole tables, not on rows inside a table that is still in use, and the repro below is on a build that includes #1321. The last comment on #1067, asking whether removed data inside inlined tables was covered, never got a reply.

So if you keep data inlined on purpose and delete rows, there is no way to reclaim the space short of converting the table to parquet, which is the thing #1067 was trying to avoid. The row limit doesn't force a flush either, because it only counts live rows.

#1219 and #1237 are both waiting on a reproducible example, so here is one.

### To Reproduce

```sql
ATTACH 'ducklake:min.db' AS dl (DATA_PATH 'min_files', METADATA_CATALOG 'meta', DATA_INLINING_ROW_LIMIT 100);
CREATE TABLE dl.t(i INTEGER);
INSERT INTO dl.t VALUES (1),(2),(3);
DELETE FROM dl.t WHERE i=2;

CALL ducklake_expire_snapshots('dl', older_than => now());
CALL ducklake_cleanup_old_files('dl', cleanup_all => true);

SELECT snapshot_id FROM ducklake_snapshots('dl');
-- 3          <- every earlier snapshot is gone

SELECT * FROM meta.ducklake_inlined_data_1_1 ORDER BY row_id;
-- row_id | begin_snapshot | end_snapshot | i
--      0 |              2 |         NULL | 1
--      1 |              2 |            3 | 2   <- dead, still here
--      2 |              2 |         NULL | 3
```

The row with `end_snapshot = 3` can't be read by the one surviving snapshot, but both cleanup calls leave it alone. `SELECT * FROM dl.t` still returns 1 and 3, so the results are fine. It is only space.

`CALL ducklake_flush_inlined_data('dl')` does clear it out, but that writes the table to parquet.

There is no upper bound on this. 150 insert-then-delete cycles against a table with `DATA_INLINING_ROW_LIMIT 10`, followed by the same two cleanup calls, leave 0 live rows, 150 rows still in `meta.ducklake_inlined_data_1_1`, and nothing written to parquet. The row limit was passed 15 times over and never tripped, because it counts live rows and there aren't any.

### OS:

### DuckDB Version:

v1.6.0-dev12414, source build @ `ac5c6d11c3`

### DuckLake Version:

ducklake `main` @ `3d8e24a83efdbb4620a108eebb590cda4f502e82`

### DuckDB Client:

CLI

### Hardware:

Not performance-related.

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [X] Yes, I have

- [X] Yes, I have

GavinPizza:
Closes #1390

`ducklake_expire_snapshots` never removes inlined rows whose `end_snapshot` has dropped below the oldest surviving snapshot. Once every snapshot that could read such a row has been expired the row is unreachable, but it stays in the metadata catalog, and nothing bounds how many pile up. 150 insert-then-delete cycles against a table with `DATA_INLINING_ROW_LIMIT 10` leave 0 live rows and 150 physical rows after `expire_snapshots` and `cleanup_old_files`, with nothing written to Parquet. The row limit counts live rows, so it never trips and never forces a flush.

Flushing was the only thing that reclaimed them, and a flush materializes the table to Parquet, which is the thing a workload that opts into inlining is trying to avoid. `rewrite_data_files` and `merge_adjacent_files` leave the rows alone, and `CHECKPOINT` clears them only because its first step is that same flush. So this is a no-op for anyone who checkpoints, and the people it helps are the ones calling `expire_snapshots` and `cleanup_old_files` directly and wanting the data to stay inlined.

## Changes

`ducklake_metadata_manager.cpp`: `DeleteSnapshots` already sweeps unreachable rows out of every other table carrying the `begin_snapshot` / `end_snapshot` pair. Inlined data tables were the one it skipped. It has to happen during expire rather than in `cleanup_old_files`, since expire is the only maintenance function that knows the surviving-snapshot set.

They get their own loop instead of joining the existing list because a v1.1 catalog prefixes the two bound columns with `_ducklake_`, so the statement text differs by catalog version. Running after the drop block means dropped tables are already unregistered, so no `table_id` filter is needed.

The inlined tables themselves are left alone. Keeping one alive for a live `(table_id, schema_version)` is intended, since the next insert reuses it.

Reverting just the `ducklake_metadata_manager.cpp` hunk makes the new test fail at its first post-expire assertion, 3 surviving rows where 2 is expected, on both the DuckDB and Postgres catalogs.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -5538,6 +5538,32 @@ WHERE end_snapshot IS NOT NULL AND NOT EXISTS(
 		}
 	}
 
+	// A v1.1 catalog prefixes the bound column names, so inlined tables cannot share the statement
+	// above. Runs after the drop block, so dropped tables are already unregistered and this needs
+	// no table_id filter.
+	{
+		auto inlined = Query("SELECT table_name FROM {METADATA_CATALOG}.ducklake_inlined_data_tables;");
+		if (inlined->HasError()) {
+			inlined->GetErrorObject().Throw("Failed to list inlined-data tables for cleanup in DuckLake: ");
+		}
+		auto col_names = InlinedColNames();
+		for (auto &row : *inlined) {
+			auto inlined_table_name = row.GetValue<string>(0);
+			auto result = Execute(StringUtil::Format(R"(
+DELETE FROM {METADATA_CATALOG}.%s
+WHERE %s IS NOT NULL AND NOT EXISTS(
+    SELECT snapshot_id
+    FROM {METADATA_CATALOG}.ducklake_snapshot
+    WHERE snapshot_id >= %s AND snapshot_id < %s
+);)",
+			                                         SQLIdentifier(inlined_table_name), col_names.end_snapshot,
+			                                         col_names.begin_snapshot, col_names.end_snapshot));
+			if (result->HasError()) {
+				result->GetErrorObject().Throw("Failed to delete from " + inlined_table_name + " in DuckLake: ");
+			}
+		}
+	}
+
 	// clean up macro implementation and parameters for deleted macros
 	tables_to_delete_from = {"ducklake_macro_impl", "ducklake_macro_parameters"};
 	for (auto &delete_tbl : tables_to_delete_from) {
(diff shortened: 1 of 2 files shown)


pdet:
Thanks, LGTM

PR #1394, opened 2026-08-19
Fix missing phantom protection for zero-row DELETEs

### LINKED ISSUE #1393: Concurrent DELETE-then-INSERT where DELETE matches no rows does not conflict check and results in duplicates
### What happens?

A `DELETE` whose predicate matches **zero rows** takes part in no conflict check, so two
concurrent transactions can both delete-nothing and then both insert the same key. The
result is silent duplicate rows from a `DELETE`+`INSERT` upsert that should have been
serialised.

`DuckLakeDelete::Finalize` returns early when `global_state.deleted_rows.empty()`, before
anything is recorded on the transaction:

```cpp
auto &global_state = input.global_state.Cast<DuckLakeDeleteGlobalState>();
if (global_state.deleted_rows.empty()) {
    return SinkFinalizeType::READY;
}
```

Because nothing is recorded, the table never enters `tables_deleted_from`, and the commit
path has no basis on which to conflict the transaction against a concurrent insert to the
same table. Each writer independently observes the key as absent, deletes nothing, and
inserts it.

The transaction that "loses" is never told. Both commits succeed and the duplicate is
persisted, so this is silent corruption rather than a visible error.

This is the zero-row case only. A `DELETE` matching **at least one** row already conflicts
correctly at table granularity, which is what makes the behaviour inconsistent: an upsert
is protected once the key exists, but unprotected the first time it is written.

Reproduces on the `duckdb`, `sqlite`, `quack` and `postgres` catalog backends, so it is in
the DuckLake extension rather than in any one catalog. 

Affects `DELETE`, `UPDATE` and `MERGE INTO`, which all plan through `DuckLakeDelete`.

For contrast, PR #906 addressed a neighbouring case, but its regression test
(`test/sql/concurrent/concurrent_insert_delete_conflict.test`) cannot detect this one: it
pre-seeds keys 0-9 and then deletes a key that always matches a real row, so the zero-row
path is never exercised.

### To Reproduce

Minimal, self-contained, no external catalog required:

```python
import duckdb, threading, tempfile, os

root = tempfile.mkdtemp()
cat  = os.path.join(root, "catalog.ducklake")
data = os.path.join(root, "data")

def attach(con):
    con.execute(f"ATTACH 'ducklake:{cat}' AS dl (DATA_PATH '{data}', DATA_INLINING_ROW_LIMIT 0)")

setup = duckdb.connect()
attach(setup)
setup.execute("CREATE TABLE dl.tbl(key INTEGER, writer INTEGER)")
setup.close()

# key 42 does not exist, so every DELETE below matches zero rows
barrier = threading.Barrier(3)

def writer(n):
    con = duckdb.connect()
    attach(con)
    barrier.wait()
    try:
        con.execute("BEGIN")
        con.execute("DELETE FROM dl.tbl WHERE key = 42")
        con.execute(f"INSERT INTO dl.tbl VALUES (42, {n})")
        con.execute("COMMIT")
    except Exception as e:
        print(f"writer {n} conflicted (expected for all but one): {e}")
    con.close()

ts = [threading.Thread(target=writer, args=(i,)) for i in range(3)]
[t.start() for t in ts]
[t.join() for t in ts]

check = duckdb.connect()
attach(check)
print("rows with key=42 :", check.execute(
    "SELECT COUNT(*) FROM dl.tbl WHERE key = 42").fetchone()[0], "(expected 1)")
```

**Expected:** exactly one row with `key = 42`. Every writer but one should fail its
`COMMIT` with a transaction conflict and be retried by the client.

**Actual:** 2 or 3 rows with `key = 42`, and every writer reports success. Re-run a few
times; it is a race, so it does not duplicate on every run.

An equivalent sqllogictest, which fails on current `main`:

```
statement ok
ATTACH 'ducklake:{DUCKLAKE_CONNECTION}' AS ducklake (DATA_PATH '{DATA_PATH}/x', DATA_INLINING_ROW_LIMIT 0)

statement ok
CREATE TABLE ducklake.tbl(key INTEGER, writer INTEGER);

concurrentloop i 0 4

statement ok
SET ducklake_max_retry_count = 100

statement ok
BEGIN

statement ok
DELETE FROM ducklake.tbl WHERE key = 42

statement ok
INSERT INTO ducklake.tbl VALUES (42, {i})

statement maybe
COMMIT
----

endloop

query I
SELECT COUNT(*) FROM ducklake.tbl WHERE key = 42
----
1
```

### OS:

Windows 11 x86_64

### DuckDB Version:

v1.5.5 (also reproduced against `main`, duckdb submodule `ac5c6d11`)

### DuckLake Version:

`v1.5-variegata` @ `5ef9e03d` and `main` @ `4a4663f3`

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Adam Smoluk

### Affiliation:

Bazefield

I have not tested with any build

Yes

- [x] Yes, I have

- [x] Yes, I have

bazeadam:
# Fix missing phantom protection for zero-row DELETEs in concurrent transactions

## Summary

A `DELETE` whose predicate matches **zero rows** currently leaves no trace in the
transaction's change set. As a result it is invisible to `CheckForConflicts`, and a
concurrent transaction that **inserts a row matching that same predicate** is allowed to
commit alongside it. The classic delete-then-insert upsert therefore silently produces
**duplicate keys** — the exact anomaly the conflict checker exists to prevent.

This PR records *delete intent* (the predicate was evaluated against the table) rather than
*delete result* (rows were actually modified), which is the conventional model in
relational systems.

## Reproduction

Two transactions run the standard upsert shape against a key that does **not yet exist**:

```sql
-- both transactions, concurrently
BEGIN;
DELETE FROM t WHERE k = 42;   -- matches 0 rows -> currently invisible to conflict detection
INSERT INTO t VALUES (42, ...);
COMMIT;
```

Both commit. The table now contains key `42` twice.

The same happens with `MERGE INTO ... WHEN MATCHED THEN UPDATE / WHEN NOT MATCHED THEN
INSERT` when the key is absent, because the update branch plans a `DuckLakeDelete` that
matches nothing.

Observed rates before the fix (3 concurrent writers, seeded table, brand-new key):

| backend  | duplicate-producing runs |
|----------|--------------------------|
| duckdb   | 36 %                     |
| quack    | 92 %                     |
| postgres | reproduces               |
| sqlite   | n/a (catalog serialises) |

The variation across backends is a race-window difference, not a mechanism difference —
all four run the same client-side conflict check.

## Root cause

`DuckLakeDelete::Finalize` returns before touching the transaction when nothing was
deleted:

```cpp
auto &global_state = input.global_state.Cast<DuckLakeDeleteGlobalState>();
if (global_state.deleted_rows.empty()) {
    return SinkFinalizeType::READY;   // <-- transaction never learns the DELETE ran
}
auto &transaction = DuckLakeTransaction::Get(context, table.catalog);
```

`tables_deleted_from` is therefore only populated when at least one row is removed, and
`CheckForConflicts` has nothing to test the concurrent insert against.

Note this is a *sink* `Finalize`: it runs once at pipeline end regardless of how many rows
arrived, which makes it the correct place to record that the predicate was evaluated.

## Fix

1. Record intent unconditionally, before the early return:

```cpp
auto &transaction = DuckLakeTransaction::Get(context, table.catalog);
transaction.MarkDeleteAttempted(table.GetTableId());
if (global_state.deleted_rows.empty()) {
    return SinkFinalizeType::READY;
}
```

2. Check it during conflict detection:

```cpp
for (auto &table_id : changes.tables_delete_attempted) {
    ConflictCheck(table_id, other_changes.inserted_tables, "delete from table", "inserted into it");
    ConflictCheck(table_id, other_changes.tables_inserted_inlined, "delete from table", "inserted into it");
}
```

The new `tables_delete_attempted` set is added **only** to `TransactionChangeInformation`
(local transaction state), never to `SnapshotChangeInformation`. It is consequently not
written to the changelog and there is **no metadata or wire-format change** for the
DuckDB/SQLite/Postgres catalogs.

For the server-side (quack) commit path the set is additionally staged, mirroring exactly
how `tables_deleted_from` is already handled, because that path rebuilds the change set
from staged artifacts rather than from local memory.

### Why only insert conflicts are checked

The hazard being closed is specifically a **phantom**: rows appearing that match a
predicate already evaluated by a concurrent transaction. Only inserts (regular and
inlined) can create them. Compaction rewrites files without changing logical row content,
and drop/alter are already covered by the existing schema-version checks, so the intent
loop deliberately does not duplicate them.

### Coverage of DML forms

All three DML paths funnel through `DuckLakeDelete`, so the single hook covers them:

| statement    | path |
|--------------|------|
| `DELETE`     | `DuckLakeCatalog::PlanDelete` -> `DuckLakeDelete::PlanDelete` |
| `UPDATE`     | `DuckLakeUpdate::PlanUpdateOperator` -> `DuckLakeDelete::PlanDelete` |
| `MERGE INTO` | update action -> `DuckLakeUpdate`; delete action -> `catalog.PlanDelete` |

## Precedent

Recording the predicate rather than the result is the established behaviour elsewhere:

- **SQL standard** defines phantom (P3) in terms of the *search condition*, not whether
  rows were found.
- **PostgreSQL** serializable isolation takes SIREAD predicate locks on what was scanned:
  *"Each tuple read ... will be locked, whether or not it meets selection criteria."* A
  sequential scan takes a relation-level predicate lock.
- **SQL Server** documents the direct analogue — a "singleton fetch of nonexistent data"
  still takes a `RangeS-S` lock on the gap.
- **Delta Lake** tracks `readPredicates` on the transaction and raises
  `ConcurrentAppendException`; a whole-table predicate degrades to table-level conflict.
- **Apache Iceberg** exposes `conflictDetectionFilter` / `validateNoConflictingData`.

Table granularity matches the coarsest of these (Postgres relation-level locks, Delta at
`WriteSerializable`) and is intentionally chosen as a correct first implementation. It can
be refined to file statistics later; that refinement must treat missing/NULL statistics and
non-range predicates as conflicting, since disjointness cannot be proven in those cases.

## This does not introduce a new over-blocking regime

This is the question the change most deserves scrutiny on, so it was measured rather than
argued. Matched A/B on one binary, reverting only the conflict-loop addition; 5 runs each,
3 concurrent writers:

| scenario | unpatched | patched | |
|---|---|---|---|
| S1 disjoint partitions, **empty** table | 3/3 committed | **1/3** | now conflicts |
| S2 same partition, disjoint objects, **empty** table | 3/3 | **1/3** | now conflicts |
| S3 same partition, same objects, **empty** table | 3/3, **20 duplicate rows** | 1/3, **0 duplicates** | bug fixed |
| S4 seeded, same objects | 1/3 | 1/3 | unchanged |
| S5 seeded, disjoint objects | 1/3 | 1/3 | unchanged |
| S6 seeded, disjoint partitions | 1/3 | 1/3 | unchanged |
| S7 different tables | 3/3 | 3/3 | unchanged |
| S8 append-only, no DELETE | 3/3 | 3/3 | unchanged |
| S10 seeded, disjoint objects + retry(5) | 3/3 | 3/3 | unchanged |
| S1R = S1 + retry(5) | 3/3 | **3/3**, 0 duplicates | recovered by retry |
| S2R = S2 + retry(5) | 3/3 | **3/3**, 0 duplicates | recovered by retry |

The decisive rows are S5/S6 versus S1/S2. **S5 and S6 already commit only 1/3 on the
unpatched build** — DuckLake already conflicts at table granularity whenever a DELETE
matches at least one row. S1 and S2 are the empty-table analogues of S6 and S5: identical
workloads whose deletes happen to match nothing. Unpatched they escape the conflict, and
that escape *is* the bug.

After the fix, S1 behaves exactly like S6 and S2 exactly like S5. **The patch makes
zero-row deletes behave identically to non-zero-row deletes.** It removes an
inconsistency; it does not add a new class of blocking.

S7, S8 and S10 bound the blast radius: different tables, append-only workloads, and
retry-enabled upserts are entirely unaffected. Retry remains the documented mitigation —
with `ducklake_max_retry_count = 5`, disjoint-object upserts still commit 3/3.

S1R and S2R close the loop on the two rows that do change. Re-running S1 and S2 with
retry(5) — the same replay a client already needs for S5 and S6 on stock DuckLake — all
three writers commit, every row lands, and there are no duplicates, across 5/5 runs. The
cost is 5–6 commit attempts where 3 would be ideal. So the newly-conflicting cases are
not newly-failing cases; they join the set that retry already resolves.

## Tests

Two new tests under `test/sql/concurrent/`:

- `concurrent_upsert_absent_key.test` — concurrent delete+insert of a key that does not
  exist (27 assertions)
- `concurrent_merge_into_absent_key.test` — the `MERGE INTO` form (15 assertions)

Both are verified to actually detect the defect, by reverting only the conflict-loop change
on an otherwise identical binary and rebuilding:

| | duckdb catalog | quack catalog |
|---|---|---|
| conflict loop reverted | **10/10 runs failed** | **3/3 runs failed** |
| patch applied | 0/10 | 0/3 |

On the quack backend the reverted build produces four copies of the affected key rather
than two, consistent with its wider race window.

Both backends are green on the full concurrent directory: duckdb 319 assertions across 9
test cases, quack 8/8 (the remainder skipped by `test/configs/quack.json`).

The complete DuckLake test suite passes on the patched build: **21456 assertions in 543
test cases, 47 skipped, 0 failures.**

The change was developed on `v1.5-variegata` (duckdb v1.5.5) and also built and tested on
`main` (`4a4663f3`, duckdb `ac5c6d11`): 741/741 build, both new tests pass, concurrent
directory green.

## Relationship to #906

#906 aimed at this behaviour ("Both delete-and-insert & merge into"), but its regression
test cannot observe it, which is why the defect is still reachable:

- line 31 pre-seeds keys `0..9`;
- line 42 deletes key `i` where `i` is `0` or `1`, so the DELETE **always matches a real
  row** and the existing `tables_deleted_from` path is exercised;
- line 46 inserts key `100 + i`, a **different** key, so the statement pair is not an
  upsert;
- line 53 asserts only that no duplicate keys exist, which is trivially satisfied because
  no two writers ever target the same key.

The tests added here cover the uncovered case: the deleted key and the inserted key are the
same, and that key is absent at the start.

___

Closes #1393


src/storage/ducklake_transaction_state.cpp
```diff
@@ -225,6 +225,11 @@ void DuckLakeTransactionState::CheckForConflicts(const TransactionChangeInformat
 		ConflictCheck(table_id, other_changes.inserted_tables, "delete from table", "inserted into it");
 		ConflictCheck(table_id, other_changes.tables_inserted_inlined, "delete from table", "inserted into it");
 	}
+	// a delete that matched no rows must still conflict with concurrent inserts to the same table
+	for (auto &table_id : changes.tables_delete_attempted) {
+		ConflictCheck(table_id, other_changes.inserted_tables, "delete from table", "inserted into it");
+		ConflictCheck(table_id, other_changes.tables_inserted_inlined, "delete from table", "inserted into it");
+	}
```
pdet: Are these duplicated from above? Maybe we could use a helper?

pdet:
Thanks for the PR, LGTM.

PR #1399, opened 2026-08-20
Recast persisted min/max stats when a column type changes

### LINKED ISSUE #1307: FLOAT->DOUBLE column retype corrupts min/max stats, causing wrong zone-map pruning

GavinPizza:
Closes #1307

Column min/max are persisted as strings written under the column's type, so a catalog-only `ALTER COLUMN c TYPE DOUBLE` on a FLOAT column reinterprets them under the new type on reload. `"0.1"` reparsed as DOUBLE is `0.1`, while the real FLOAT `0.1` widened is `0.10000000149011612`. The recorded bound stops covering the data and the zone map prunes live rows, so queries return wrong results with no error.

The bound is taken conservatively, as the wider of two candidate readings, rather than the value cast through the source type, because which type the data is physically stored as cannot be recovered from the catalog: a promotion chain leaves intermediate types no file ever had, compaction carries the sources' physical type forward, and time travel reads the same rows back under an older type. Widening can only over-cover, never prune. The cast also runs in C++ rather than in SQL, because a SQL cast would be a silent no-op on SQLite, which maps FLOAT and DOUBLE both to VARCHAR.

Cost: this makes a retyping ALTER O(number of files) where it was O(1), measured at 1.4s for 2000 files on a FLOAT to DOUBLE change. Only approximate source types can produce a rewrite at all, so every integer, decimal and temporal widening is skipped entirely and stays O(1).

A transaction that both retypes and inserts can stage its file on either side of the ALTER, and only one of those needs correcting. File stats are built with the field's type as of the write, so that type is the provenance: a file staged before the ALTER is corrected, one staged after is already exact and is left alone. Where that type is trustworthy the correction is exact rather than widened, which matters because MIN and MAX are answered from these bounds when no row has been deleted, and a widened bound would make them report a value no row holds.

`alter_type_stats_recast.test` fails on main both on the persisted `min_value` and on a wrong-result probe that returns 0 rows where 1 qualifies. The correction also has to survive a commit that loses a race and replays its batch, and a writer that lands a file underneath it; `alter_type_stats_recast_concurrent.test` covers both and is red on main on the same wrong-result probe. The probes run with `filter_pushdown` disabled for an unrelated reason: pushdown narrows the constant to the file's physical type and drops the boundary row whatever the stats say. That is duckdb/duckdb#24900, fixed by duckdb/duckdb#24901, which is not yet in the pinned DuckDB.

### DIFF
--- a/src/include/storage/ducklake_commit_state.hpp
+++ b/src/include/storage/ducklake_commit_state.hpp
@@ -29,6 +29,8 @@ struct NewTableInfo {
 	vector<DuckLakeNewColumn> new_columns;
 	vector<DuckLakeTableInfo> new_inlined_data_tables;
 	vector<DuckLakeSortInfo> new_sort_keys;
+	//! Tables whose commit includes an ALTER ... TYPE.
+	set<TableIndex> retyped_tables;
 };
 
 struct NewMacroInfo {
@@ -42,7 +44,18 @@ struct NewNameMapInfo {
 struct NewDataInfo {
 	vector<DuckLakeFileInfo> new_files;
 	vector<DuckLakeInlinedDataInfo> new_inlined_data;
+	//! Tables whose global stats this commit rewrites from memory; RecastCommittedStats must not correct them again.
+	set<TableIndex> global_stats_written;
 };
+
+struct DuckLakeRetypedColumn {
+	//! The committed type before this ALTER. Only a guess at what the stats strings were rendered under, see
+	//! TryRecastStatsBound.
+	LogicalType source_type;
+	LogicalType target_type;
+};
+
+using DuckLakeRetypePlan = map<TableIndex, map<FieldIndex, DuckLakeRetypedColumn>>;
 
 struct CompactionInformation {
 	vector<DuckLakeCompactedFileInfo> compacted_files;
--- a/src/include/storage/ducklake_metadata_info.hpp
+++ b/src/include/storage/ducklake_metadata_info.hpp
@@ -361,6 +361,9 @@ struct DuckLakeNewColumn {
 	TableIndex table_id;
 	DuckLakeColumnInfo column_info;
 	optional_idx parent_idx;
+	//! Set only when an ALTER ... TYPE drops and re-creates this field; empty for ADD COLUMN and RENAME. Not
+	//! persisted - a table created and altered in the same transaction has no committed old type to look up.
+	string promoted_from_type;
 };
 
 struct DuckLakeCatalogInfo {
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -399,9 +399,19 @@ public:
 	//! Caller substitutes `{METADATA_CATALOG}` / `{SNAPSHOT_ID}` and executes via the commit context's executor.
 	static string ReadInlinedDataAggregatesSql(const string &inlined_table_name, const string &select_list,
 	                                           const DuckLakeInlinedColNames &col_names);
-	static string ReadFileColumnStatsForTableSql(TableIndex table_id);
+	//! An empty `column_ids` reads every column. The recast pass passes the retyped columns only: it runs over every
+	//! file of the table, so it must not also drag in every column of a wide one. A non-empty filter also drops files
+	//! that have no stats row at all, which the LEFT JOIN would otherwise keep.
+	static string ReadFileColumnStatsForTableSql(TableIndex table_id, const set<FieldIndex> &column_ids = {});
 	//! Throws on a failed inlined data read, hinting at the migration for legacy named catalogs
 	void CheckInlinedDataReadError(QueryResult &result, const string &inlined_table_name);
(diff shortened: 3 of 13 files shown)


pdet:
Is this already fixed by https://github.com/duckdb/ducklake/pull/1397 ?

GavinPizza:
No, it still reproduces when built from `a92e65b8`:

```sql
ATTACH 'ducklake:meta.db' AS lake (DATA_PATH 'data', DATA_INLINING_ROW_LIMIT 0);
USE lake;
CREATE TABLE t(c FLOAT);
INSERT INTO t VALUES (0.03), (0.05), (0.07), (0.1);
ALTER TABLE t ALTER COLUMN c SET DATA TYPE DOUBLE;

SELECT count(*) FROM t WHERE c >= 0.1000000005;  -- 0, the 0.10000000149011612 row qualifies
SELECT min(c), max(c) FROM t;                    -- 0.03, 0.1, neither value is in the table
```

Both halves of #1397 miss this shape. The cache stamp and the `GetColumnStats` type check don't help, because `ConstructStatsMap` rebuilds the stats at the column's current type, so the check passes and the reinterpreted bound is served. The `MergeStats` invalidation doesn't cover it either: after `ALTER` plus an `INSERT`, the persisted global min is still `0.03` on `a92e65b8`.

The stale strings feed both consumers, `statistics_propagation` for the `min`/`max` above and the zone-map pruning SQL for the count. Correcting them at the ALTER is what fixes both.

`alter_type_stats_recast_regressions.test:300` is red on the rebase. This is not from this branch and instead the block inserts a NULL first, and on `a92e65b8` that column never gets its bounds back. Filed as #1411. Leaving it red rather than weakening it, since it goes green on a rebase once that lands.

Issue #1351, opened 2026-07-28
Pluggable key-management (envelope encryption) for data-file keys


tiagotaveiragomes:
## Problem

In `ENCRYPTED` mode the per-file encryption key is stored in the catalog
(`ducklake_data_file.encryption_key`) as base64 of the raw key - i.e. in
cleartext (`Blob::ToBase64` on write, `Blob::FromBase64` on read in
`ReadDataFile`). The catalog is therefore a plaintext key store: anyone with
catalog read access can decrypt every data file.

This blocks the zero-trust deployment model (untrusted object storage + a
separately-guarded catalog) for regulated data. A catalog compromise becomes a
full data compromise, and there is no way to root the per-file keys in a
KMS/HSM.

## Proposal

Introduce an optional `EncryptionKeyProvider` so per-file keys are stored
**wrapped** (envelope encryption) and unwrapped on read via a configured
provider. Default = identity (passthrough), preserving today's behaviour and
on-disk/catalog format byte-for-byte.

- `wrap(plaintext_key) -> (wrapped_blob, provider, key_id)`
- `unwrap(wrapped_blob, provider, key_id) -> plaintext_key`
- Selected by an `ATTACH` option; provider credentials via the existing secret
  mechanism.
- New catalog columns `encryption_key_provider`, `encryption_key_id`
  (NULL = raw key, fully back-compatible). This is a spec change, so it targets
  `main`.
- Wrap at commit-time serialization; unwrap at `ReadDataFile` / `ReadDeleteFile`.
  The Parquet cipher path is untouched.
- Ships with an in-tree test provider (e.g. AES-KW with a static key) for
  coverage across the sqlite/postgres/quack catalog backends.

The per-key `key_id` also enables KEK rotation (re-wrap without rewriting data)
and per-file key provenance for audit.

## Related hardening (happy to split into separate PRs)

- Key bytes are currently drawn from `RandomEngine` (PCG32, a non-cryptographic
  PRNG) in `GenerateEncryptionKey`. Encryption keys should come from a CSPRNG.
- Key size is a hardcoded 16 bytes (128-bit); add an AES-256 option.

## Coordination

Per the contribution guide (new features / spec changes coordinate first,
target `main`), opening this to align on the interface shape and the catalog
columns before implementing. Happy to bring a PR once the design is agreed.

mrendi29:
+1

yokasaka:
+1

mrendi29:
[...]

PR #1403, opened 2026-08-23
fix(encryption): CSPRNG-generated, 256-bit data-file keys

tiagotaveiragomes:
## What

Two related hardening fixes to per-file encryption key generation, split out
per #1351's "happy to split into separate PRs":

1. `DuckLakeCatalog::GenerateEncryptionKey` drew the per-file AES key from
   `RandomEngine` (DuckDB's general-purpose PCG PRNG), which is not
   cryptographically secure: its internal state is recoverable from output,
   and every key in a write session comes from the same per-context stream,
   so one leaked key helps predict its siblings. Switched to DuckDB's
   cryptographic RNG (`EncryptionState::GenerateRandomData`, mbedtls-backed) -
   the same source already used for GCM nonces and WAL/block encryption.

2. Key size was hardcoded to 16 bytes (AES-128) with no way to raise it. Now
   32 bytes (AES-256). Parquet modular encryption accepts 128/192/256-bit
   keys (DuckDB's own writer enforces this in `parquet_crypto.cpp`, and its
   mbedtls wrapper switches on all three lengths), and the read path never
   assumes a fixed key length - a stored key is used at whatever length it
   was written, so files encrypted with the older 16-byte keys keep
   decrypting unchanged. No migration or re-encryption is implied.

Targets `v1.5-variegata` per CONTRIBUTING - this is a bug fix with no
DuckLake spec change and no dependency on DuckDB `main`.

## Why

Both are hardening items called out in #1351 as related-but-separable from
the pluggable key-management proposal there, since they land independently
of the wrap/unwrap interface discussion.

Refs #1351


tiagotaveiragomes:
Reopening to trigger CI — no workflow run has ever been attached to this branch since it was opened.

tiagotaveiragomes:
Housekeeping note for reviewers: the trailing `(#4)` and `(#5)` in the two commit subjects are stale numbers from my own fork's PR sequence, not references to issues in this repository. GitHub renders them as links to unrelated items here — please disregard them. Likewise the `cherry picked from commit ...` trailers point at SHAs that exist only in my fork.

I have deliberately left the history unrewritten rather than force-pushing, so the review state on this PR stays intact. Happy to reword the subjects if you would prefer a clean history before merge.

mrendi29:
I wonder if @pdet  can take a look at this one. I think this change is very valuable for an enterprise setting where AES-256 is the default acceptable encryption level.

tiagotaveiragomes:
I will be more than happy to take the work myself 😀 just let me know what must be corrected or improved

src/storage/ducklake_insert.cpp
```diff
@@ -714,13 +717,27 @@ string DuckLakeCatalog::GenerateEncryptionKey(ClientContext &context) const {
...
+	// 32 bytes = AES-256. The size was previously hardcoded to 16 (AES-128) with no way to raise
+	// it. Parquet modular encryption accepts 128, 192 or 256-bit keys - DuckDB's own writer says
+	// so explicitly ("Invalid AES key. Must have a length of 128, 192, or 256 bits (16, 24, or 32
+	// bytes)", parquet_crypto.cpp) and its mbedtls wrapper switches on all three - so 256 costs
+	// nothing but the extra 16 bytes per file in the catalog.
+	//
+	// This affects key GENERATION only. The key length is never assumed on the read path: a stored
+	// key is used at whatever length it was written, so data files encrypted with the older 16-byte
+	// keys keep decrypting unchanged. There is no migration and no re-encryption implied by this
+	// change - it only means keys minted from here on are 256-bit.
```
pdet: Let's avoid multiline comments unless necessary

pdet (commented):
Thanks for the PR, looks great, just a nit comment.
Let's also move to main, I don't think I'll do another DuckLake release for v1.5.x DuckDB

tiagotaveiragomes:
Thanks @pdet — both points addressed.

**Moved to `main`.** Retargeted and rebased onto `main` rather than `v1.5-variegata`. Only one conflict, in the include block, resolved as the union of both sides.

**Multiline comment trimmed.** It was a 14-line block; it is now four short lines in the style of the surrounding code:

```cpp
// generate an encryption key using the cryptographic RNG - RandomEngine is a non-cryptographic
// PRNG and must not be used for key material
// 32 bytes = AES-256; generation only, existing keys are read back at whatever length they were
// written, so no migration is implied
```

**Also cleaned up the history**, since it had to be rewritten for the retarget anyway: dropped the stale `(#4)`/`(#5)` subjects and the fork-only `cherry picked from commit ...` trailers I flagged earlier, so the log reads clean on `main`.

The diff is unchanged in substance — one file, `+13/-6`.

The earlier red was an infrastructure failure, not the change: `pull access denied for minio/mc` from Docker Hub in the MinIO setup step, after the build had already succeeded, so the suite never ran.

pdet:
Thanks!

PR #1405, opened 2026-08-24
Retry SQLite catalog commits when the database is locked

### LINKED ISSUE #1309: SQLite catalog COMMIT does not retry SQLITE_BUSY / database is locked
### What happens?

A DuckLake write using a SQLite catalog can fail during `COMMIT` with `database is locked`, even when DuckLake retries are set high and the catalog has a long `META_BUSY_TIMEOUT`.

This happens deterministically when:

1. connection A starts a DuckLake transaction and stages an insert (and therefore has read the metadata catalog),
2. connection B writes an ordinary table in the same SQLite catalog and briefly holds the write lock, then
3. connection A commits and tries to upgrade its metadata transaction to a writer.

This is related to #128, but the `busy_timeout` recommendation there cannot cover this read-to-write upgrade case. SQLite may return `SQLITE_BUSY` without invoking the busy handler when waiting would create a deadlock: B cannot commit until A releases its read lock, while A cannot upgrade while B is the pending writer. See the [SQLite busy-handler documentation](https://www.sqlite.org/c3ref/busy_handler.html).

### Reproduction

Environment:

- DuckDB Python package `1.5.4`
- DuckLake extension distributed for DuckDB `v1.5.4`, `extension_version = d318a545`
- Linux x86-64

Install with `python -m pip install duckdb==1.5.4`, then run:

```python
import tempfile
import threading
import time
from pathlib import Path

import duckdb


root = Path(tempfile.mkdtemp(prefix="ducklake-sqlite-busy-"))
catalog = root / "catalog.sqlite"
data = root / "data"
data.mkdir()


def connect(data_path=None):
    con = duckdb.connect()
    con.execute("INSTALL ducklake")
    con.execute("LOAD ducklake")
    con.execute("INSTALL sqlite")
    con.execute("LOAD sqlite")
    con.execute("SET ducklake_max_retry_count = 100")
    con.execute("SET ducklake_retry_wait_ms = 10")
    con.execute("SET ducklake_retry_backoff = 1.0")
    options = "METADATA_CATALOG 'meta', META_BUSY_TIMEOUT 10000"
    if data_path is not None:
        options = f"DATA_PATH '{data_path}', {options}"
    con.execute(f"ATTACH 'ducklake:sqlite:{catalog}' AS lake ({options})")
    return con


setup = connect(data)
setup.execute("CREATE TABLE lake.main.items (id INTEGER)")
setup.execute("CREATE TABLE meta.application_lease (id INTEGER PRIMARY KEY, value INTEGER)")
setup.execute("INSERT INTO meta.application_lease VALUES (1, 0)")
setup.close()

model = connect()
blocker = connect()
lock_acquired = threading.Event()


def briefly_hold_sqlite_write_lock():
    blocker.execute("BEGIN TRANSACTION")
    blocker.execute(
        "UPDATE meta.application_lease SET value = value + 1 WHERE id = 1"
    )
    lock_acquired.set()
    time.sleep(0.5)
    blocker.execute("COMMIT")


model.execute("BEGIN TRANSACTION")
model.execute("INSERT INTO lake.main.items VALUES (42)")

thread = threading.Thread(target=briefly_hold_sqlite_write_lock)
thread.start()
assert lock_acquired.wait(timeout=5)

started = time.monotonic()
try:
    model.execute("COMMIT")
except Exception as exc:
    print(f"COMMIT failed after {time.monotonic() - started:.3f}s")
    print(exc)
else:
    print(f"COMMIT succeeded after retrying for {time.monotonic() - started:.3f}s")

thread.join()
```

Output:

```text
COMMIT failed after 0.003s
Failed to flush changes into DuckLake: database is locked
```

The exact elapsed time varies slightly, but the failure occurs before connection B releases the lock and far below the configured 10-second busy timeout. Increasing `ducklake_max_retry_count` has no effect.

### Expected behavior

DuckLake should roll back only the failed metadata transaction, wait/back off, re-check conflicts, and retry the metadata commit. Once connection B commits after 500 ms, connection A should commit successfully without re-running the input query or rewriting the already-created data files.

### Suggested fix (AI)

`DuckLakeTransaction::RetryOnError` currently classifies primary-key/unique, conflict, and concurrent-access messages as retryable, but not SQLite's `database is locked` message ([current source](https://github.com/duckdb/ducklake/blob/d318a545571d7d46eb751fa2aa5f6f4389285d3c/src/storage/ducklake_transaction.cpp#L1261-L1275)).

The smallest fix is to classify SQLite `SQLITE_BUSY` / `database is locked` as retryable at the existing `FlushChanges` metadata-commit boundary. Prefer matching the structured SQLite error code if the catalog interface retains it. If it does not, matching the normalized `database is locked` message is consistent with the current classifier:

```cpp
if (StringUtil::Contains(message, "database is locked")) {
    return true;
}
```

The existing commit retry loop is the right layer: transaction changes are already retained across its attempts, and its rollback/conflict-check/prepare-retry behavior lets the competing SQLite writer finish without recomputing or rewriting the data operation.

1fanwang:
A DuckLake write using a SQLite catalog can fail during `COMMIT` with `database is locked`, even with a long SQLite busy timeout and DuckLake retries enabled. This happens when the transaction has read the catalog and another connection becomes the pending writer before the DuckLake commit upgrades to a write lock.

SQLite can skip its busy handler for this read-to-write deadlock. The bounded commit loop now retries SQLite metadata-lock errors only while rollback is safe, excludes failed `COMMIT` wrappers, and keeps post-commit cleanup outside the retry scope. The retry predicate uses simple branches because Emscripten 3.1.71 crashes while optimizing the equivalent compound expression.

Fixes https://github.com/duckdb/ducklake/issues/1309

## Testing

The regression uses two real DuckDB connections against one SQLite catalog. A zero-retry control requires the exact lock error, then the retry path commits the staged row once. An unrelated metadata error remains non-retryable.

<details>
<summary>Raw before/after output</summary>

```console
$ SOURCE=src/storage/ducklake_transaction.cpp
$ TEST=test/sql/issues/issue_1309.test
$ TESTER=build/release/test/unittest
$ RUNNER=duckdb/scripts/ci/run_tests.py
$ git checkout origin/main -- "$SOURCE"
$ cmake --build build/release --config Release -j8
$ "$TESTER" "$TEST"
Query unexpectedly failed! (test/sql/issues/issue_1309.test:71)!
COMMIT;
Failed to flush changes into DuckLake: database is locked
test cases: 1 | 0 passed | 1 failed

$ git checkout HEAD -- "$SOURCE"
$ cmake --build build/release --config Release -j8
$ "$TESTER" "$TEST"
All tests passed (34 assertions in 1 test case)

$ python "$RUNNER" --stabilize-tests --workers 1 --batch-size 1 "$TESTER" "$TEST"
stabilization rerun 9/9 for fast tests
all tests passed in 9s

$ "$TESTER" --test-config test/configs/sqlite.json --test-dir ./ "$TEST"
All tests passed (34 assertions in 1 test case)

$ make wasm_eh  # Emscripten 3.1.71, before the predicate rewrite
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/
Running pass "instcombine<max-iterations=1;no-verify-fixpoint>" on function "DuckLakeTransactionState::Commit"
clang++: error: clang frontend command failed with exit code 139

$ make wasm_eh  # after the predicate rewrite
[ 69%] Built target ducklake_storage
[100%] Built target duckdb_local_extension_repo
```

</details>

The regular DuckLake suite also passes.

### DIFF
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -186,7 +186,7 @@ public:
 	virtual void Commit();
 	virtual void Rollback();
 
-	//! Returns true if `message` indicates a retryable conflict (PK/unique/conflict/concurrent).
+	//! Returns true if `message` indicates a retryable conflict.
 	static bool RetryOnError(const string &message);
 
 	DuckLakeCatalog &GetCatalog() {
--- a/src/include/storage/ducklake_transaction_state.hpp
+++ b/src/include/storage/ducklake_transaction_state.hpp
@@ -30,6 +30,10 @@ struct DuckLakeCommitContext {
 	std::function<DuckLakeSnapshot()> get_snapshot;
 	//! Executes the batched snapshot/changes SQL against the metadata DB.
 	std::function<unique_ptr<QueryResult>(DuckLakeSnapshot, string &)> execute_commit_batch;
+	//! Classifies metadata-catalog errors that are safe to retry.
+	std::function<bool(const string &)> is_retryable_metadata_error = [](const string &) {
+		return false;
+	};
 	//! Optional hooks below default to a no-op/constant; callers override only the ones they need.
 	//! Clears the metadata manager cache if a clear was pending.
 	std::function<void()> flush_cache_if_pending = []() {
@@ -100,6 +104,9 @@ struct DuckLakeCommitContext {
 	std::function<void(idx_t)> set_committed_snapshot_id;
 	//! Invalidates the cached stats entry for a table after a stats-affecting file drop.
 	std::function<void(idx_t, TableIndex)> invalidate_table_stats_cache = [](idx_t, TableIndex) {
+	};
+	//! Reports a failure after the metadata commit is already durable.
+	std::function<void(const string &)> report_post_commit_error = [](const string &) {
 	};
 	//! Author / message / extra info for the snapshot row.
 	DuckLakeSnapshotCommit commit_info;
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1362,6 +1362,10 @@ bool DuckLakeTransaction::RetryOnError(const string &original_message) {
 	if (StringUtil::Contains(message, "concurrent")) {
 		return true;
 	}
+	// SQLite can skip its busy handler when waiting would deadlock.
+	if (StringUtil::Contains(message, "database is locked")) {
+		return true;
+	}
 	return false;
 }
 
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1362,10 +1362,6 @@ bool DuckLakeTransaction::RetryOnError(const string &original_message) {
 	if (StringUtil::Contains(message, "concurrent")) {
 		return true;
 	}
-	// SQLite can skip its busy handler when waiting would deadlock.
-	if (StringUtil::Contains(message, "database is locked")) {
-		return true;
-	}
 	return false;
 }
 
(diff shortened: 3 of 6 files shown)


pdet:
Hi @1fanwang, thanks for the PR! Could you run the formatter? Also, I imagine we might be able to remove the skip the tests in this case. https://github.com/duckdb/ducklake/blob/main/test/configs/sqlite.json#L41

PR #1409, opened 2026-08-24
Inline table cleanup sqlite

jprafael:
This PR allows for inline data tables that already have been flushed to be dropped, even if they are on the latest schema version. This prevents (empty) inline table accumulation for tables that don't suffer schema changes.

### DIFF
--- a/src/include/storage/ducklake_metadata_manager.hpp
+++ b/src/include/storage/ducklake_metadata_manager.hpp
@@ -347,6 +347,7 @@ public:
 	DuckLakeInlinedColNames InlinedColNames() const;
 	static string InlinedTableRegistrationTuple(idx_t table_id, const string &table_name, idx_t schema_version);
 	static string LatestInlinedTableQuery(idx_t table_id);
+	static string InlinedDataTableCleanupTargetsSql(bool superseded_only);
 	static string DropDataFiles(const set<DataFileIndex> &dropped_files);
 	static string DropDeleteFiles(const set<DataFileIndex> &dropped_files);
 	//! Caller supplies one resolved path per overwritten file, in the same order.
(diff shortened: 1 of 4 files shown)


pdet:
Hi @jprafael, we keep empty inlined tables (if not superseded) in the catalog by design, as dropping them can cause more contention.

I'm wondering if there is a specific reason why you would want this change.

jprafael:
I run a system that without this patch accumulated ~7k empty inline tables.

Querying that underlying table was noticeably slower (e.g. 30 seconds extra). I suspect it the query planner (and plan execution?) had to walk over those tables without performing any actual work. Flushing inline tables and dropping expired data files had no effect. 

I manually patched the system by dropping the actually empty inline data tables (and their metadata) which recovered the performance.

IMO the inline data tables should work similar to small data files: they are good for fast writes, but after collecting a few of them and running maintenance functions I should be able to recover the full read performance by coalescing them into the main parquet files. In this view, dropping inlined data tables on `ducklake_cleanup_old_files` makes sense, or perhaps a dedicated `ducklake_cleanup_empty_inline_data_tables`.

pdet:
To clarify, the intent of the inline cleanup is that superseded tables get deleted (i.e., inlined tables that can't be inserted anymore in the future, for example, due to a schema change), but live tables are kept even if they are empty, as we don't want to keep dropping and recreating the same table to avoid extra contention.

In summary, if you have a test that fails to delete superseded tables, that's a bug (it's unclear to me if you have that from a quick look), otherwise, it's a desired functionality, and we don't want to drop live tables. (

jprafael:
That makes sense. I think I was observing ducklake accumulating more than one inline table per table even after  `CHECKPOINT`. I guess the reason is that the system was on version 1.5.3 and the changes to cleanup older inline tables were only released on 1.5.4 ([by this commit](https://github.com/duckdb/ducklake/commit/6b1a1123bf903a820e6207dbaada968a9ffebda8)). If this is the case then  the underlying issue motivating this PR is already non-existent in 1.5.5+ and this can be ignored

pdet:
Thanks, if you stumble upon an issue here, feel free to open a PR!

PR #1410, opened 2026-08-24
Rewrite deletes candidate filter

jprafael:
This is an optimization for the REWRITE_DELETES compation on ducklakes with many unrelated tables/files.
With the patch the test case runs in a few ms, without it doesn't finish in 30mins.

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -2427,28 +2427,24 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
 		string inlined_candidates;
 		if (!inlined_deletion_table.empty()) {
 			inlined_candidates = StringUtil::Format(
-				R"(
+			    R"(
 					UNION
 					SELECT file_id AS data_file_id
 					FROM {METADATA_CATALOG}.%s
 					WHERE begin_snapshot <= %llu
 				)",
-				SQLIdentifier(inlined_deletion_table),
-				snapshot.snapshot_id
-			);
+			    SQLIdentifier(inlined_deletion_table), snapshot.snapshot_id);
 		}
 
 		candidate_files_cte = StringUtil::Format(
-			R"(
+		    R"(
 				candidate_files AS (
 					SELECT data_file_id
 					FROM {METADATA_CATALOG}.ducklake_delete_file
 					WHERE table_id=%d AND end_snapshot IS NULL%s
 				),
 			)",
-			table_id.index,
-			inlined_candidates
-		);
+		    table_id.index, inlined_candidates);
 
 		data_file_source = R"(candidate_files candidates
 JOIN {METADATA_CATALOG}.ducklake_data_file data USING (data_file_id))";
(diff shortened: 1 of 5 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2416,46 +2418,80 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
 			                       "snapshot_time::TIMESTAMPTZ >= '%s')",
 			                       timestamp_filter);
 		}
+	} else if (type == CompactionType::REWRITE_DELETES) {
+		// REWRITE_DELETES only needs files with an active delete file or visible inlined deletions. Load the
+		// inlined row IDs up front so they can both drive candidate selection and be attached to the selected files.
```
pdet: nit: lets avoid multi-line comments

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2416,46 +2418,80 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
...
+		string inlined_deletion_table = GetInlinedDeletionTableName(table_id, snapshot);
```
pdet: I think `GetInlinedDeletionTableName` is being called in `ReadInlinedFileDeletions` internally

pdet: Should we do sth like:
```cpp
else if (type == CompactionType::REWRITE_DELETES) {
      inlined_deletions = ReadInlinedFileDeletions(table_id, snapshot);

      string inlined_candidates;
      if (!inlined_deletions.empty()) {
          inlined_candidates = StringUtil::Format(
              R"(
              UNION
              SELECT file_id AS data_file_id
              FROM {METADATA_CATALOG}.%s
              WHERE begin_snapshot <= %llu
          )",
              InlinedFileDeletionTableName(table_id), snapshot.snapshot_id);
      }
```

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2416,46 +2418,80 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
...
+		    R"(
+				candidate_files AS (
+					SELECT data_file_id
+					FROM {METADATA_CATALOG}.ducklake_delete_file
+					WHERE table_id=%d AND end_snapshot IS NULL%s
+				),
+			)",
+		    table_id.index, inlined_candidates);
+
+		data_file_source = R"(candidate_files candidates
```
pdet: Is candidates being used?

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -2416,46 +2418,80 @@ vector<DuckLakeCompactionFileEntry> DuckLakeMetadataManager::GetFilesForCompacti
...
+JOIN {METADATA_CATALOG}.ducklake_data_file data USING (data_file_id))";
+		delete_file_filter = " AND end_snapshot IS NULL";
+		partition_value_filter = "\n\tWHERE data_file_id IN (SELECT data_file_id FROM candidate_files)";
+		file_filter_clause = " AND data.end_snapshot IS NULL";
```
pdet: It seems we now use it in both branches, maybe we can merge them:

```cpp
  string file_filter_clause = " AND data.end_snapshot IS NULL";
  if (type == CompactionType::MERGE_ADJACENT_TABLES) {
      if (options.min_file_size.IsValid()) { file_filter_clause += ...; }
      file_filter_clause += StringUtil::Format(" AND data.file_size_bytes < %llu", ...);
      ...
  } else if (type == CompactionType::REWRITE_DELETES) {
      ...   // no file_filter_clause line here
  }
```

pdet (commented):
Hi @jprafael thanks for the PR, is looking good, just have some minor comments.

pdet:
Thanks!

PR #1416, opened 2026-08-27
Push file-list metadata queries to PostgreSQL

bill-ph:
Follow-up to #1312.

This follows the metadata-manager convention established in #514: GetFilesForTable supplies structured file-list inputs, while each metadata manager owns backend-specific SQL generation. The default manager retains the existing attached-catalog query; PostgreSQL generates the complete operation as one native statement executed through postgres_query.

The existing CTE requirement assembler remains shared for naming, projected statistics, and materialization; PostgreSQL supplies only the native file-column-stats CTE body. This keeps data-file, delete-file, selective column-stat CTE, and bucket-partition pruning inside PostgreSQL. It also follows #622's goal of avoiding unnecessary metadata round trips, without adding a generic query passthrough.

Includes regression coverage for snapshots, deletes, static and Top-N pruning, shared stats CTEs, bucket pruning, and missing/NULL stats.


src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -4,9 +4,167 @@
...
+	}
+	if (!RequiresValueComparison(type) && type.id() != LogicalTypeId::VARCHAR) {
+		return string();
+	}
+	if (IsPostgresTemporalStatsType(type) && !CanCastPostgresTemporalValue(value, type)) {
+		return string();
+	}
+	if (type.IsNumeric()) {
+		return value.ToString();
+	}
```
pdet, line 102: Does this work for float filters? Can you add a test?

src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -143,6 +301,103 @@ string PostgresMetadataManager::GenerateFileColumnStatsCTEBody(const CTERequirem
...
+		)",
+	                                   select_list, stats_join_list, table_id.index, table_id.index);
+	if (!filter_result.where_conditions.empty()) {
+		remote_query += "\nAND " + filter_result.where_conditions;
+	}
+	remote_query += order_by_clause;
+
+	return StringUtil::Format("SELECT * FROM postgres_query({METADATA_CATALOG_NAME_LITERAL}, %s)",
+	                          SQLString(remote_query));
+}
```
pdet, line 377: This looks similar to src/storage/ducklake_metadata_manager.cpp:1877-1969, maybe we can reuse?

src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -4,9 +4,167 @@
...
+	if (CanCastPostgresStatsForValueComparison(type)) {
+		return stats + "::" + GetPostgresStatsType(type);
+	}
+	if (type.id() == LogicalTypeId::VARCHAR) {
+		return WithPostgresBinaryCollation(stats);
+	}
+	return string();
+}
+
+static string GeneratePostgresNativeFileColumnStatsCTEBody(const CTERequirement &requirement, TableIndex table_id) {
```
pdet, line 162: `GeneratePostgresNativeFileColumnStatsCTEBody` and `PostgresMetadataManager::GenerateFileColumnStatsCTEBody` emit the same body?

src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -4,9 +4,167 @@
...
 #include "storage/ducklake_transaction.hpp"
 #include "storage/ducklake_metadata_info.hpp"
+#include "storage/ducklake_table_entry.hpp"
 
 namespace duckdb {
 
+static bool IsDigit(char c) {
+	return c >= '0' && c <= '9';
+}
+
```
pdet: `StringUtil::CharacterIsDigit`

src/metadata_manager/postgres_metadata_manager.cpp
```diff
@@ -4,9 +4,167 @@
...
+static bool CanCastPostgresTemporalValue(const Value &value, const LogicalType &type) {
+	auto string_value = value.ToString();
+	if (!HasFourDigitDatePrefix(string_value)) {
+		return false;
+	}
+	return type.id() != LogicalTypeId::DATE || string_value.size() == 10;
+}
+
+static string PostgresCastValueToTarget(const Value &value, const LogicalType &type) {
+	if (value.IsNull() || value.ToString().find('\0') != string::npos || type.id() == LogicalTypeId::BLOB) {
```
pdet, line 85: can `type.id() == LogicalTypeId::BLOB` hit here?

src/include/storage/ducklake_metadata_manager.hpp
```diff
@@ -517,30 +525,50 @@ class DuckLakeMetadataManager {
 	DuckLakeFileData ReadDeleteFile(DuckLakeTableEntry &table, T &row, idx_t &col_idx, bool is_encrypted);
 
 	bool IsEncrypted() const;
+
+protected:
+	struct ColumnStatsFilterSQL {
+		std::function<string(const Value &, const LogicalType &)> cast_value;
+		std::function<string(const string &, const LogicalType &, bool)> cast_stats;
+	};
+
```
pdet: why can't we just override the base class virtual point?

pdet (commented):
Hi @bill-ph thanks for the PR, I left a few comments.

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: added [test](https://github.com/duckdb/ducklake/pull/1416/changes/76b185ad5acd36e69e039c888b1fb80b5886a9e0#diff-6aa5fb37c1f877f7f632ddd65c14938239b679e805ddf2399e84336ac350a621R158)

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: This is a tricky decision, I see 3 options:
1, as in this PR, duplicate in postgres manager, minimal impact/exposure to parent and other sibling manager implementations.
2, abstract and reuse GenerateFileListQuery, max code reuse, but exposes complexity to sibling manager implementations unnecessarily.
3, a second, non-virtual overload with additional SQL-generation parameters, called from old GenerateFileListQuery and only used by base and postgres manager, middle-ground, but feels awkward.

I made implementation for [option 2](https://github.com/bill-ph/ducklake/pull/2/), let me know if this is preferred.

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: BLOB cannot reach PostgresCastValueToTarget, this is just defensive in case call path changes causing blob to reach here, and when that happens, only metadata pruning is omitted, DuckDB filter still runs and preserves correctness.

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: good pointer, fixed.

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: good eye, reused

src/include/storage/ducklake_metadata_manager.hpp
bill-ph: Because a PostgresMetadataManager currently emits SQL for two different parsers: GetFilesForTable (using postgres) and GetExtendedFilesForTable(using duckdb), override the base class would require migrating GetExtendedFilesForTable to postgres parser, it is the cleaner long term implementation, but would make this PR larger / a small scope creep.

Here is [a potential follow-up PR](https://github.com/bill-ph/ducklake/pull/4) that cleans this up.

bill-ph:
Thanks for the review @pdet ! made updates based on your suggestions, left 2 review comments unresolved with reply, let me know what you think.

src/metadata_manager/postgres_metadata_manager.cpp
pdet: Yes, lets go with that

src/include/storage/ducklake_metadata_manager.hpp
pdet: Yes, lets go with that

bill-ph: done, merged into this PR.

src/metadata_manager/postgres_metadata_manager.cpp
bill-ph: done, merged option 2 into this PR.

bill-ph:
all comments addressed

pdet:
Hi @bill-ph looks good to me, thanks for the work, do you mind fixing the conflict?

bill-ph:
Resolved conflicts with #1428 by preserving prefix-range and Top-N stats through PostgreSQL file-list pushdown and adding unique output aliases.

The failed CI seems unrelated flakiness, can you try rerun it?

Alex-Monahan:
A small commit should retrigger CI for you! Thanks for this work!

pdet:
Thanks!

PR #1417, opened 2026-08-27
Backport of #1397 (pdet/retype_stats)

ThijsDreef:
A backport of #1397 for 1.5, two changes in main make this a non clean cherry-pick.

DuckDB on main currently prints stats using variant (different notation then on 1.5), so tests are changed to expect the old stats notation.
DuckLake on main has support for negative caching of empty stats, not supported on 1.5 ducklake some changes are pulled in by the cherry-pick which are removed to not introduce new behaviour.


ThijsDreef:
I excluded the reseed test for the quack catalog as it failed due to an existing skip reason. 
Let me know if that works for the backport or if I missed something.

PR #1418, opened 2026-08-27
Move file-list query generation to metadata managers

bill-ph:
Preparatory refactor for #1416.

Move the SQL generation used by `GetFilesForTable` behind a structured metadata-manager operation. The default manager generates the existing attached-catalog query, while execution and result parsing remain shared.

This preserves DuckDB, SQLite, and PostgreSQL behavior and provides the backend-owned query-generation seam used by the stacked PostgreSQL optimization.

Tested the focused filter, Top-N, bucket-pruning, and delete suites under the default, SQLite, and PostgreSQL configurations.


bill-ph:
Superseded by the focused single-PR implementation in #1416.

PR #1420, opened 2026-08-28
Retry SQLite catalog commits when the database is locked

### LINKED ISSUE #1309: SQLite catalog COMMIT does not retry SQLITE_BUSY / database is locked

1fanwang:
A DuckLake write using a SQLite catalog can fail during `COMMIT` with `database is locked`, even with a long SQLite busy timeout and DuckLake retries enabled. This happens when the transaction has read the catalog and another connection becomes the pending writer before the DuckLake commit upgrades to a write lock.

SQLite can skip its busy handler for this read-to-write deadlock. The SQLite metadata manager classifies that lock error, while the bounded commit loop retries only when rollback is safe and excludes failed `COMMIT` wrappers. After the metadata commit becomes durable, only inlined-table cleanup is allowed to fail without changing the commit result.

Fixes https://github.com/duckdb/ducklake/issues/1309

## Testing

The regression uses two real DuckDB connections against one SQLite catalog. A zero-retry control requires the exact lock error, then the retry path commits the staged row once. An unrelated metadata error remains non-retryable.

The SQLite suite runs the lock-conflict cases except three stress tests that set 100 retries and can hold catalog locks across loop connections.

<details>
<summary>Raw before/after output</summary>

```console
$ SOURCE=src/storage/ducklake_transaction.cpp
$ TEST=test/sql/issues/issue_1309.test
$ TESTER=build/release/test/unittest
$ RUNNER=duckdb/scripts/ci/run_tests.py
$ git restore --source=origin/main --worktree "$SOURCE"
$ cmake --build build/release --config Release -j8
$ "$TESTER" "$TEST"
Query unexpectedly failed! (test/sql/issues/issue_1309.test:71)!
COMMIT;
Failed to flush changes into DuckLake: database is locked
test cases: 1 | 0 passed | 1 failed

$ git restore --source=HEAD --worktree "$SOURCE"
$ cmake --build build/release --config Release -j8
$ "$TESTER" "$TEST"
All tests passed (34 assertions in 1 test case)

$ python "$RUNNER" --stabilize-tests --workers 1 --batch-size 1 "$TESTER" "$TEST"
stabilization rerun 9/9 for fast tests
all tests passed in 9s

$ "$TESTER" --test-config test/configs/sqlite.json --test-dir ./ "$TEST"
All tests passed (34 assertions in 1 test case)

$ "$TESTER" --test-config test/configs/sqlite.json
All tests passed (32 skipped tests, 17583 assertions in 483 test cases)

$ jq '.skip_tests += [{"reason":"isolated after full-suite lock","paths":["test/sql/concurrent/concurrent_insert_delete_conflict.test"]}]' \
    test/configs/sqlite.json > /tmp/sqlite-with-known-flake-isolated.json
$ "$TESTER" --test-config /tmp/sqlite-with-known-flake-isolated.json
All tests passed (33 skipped tests, 20348 assertions in 539 test cases)

$ "$TESTER" test/sql/concurrent/concurrent_insert_delete_conflict.test
All tests passed (13 assertions in 1 test case)
```

</details>

### DIFF
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -186,7 +186,7 @@ public:
 	virtual void Commit();
 	virtual void Rollback();
 
-	//! Returns true if `message` indicates a retryable conflict (PK/unique/conflict/concurrent).
+	//! Returns true if `message` indicates a retryable conflict.
 	static bool RetryOnError(const string &message);
 
 	DuckLakeCatalog &GetCatalog() {
--- a/src/include/storage/ducklake_transaction.hpp
+++ b/src/include/storage/ducklake_transaction.hpp
@@ -186,7 +186,7 @@ public:
 	virtual void Commit();
 	virtual void Rollback();
 
-	//! Returns true if `message` indicates a retryable conflict.
+	//! Returns true if `message` indicates a retryable conflict (PK/unique/conflict/concurrent).
 	static bool RetryOnError(const string &message);
 
 	DuckLakeCatalog &GetCatalog() {
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1362,6 +1362,10 @@ bool DuckLakeTransaction::RetryOnError(const string &original_message) {
 	if (StringUtil::Contains(message, "concurrent")) {
 		return true;
 	}
+	// SQLite can skip its busy handler when waiting would deadlock.
+	if (StringUtil::Contains(message, "database is locked")) {
+		return true;
+	}
 	return false;
 }
 
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1446,14 +1446,7 @@ void DuckLakeTransaction::RunCommitLoop(DuckLakeSnapshot transaction_snapshot,
 		return metadata_manager->Execute(snapshot, query);
 	};
 	context.is_retryable_metadata_error = [&](const string &message) {
-		auto &metadata_type = ducklake_catalog.MetadataType();
-		if (metadata_type != "sqlite" && metadata_type != "sqlite_scanner") {
-			return false;
-		}
-		auto lower_message = StringUtil::Lower(message);
-		StringUtil::Trim(lower_message);
-		return StringUtil::EndsWith(lower_message, "database is locked") &&
-		       !StringUtil::Contains(lower_message, "failed to execute query \"commit\": database is locked");
+		return metadata_manager->IsRetryableCommitError(message);
 	};
 	context.flush_cache_if_pending = [&]() {
 		if (metadata_manager->TakePendingCacheClear()) {
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -1362,10 +1362,6 @@ bool DuckLakeTransaction::RetryOnError(const string &original_message) {
 	if (StringUtil::Contains(message, "concurrent")) {
 		return true;
 	}
-	// SQLite can skip its busy handler when waiting would deadlock.
-	if (StringUtil::Contains(message, "database is locked")) {
-		return true;
-	}
 	return false;
 }
 
@@ -1448,6 +1444,16 @@ void DuckLakeTransaction::RunCommitLoop(DuckLakeSnapshot transaction_snapshot,
 	};
 	context.execute_commit_batch = [&](DuckLakeSnapshot snapshot, string &query) {
 		return metadata_manager->Execute(snapshot, query);
+	};
+	context.is_retryable_metadata_error = [&](const string &message) {
+		auto &metadata_type = ducklake_catalog.MetadataType();
+		if (metadata_type != "sqlite" && metadata_type != "sqlite_scanner") {
+			return false;
+		}
+		auto lower_message = StringUtil::Lower(message);
+		StringUtil::Trim(lower_message);
+		return StringUtil::EndsWith(lower_message, "database is locked") &&
+		       !StringUtil::Contains(lower_message, "failed to execute query \"commit\": database is locked");
 	};
 	context.flush_cache_if_pending = [&]() {
 		if (metadata_manager->TakePendingCacheClear()) {
@@ -1554,6 +1560,9 @@ void DuckLakeTransaction::RunCommitLoop(DuckLakeSnapshot transaction_snapshot,
 	};
 	context.invalidate_table_stats_cache = [&](idx_t next_file_id, TableIndex table_id) {
 		ducklake_catalog.InvalidateTableStatsCache(next_file_id, table_id);
+	};
+	context.report_post_commit_error = [&](const string &message) {
+		DUCKDB_LOG_WARNING(db, StringUtil::Format("DuckLake post-commit cleanup failed: %s", message));
 	};
 	context.commit_info = state->commit_info;
 	context.supports_v1_1_metadata = ducklake_catalog.SupportsV1_1Metadata();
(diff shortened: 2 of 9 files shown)


src/storage/ducklake_transaction.cpp
```diff
@@ -1445,6 +1445,16 @@ void DuckLakeTransaction::RunCommitLoop(DuckLakeSnapshot transaction_snapshot,
 	context.execute_commit_batch = [&](DuckLakeSnapshot snapshot, string &query) {
 		return metadata_manager->Execute(snapshot, query);
 	};
+	context.is_retryable_metadata_error = [&](const string &message) {
+		auto &metadata_type = ducklake_catalog.MetadataType();
+		if (metadata_type != "sqlite" && metadata_type != "sqlite_scanner") {
+			return false;
+		}
```
pdet: Can we move this to a virtual function in ducklake_metadata_manager.hpp and have it specialized for sqlite

src/storage/ducklake_transaction_state.cpp
```diff
@@ -2007,6 +2008,21 @@ void DuckLakeTransactionState::Commit(DuckLakeSnapshot transaction_snapshot,
 	}
 	// If we got here, this snapshot was successful
 	context.set_committed_snapshot_id(commit_snapshot.snapshot_id);
+	try {
+		for (auto &entry : dropped_file_stats) {
+			context.invalidate_table_stats_cache(commit_snapshot.next_file_id, entry.first);
+		}
+		if (flushed_inlined && !context.skip_drop_empty_inlined) {
+			DropEmptySupersededInlinedTables(context);
```
pdet, line 2008: I think this is the only step here that can actually throw, which would cause the remainder of the cleanup not to run, maybe we can move the other steps outside the try?

src/storage/ducklake_transaction_state.cpp
```diff
@@ -1949,34 +1951,33 @@ void DuckLakeTransactionState::Commit(DuckLakeSnapshot transaction_snapshot,
...
+			bool should_abort = finished_retrying;
+			if (!retry_on_error) {
+				should_abort = true;
+			}
+			if (!can_retry) {
+				if (!retryable_metadata_error) {
+					should_abort = true;
+				}
+			}
+			if (should_abort) {
```
pdet: Since retryable_metadata_error already implies retry_on_error, can this be simplified?

e.g.,

```cpp
bool retry_on_error = retryable_metadata_error || (can_retry && DuckLakeTransaction::RetryOnError(error.Message()));
if (!retry_on_error || finished_retrying) {
...
```

src/include/storage/ducklake_transaction.hpp
```diff
@@ -186,7 +186,7 @@ class DuckLakeTransaction : public Transaction, public enable_shared_from_this<D
 	virtual void Commit();
 	virtual void Rollback();
 
-	//! Returns true if `message` indicates a retryable conflict (PK/unique/conflict/concurrent).
+	//! Returns true if `message` indicates a retryable conflict.
```
pdet: Why we removed this?

pdet (commented):
Thanks for the PR, I had a few comments.

src/storage/ducklake_transaction.cpp
1fanwang: Done in 0d219ba9.

src/storage/ducklake_transaction_state.cpp
1fanwang: Done in 0d219ba9.

src/storage/ducklake_transaction_state.cpp
1fanwang: Done in 0d219ba9.

src/include/storage/ducklake_transaction.hpp
1fanwang: Done in 0d219ba9.

pdet:
Thanks!

PR #1421, opened 2026-08-28
Cross column pushdown

J-Meyers:
Continuation of breaking up #477, following #471, #513 and #514. This is the case #471 flagged as still missing:

```sql
SELECT * FROM partitoned_table
WHERE (partition_col_0 = 0 AND partition_col_1 = 1) OR partition_col_1 = 2
```

`GenerateTableScanFilters` returns a `TableFilterSet`, which is one filter per column with all of them ANDed. Projecting a disjunction onto that loses information twice over: a column that is not constrained in *every* branch cannot be constrained at all, and even when it is, the correlation between columns is dropped. Since we are talking to a database rather than reading a file footer, we do not have to flatten - `data_file_id IN (SELECT ... FROM col_N_stats WHERE ...)` is already a per-file boolean, so an AND/OR tree over those is just SQL.

`FilterPushdownInfo` gains a list of filter trees alongside `column_filters` rather than replacing it, since the Top-N stats join, `CanSkipFileByTopNDynamicFilter` and bucket partition pruning are only correct for that AND-spine. Each OR branch gets a fresh `FilterCombiner`, so equality propagation and contradiction detection still come from the combiner. The tree walk emits only `AND`, `OR`, parens and `1=0`, so every dialect-specific decision stays in the existing virtual leaf methods and no metadata backend needs changes.

Two things the combiner cannot express are now handled directly, since `GenerateFilterFromExpression` already renders both: null checks (`TryPushdownExpression` covers prefix, LIKE, IN, OR-clause and temporal casts, but not `IS [NOT] NULL`), and nested struct fields, where stats are stored against the leaf field but the path only exists in the expression, so these resolved to the struct parent, which has no stats rows at all.

`DynamicFilterPushdown` now merges into the filter info it inherits. The final `TableFilterSet` cannot round-trip a tree, so otherwise the tree is built during optimization and dropped again at `MultiFileInitGlobal`.

Files read on a three file table (`id` 1-100 / 101-200 / 201-300, one `category` per file):

| filter | before | after |
| --- | --- | --- |
| `id = 50 OR category = 3` | 3 | 2 |
| `(id <= 100 AND category = 3) OR (id >= 201 AND category = 1)` | 2 | 0 |
| `id = 50 OR (id = 150 AND id = 200)` | 2 | 1 |
| `v IS NULL OR id = 250` | 3 | 2 |
| `st.a = 50` | 3 | 1 |
| `st.a = 50 AND st.b = 100` | 3 | 1 |

Unfilterable branches still prune nothing, and same-column disjunctions are unchanged.

## Budgets

Every leaf becomes its own membership condition, so a wide disjunction can generate a lot of SQL - `(a = ? AND b = ?) OR ...` repeated N times. Two separate limits, because they bound different things:

- `MAX_VISITED_NODES` bounds how much of the expression we walk, since every step runs a `FilterCombiner` over it.
- `MAX_LEAVES` bounds the generated query. One visit can yield a leaf per column the branch constrains, so counting walk steps does not bound this.

When either runs out the tree is dropped rather than kept half-built, so pruning does not depend on where the walk happened to stop; the per-column filters still apply. `MAX_LEAVES` is deliberately low for now - the file-list query is ~125 ms at the limit on a 200 file table and grows superlinearly past it. That cost is almost entirely the repeated `IN (SELECT ...)` subqueries, so the limit is worth raising once those are joins.

### Notes

- `ducklake_file_variant_stats` is written but never read for pruning, so shredded VARIANT paths cannot prune at all. Separate from this.
- Overlaps #1416 only in `ConvertFilterPushdownToSQL` and the membership condition body. Whichever lands second should thread `ColumnStatsFilterSQL` through the tree walk, since `TRY_CAST` does not exist in PostgreSQL and a missing `COLLATE "C"` would quietly give wrong VARCHAR comparisons.
- Does not include support for filters like `id < category` where the leaf is a comparison between different columns

Done with AI help, cobbling together various local things and hopefully now able to get this cool feature in, much easier now that the TableFilter interface has been simplified / unified


danthegoodman1:
I just found this bug while implementing a custom metadata provider too!

danthegoodman1:
Actually, to add a data point: the `matches_null_rows change` fixes a silent wrong-answer bug on main too.

On `4d3cb0b`, a DuckDB-catalog DuckLake with one all-NULL file:

```sql
CREATE TABLE t(id INTEGER, region VARCHAR);
INSERT INTO t SELECT i, NULL   FROM range(0,100)   t(i);
INSERT INTO t SELECT i, 'emea' FROM range(100,200) t(i);
INSERT INTO t SELECT i, 'apac' FROM range(200,300) t(i);
SELECT count(*) FROM t WHERE region IS NULL OR region = 'emea';  -- 100, expected 200
SELECT count(*) FROM t WHERE region IS NULL;                     -- 100, correct
SELECT count(*) FROM t WHERE region = 'emea';                    -- 100, correct
```

Each branch alone is right, only the disjunction drops the 100 NULL rows. The all-NULL file records value_count = 0 with both bounds absent, so it fails (`value_count IS NULL OR value_count > 0)` even though null_count = 100 is what the `IS NULL` branch asks for.

Found while building a metadata backend against the `DuckLakeMetadataManager` interface. Testing the pruning rules reproduced the row loss (Claude Opus 5 found it).

TLDR; this PR fixes it, but it doesn't mention it, and i'd argue this is a more significant bug.

src/storage/ducklake_multi_file_list.cpp
```diff
@@ -104,18 +385,47 @@ unique_ptr<MultiFileList> DuckLakeMultiFileList::ComplexFilterPushdown(ClientCon
...
+		}
+		FilterTreeState state;
+		auto root = BuildFilterTree(context, info, *filter, state);
+		if (!root || state.exhausted) {
+			// too big to express as a tree - the per-column filters still apply
+			continue;
+		}
+		unordered_set<idx_t> columns;
+		GetFilterTreeColumns(*root, columns);
+		if (columns.size() < 2 && !state.removed_branch) {
```
guillesd: I think if there is a `region IS NULL OR region = 'emea'` we are throwing away the filter since the combiner returns nothing for `region IS NULL`. Can you confirm?
I think this can be fixed in `FilterTreeState` where there is already similar behaviour I believe (there is a flag called `removed_branch`)

src/common/ducklake_util.cpp
```diff
@@ -360,6 +366,116 @@ shared_ptr<DynamicFilterData> DuckLakeUtil::GetOptionalDynamicFilterData(const T
...
+optional_ptr<const Expression> DuckLakeUtil::GetFilterSubject(const Expression &expr) {
+	optional_ptr<const Expression> subject;
+	bool conflict = false;
+	FindFilterSubject(expr, subject, conflict);
+	return conflict ? nullptr : subject;
+}
+
+const Expression &DuckLakeUtil::GetFilterSubjectPath(const Expression &subject, vector<string> &path) {
+	reference<const Expression> current = subject;
+	while (IsStructExtract(current.get())) {
```
guillesd, line 439: This is case sensitive, which can lead to a filter loss. `struct_extract` is case insensitive, so it will bind ok for a user regardless of whether the struct key is `a` or `A`. I think we should use the binded info:
`path.push_back(StructType::GetChildName(input_type, position).GetIdentifierName());`

src/include/storage/ducklake_multi_file_list.hpp
```diff
@@ -63,6 +63,44 @@ class DuckLakeMultiFileList : public MultiFileList {
...
+	//! Budgets for building one filter tree, and what the walk found that per-column filters cannot express
+	struct FilterTreeState {
+		//! Bounds how much of the expression we walk - every step runs a FilterCombiner over it
+		static constexpr idx_t MAX_VISITED_NODES = 64;
+		//! Bounds what the generated query costs: one stats condition per leaf, all nested in one OR/AND
+		//! chain. Separate from the walk, since one visit can yield a leaf per column. The leaves share
+		//! the column's stats join, but planning the chain still grows superlinearly in its size.
+		static constexpr idx_t MAX_LEAVES = 64;
+
+		idx_t visited_nodes = 0;
```
guillesd, line 90: Maybe add a test for the exhaustion path

test/sql/stats/complex_filter_pushdown.test
```diff
@@ -0,0 +1,117 @@
...
+1
+
+query II
+EXPLAIN ANALYZE SELECT COUNT(*) FROM ducklake.t WHERE id = 50 OR (category = 2 AND category = 3);
+----
+analyzed_plan	<REGEX>:.*Total Files Read: 1.*
+
+# nested disjunctions
+query I
+SELECT COUNT(*) FROM ducklake.t WHERE (id < 50 OR category = 3) AND (id > 20 OR category = 2);
```
guillesd, line 64: Only checks rows, not files

guillesd (changes requested):
Looks good! Some comments. There is also a minor conflict merge conflict to solve

src/storage/ducklake_multi_file_list.cpp
J-Meyers: That specific case was handled since the filter combiner was able to combine, but some others weren't like `region IS NULL OR region LIKE 'em%'` and `region LIKE 'ap%' OR region LIKE 'em%'` these are now handled

Removed reliance on removed_branch

Additionally `NOT (x IS NULL)` and `NOT (x IS NOT NULL)` are still not handled and while they could be I haven't made that change because it seemed a bit overly specific with `NOT` not able to be generally handled

src/common/ducklake_util.cpp
J-Meyers: I think it was okay because it relied on a `case_insensitive_map_t`, but updated it to reuse the duckdb `TryGetStructExtractChildIndex`

src/include/storage/ducklake_multi_file_list.hpp
J-Meyers: Added

test/sql/stats/complex_filter_pushdown.test
J-Meyers: Added file check

J-Meyers:
Not sure about the cause of the Minio test failure, the specific test that fails doesn't seem to exercise anything in this PR

For the musl test failure I created https://github.com/duckdb/duckdb/pull/25378 in duckdb core

guillesd:
Thanks! I think both tests that failed are flaky. I'm re-running now

guillesd:
I think this is ready to merge. The musl bug seems independent. @pdet do you prefer for that to be merged in core and bump the pin?

pdet:
> I think this is ready to merge. The musl bug seems independent. @pdet do you prefer for that to be merged in core and bump the pin?

Yes, I also do that to my own PRs, first fix the issue in core, then bump then we can merge the parent.

J-Meyers:
Issues seem to be driven by stack based limits being hit with the PEG parser crashing, which musl with small stacks is extremely vulnerable to, seems to be resolved https://github.com/duckdb/duckdb/issues/24618

Also there were easy changes to make to reduce the parens that then significantly reduced the depth that the parser went through that allows raising the limit / keeping it high anyways so we don't need actually need to wait for them to make it in

pdet:
Thanks!

PR #1423, opened 2026-08-28
KMS-agnostic envelope encryption: abstract provider interface

tiagotaveiragomes:
## What this is

An abstract `DuckLakeEncryptionProvider` interface plus the ATTACH options that will drive it (`ENCRYPTION_SOCKET`, `ENCRYPTION_LAKE_ID`, `ENCRYPTION_CACHE_TTL_SECONDS`). It lets a build that links a key-management service wrap the per-file DEKs DuckLake stores in `encryption_key`, so the metadata catalog at rest never holds a usable key — while keeping DuckLake itself free of any dependency on a particular KMS.

This commit is inert on its own: nothing constructs a provider, no behavior changes, and with no factory registered the new ATTACH options are refused. The catalog wiring, fail-closed guards, key-rotation sweep, and an executable reference provider with end-to-end SQL tests are in the follow-up PR #1424 (built on this branch; the two can be reviewed together and merged as one if preferred).

## Design notes

**The factory takes `DatabaseInstance &`.** A provider needs crypto primitives (AEAD, random). Taking them from the attaching database via `GetEncryptionUtil()` means a provider inherits whatever crypto policy that database is configured with — OpenSSL when httpfs is loaded, mbedtls otherwise, and the `force_mbedtls_unsafe` escape hatch when a test asks for it. A provider that instantiated its own crypto would silently ignore that policy.

**Process-global factory slot.** The smallest mechanism that lets an integrating extension make a provider selectable at ATTACH time without DuckLake depending on it. `RegisterFactory` refuses a second registration rather than overwriting — otherwise the provider a lake gets would depend on extension load order. `HasFactory()` lets a loader that runs once per `DatabaseInstance` skip re-registration, since registration is per-process. If you'd rather hang this off `DBConfig` or an extension callback, the interface is unchanged; only `RegisterFactory`/`GetFactory`/`HasFactory` move.

**`LooksWrapped` is static, not virtual**, because the reader that needs it may have no provider at all — that is exactly the fail-closed case: a lake re-attached without envelope options must recognize "this stored key is a wrapped blob I cannot open" without a service call, and refuse rather than hand ciphertext to the Parquet reader as an AES key. The check is base64 prefix `RExL` (the first three bytes of the `DLK1` header) **and** length > 44 (the base64 length of a 32-byte key, the longest a plain lake stores), so no plaintext key can be misclassified.

**Batch-first API.** Every operation takes a batch: a scan materializes its whole file-and-key list before reading any file, so one KMS round trip serves a scan.

**`encryption_key` stays an opaque VARCHAR** — no catalog schema change.

## Verification

- Builds clean standalone on this base.
- Exercised end to end (wrap/unwrap, fail-closed re-attach, rotation sweep) by the SQL test suite in #1424 — all green, run against the built extension.
- Related: #1403 (CSPRNG for key generation) is independent of this interface.

### DIFF
--- a/src/include/storage/ducklake_encryption_provider.hpp
+++ b/src/include/storage/ducklake_encryption_provider.hpp
@@ -1,0 +1,103 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// storage/ducklake_encryption_provider.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/common.hpp"
+#include "duckdb/common/vector.hpp"
+
+#include <functional>
+
+namespace duckdb {
+class DatabaseInstance;
+
+//! What a per-file encryption key is cryptographically bound to, so a wrapped blob minted for one
+//! file cannot be used as the key for another.
+struct DuckLakeFileIdentity {
+	string lake_id;
+	int64_t table_id = 0;
+	//! Data files and delete files have independent id spaces and separate catalog tables, so the
+	//! kind of file is part of the binding.
+	bool is_delete_file = false;
+	//! The path as stored in ducklake_data_file / ducklake_delete_file, before it is resolved
+	//! against the table's data path - which is a mutable table property.
+	string stored_path;
+};
+
+//! One entry of a rewrap batch reply.
+struct DuckLakeRewrapResult {
+	//! The value the row should carry from now on. When `rewrapped` is false this is the value that
+	//! was sent, so a converged sweep writes nothing.
+	string wrapped;
+	//! Whether the provider moved this entry onto its active key.
+	bool rewrapped = false;
+};
+
+//! Envelope encryption provider: wraps the per-file DEKs DuckLake stores in `encryption_key` so the
+//! metadata catalog never holds a usable key. One instance per attached lake.
+//!
+//! Every operation takes a batch: a scan materialises its whole file-and-key list before reading any
+//! file, so one round trip serves a scan.
+//!
+//! `encryption_key` stays an opaque VARCHAR, so no catalog schema change is required.
+class DuckLakeEncryptionProvider {
+public:
+	//! Lifetime of a cached unwrapped DEK when ATTACH does not specify one.
+	static constexpr int64_t DEFAULT_CACHE_TTL_SECONDS = 300;
+	//! Ceiling for a configured cache TTL, so it cannot be raised to an effectively unbounded one.
+	static constexpr int64_t MAX_CACHE_TTL_SECONDS = 3600;
+
+	virtual ~DuckLakeEncryptionProvider() = default;
+
+	//! Wrap one commit's worth of keys. `deks` are raw key bytes; the returned strings are the
+	//! values the `encryption_key` column will carry.
+	virtual vector<string> WrapKeys(const vector<DuckLakeFileIdentity> &identities, const vector<string> &deks) = 0;
+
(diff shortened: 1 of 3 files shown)


tiagotaveiragomes:
Some concrete detail on what's already built against this interface in our fork, so the shape of the eventual follow-up is clear rather than an unspecified "smaller PR":

- **Catalog wiring** (`ducklake_catalog.cpp`/`.hpp`): ~12 touch points — ATTACH-time provider construction from the three options in this PR, and threading the constructed provider through to the read/write paths that need it.
- **Fail-closed refusal guards** (`ducklake_envelope_guards.cpp`, 210 lines): the paths that must hard-refuse rather than silently proceed — attaching a wrapped lake with no provider configured, a provider that fails `SelfTest()`, a rewrap whose identity doesn't match the stored file.
- **Key-rotation table function** (`ducklake_rewrap_keys.cpp`, 407 lines): a SQL-callable function that walks a lake's files and re-wraps their DEKs under a new provider/key version, using `RewrapKeys` from the interface in this PR.
- **Boot-time self-test function** (`ducklake_self_test.cpp`, 153 lines): calls `SelfTest()` at ATTACH time so a misconfigured provider fails fast instead of surfacing as a mysterious read/write error later. This one's otherwise standalone and provider-agnostic, but it calls a catalog accessor (`DuckLakeCatalog::EncryptionProvider()`) that only exists once the catalog wiring above lands — so it has to travel with that PR, not ahead of it.

None of this is provider-specific — no KMS wire protocol, no vendor SDK — it's all catalog/interface plumbing. Total follow-up is roughly 770 lines across the three functional pieces plus the catalog touch points, which is bigger than "small" — flagging that now rather than understating it. Happy to prepare it as a draft PR once there's a read on whether the interface shape here is the right direction, so it's not built twice.

tiagotaveiragomes:
Adopted this interface in a downstream fork that actually registers a factory. One finding on `RegisterFactory`, from running it rather than reading it.

**The refusal is right; the interface gives an integrator no way to satisfy it.**

`RegisterFactory` now throws when a factory is already registered. Registration is per-PROCESS. But the natural place for an integrator to call it is extension load, and `LoadInternal` runs once per `DatabaseInstance` - so the second database opened in one process throws at load time:

```
extension 'ducklake' load threw an exception: Invalid Input Error: a DuckLake
encryption provider factory is already registered - registering a second one
would make the provider a lake gets depend on extension load order
```

Measured: 10 of our 11 provider fixtures went red at `require ducklake`, because the sqllogictest runner builds one `DatabaseInstance` per `.test` file. The fixtures only made it cheap to see; the same throw reaches any embedding host that opens two databases.

Nothing in tree hits this today - 594098c1 has no caller of `RegisterFactory` anywhere - so this is a defect that ships dormant and fires on the first integrator.

The only thing an integrator can do with the interface as published is keep a caller-side `static bool registered` flag, which is a worse version of state the provider module already owns. Suggest adding the accessor:

```cpp
//! Whether a factory is already registered.
static bool HasFactory();
```

so the loader can ask before it registers. That does not weaken the refusal: two DIFFERENT providers still each call `RegisterFactory` from their own translation unit and the second still throws. What the guard skips is the same build re-registering the same provider, which is not a load-order ambiguity at all.

Alternative, if you would rather not grow the surface: make `RegisterFactory` idempotent for an identical registration and keep the throw only for a genuinely different one. That needs provider identity in the signature, which is why we did not take it.

tiagotaveiragomes:
Both fixed in the updated branch (4fcd68b1): `HasFactory()` added next to `RegisterFactory`, and the reference provider's loader in the follow-up now uses it as the intended once-per-process guard.

PR #1424, opened 2026-08-29
Catalog wiring for the KMS-agnostic encryption provider (#1423 follow-up)

tiagotaveiragomes:
## What this is

The working half of #1423: the `DuckLakeEncryptionProvider` interface wired into the catalog, commit, and read paths, with fail-closed guards, a `ducklake_rewrap_keys` rotation sweep, a `ducklake_self_test` health check — and a reference provider (test fixture) that makes the whole path executable end to end in SQL, with no external KMS, wire protocol, or network. Two commits: the wiring, then the fixture + tests. Built on #1423's branch; happy to merge the two PRs as one if that reviews easier.

## How it wires — two symmetric choke points

**Write.** `DuckLakeCatalog::PrepareFileKeysForCommit` turns each raw DEK into the value the `encryption_key` column will carry — a wrapped blob when a provider is configured, plain base64 otherwise. It runs once per commit via the `DuckLakeCommitContext::prepare_file_keys` hook, after stored paths are resolved and before either writer (Appender fast path or SQL batch) emits a row.

**Read.** `DuckLakeCatalog::ResolveStoredEncryptionKey` turns the stored value back into raw key bytes, through the provider when one is configured. Every table-scan read routes through it via `ReadDataFile`.

**Encode symmetry.** The invariant is: the value handed to a writer is already in its stored form, so a writer only quotes it. The new `StoredEncryptionKeyLiteral` quotes without encoding; the Appender fast path stopped base64-encoding; on a lake with no provider the default hook does exactly the `Blob::ToBase64` the writers used to do — emitted SQL and appended rows are byte-identical to before, which is what makes this a no-op for existing lakes.

**Server-side commits are routed, not wrapped.** A wrapped key is bound to the *stored* path (AAD), which on the server-side commit path only the server knows. So an enveloped lake takes the existing client-side commit fallback instead — a route change, not a refusal; unencrypted lakes are untouched.

## What the key is bound to

`DuckLakeFileIdentity` = `{lake_id, table_id, is_delete_file, stored_path}`. A blob minted for one file is useless as the key for another: swap the table, the path, or a data file for a delete file, and the unwrap fails rather than returning something. `stored_path` (the `path` column value, not the resolved path) keeps keys stable under data-path moves; `is_delete_file` is needed because data and delete files have independent id spaces.

## Fail-closed guards

Each refusal covers a specific way this could silently degrade into "encrypted, but not really":

- Re-attach without a provider and a wrapped key → refused (never hands ciphertext to the Parquet reader as a key).
- Missing/empty stored key on an encrypted lake → refused.
- Decoded key of a length `ParquetCrypto` won't accept → refused as corrupt, plainly.
- Temp spill: ATTACH turns `temp_file_encryption` on and refuses if it can't; read and commit paths re-check, since the setting can be flipped after ATTACH. Note: `temp_file_encryption` is a database-wide setting, so attaching an enveloped lake enables it for the process.
- Data inlining is forced off (inlined rows and their min/max would land in the catalog as cleartext literals, covered by no file key): `DataInliningRowLimit` returns 0, ATTACH refuses an explicit non-zero limit, and `ducklake_set_option` refuses to raise it later.

Pre-existing behavior worth knowing: `ducklake_list_files` reconstructs live DEKs for a session attached with the provider, exactly as it always has for plain `ENCRYPTED` lakes. The envelope's claim is about the catalog at rest, which now holds only wrapped blobs.

## `ducklake_rewrap_keys`

The consumer half of key rotation: moves every stored blob onto the provider's current key so the outgoing key can be retired. Runs on its own autocommit connection (an interrupted sweep leaves a committed prefix), keyset pagination, compare-and-swap updates (concurrent writers are left for the next pass, never clobbered), refuses at bind on a lake with no provider, throws on an unwrapped row rather than silently stranding it, sweeps delete files too, and reports no key material.

## The reference provider is a test fixture, not a KMS

Stated in its class doc so it cannot be mistaken for production-grade:

- KEKs derive from a seed published in the source file.
- The tests run with `force_mbedtls_unsafe = true`, so GCM nonces come from a non-cryptographic PRNG — fine for a fixture, disqualifying for anything else.
- It registers only when `DUCKLAKE_TEST_REFERENCE_ENCRYPTION_PROVIDER=1` is set; without it no factory exists and runtime behavior is identical to a build without the file.

What it does exercise is real: AES-256-GCM through the database's `GetEncryptionUtil()`, AAD binding with length-prefixed (injective) identity serialization, and a genuine key-version rotation driven through `ducklake_rewrap_keys`. Its `SelfTest` proves the AAD binding is load-bearing: round-trip, then assert that unwrapping the same blob under a different identity is refused.

## Tests (executed, green)

`DUCKLAKE_TEST_REFERENCE_ENCRYPTION_PROVIDER=1 ./build/release/test/unittest "test/sql/encryption_provider/*"`

```
All tests passed (72 assertions in 5 test cases)
```

- `round_trip.test` — write, read back, assert the catalog holds a **wrapped** key (envelope prefix and longer than any raw key's base64) for both data and delete files, assert the Parquet files are really encrypted, assert a fresh ATTACH still decrypts.
- `fail_closed_no_provider.test` — re-attach without envelope options refuses the read; a fresh lake with envelope options but no ENCRYPTED is refused; raising `data_inlining_row_limit` on an enveloped lake is refused. The contrast block writes past the inlining limit so a real data file with a real key exists.
- `attach_options.test` — `ENCRYPTION_LAKE_ID` without `ENCRYPTION_SOCKET`, an empty `ENCRYPTION_SOCKET`, and a cache TTL above the ceiling are each refused at ATTACH.
- `rewrap_keys.test` — rotate, sweep, old data still reads, second sweep is a no-op, fresh ATTACH reads under the rotated key.
- `self_test.test` — no lake, enveloped lake, plain lake (a plain lake reports healthy).

Every refusal test is proven load-bearing: with the guards disabled in a mutant build, each guarded scenario succeeds — so each `statement error` block fails against a build missing its guard.

Each commit builds standalone; the full unittest suite is green at both the wiring commit (provider absent — the regression case for existing lakes) and the tip. The four new tests self-skip when the env var is unset, so CI without the fixture is unaffected.

### DIFF
--- a/src/common/ducklake_util.cpp
+++ b/src/common/ducklake_util.cpp
@@ -489,6 +489,21 @@ string DuckLakeUtil::EncryptionKeyLiteral(const string &key) {
 	return "'" + Blob::ToBase64(string_t(key)) + "'";
 }
 
+string DuckLakeUtil::StoredEncryptionKeyLiteral(const string &stored_key) {
+	if (stored_key.empty()) {
+		return "NULL";
+	}
+	return SQLLiteralToString(stored_key);
+}
+
+void DuckLakeUtil::EncodeStoredEncryptionKeys(vector<string> &keys) {
+	for (auto &key : keys) {
+		if (!key.empty()) {
+			key = Blob::ToBase64(string_t(key));
+		}
+	}
+}
+
 const char *DuckLakeUtil::BoolLiteral(bool v) {
 	return v ? "true" : "false";
 }
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -124,6 +124,11 @@ static void LoadInternal(ExtensionLoader &loader) {
 	// Register murmur3_32 scalar function for Iceberg-compatible bucket partitioning
 	auto murmur3_func = DuckLakeMurmur3Function();
 	loader.RegisterFunction(murmur3_func);
+
+	auto rewrap_keys = DuckLakeRewrapKeysFunction();
+	loader.RegisterFunction(rewrap_keys);
+	auto self_test = DuckLakeSelfTestFunction();
+	loader.RegisterFunction(self_test);
 }
 
 void DucklakeExtension::Load(ExtensionLoader &loader) {
--- a/src/ducklake_extension.cpp
+++ b/src/ducklake_extension.cpp
@@ -12,6 +12,8 @@
 #include "duckdb/function/scalar_function.hpp"
 #include "duckdb/storage/storage_extension.hpp"
 #include "storage/ducklake_log_type.hpp"
+#include "storage/ducklake_reference_encryption_provider.hpp"
+#include <cstdlib>
 
 namespace duckdb {
 
@@ -129,6 +131,15 @@ static void LoadInternal(ExtensionLoader &loader) {
 	loader.RegisterFunction(rewrap_keys);
 	auto self_test = DuckLakeSelfTestFunction();
 	loader.RegisterFunction(self_test);
+
+	// Test fixture only, and opt-in: a build that does not set the variable registers no factory, so an
+	// ATTACH with encryption_socket is refused rather than served by a provider with a public key.
+	auto *reference_provider = std::getenv("DUCKLAKE_TEST_REFERENCE_ENCRYPTION_PROVIDER");
+	if (reference_provider && string(reference_provider) == "1") {
+		RegisterDuckLakeReferenceEncryptionProviderForTests();
+		auto set_active_version = DuckLakeReferenceProviderSetActiveVersionFunction();
+		loader.RegisterFunction(set_active_version);
+	}
 }
 
 void DucklakeExtension::Load(ExtensionLoader &loader) {
--- a/src/functions/CMakeLists.txt
+++ b/src/functions/CMakeLists.txt
@@ -18,7 +18,9 @@ add_library(
   ducklake_table_changes.cpp
   ducklake_table_info.cpp
   ducklake_table_insertions.cpp
-  ducklake_murmur3.cpp)
+  ducklake_murmur3.cpp
+  ducklake_rewrap_keys.cpp
+  ducklake_self_test.cpp)
 set(ALL_OBJECT_FILES
     ${ALL_OBJECT_FILES} $<TARGET_OBJECTS:ducklake_functions>
     PARENT_SCOPE)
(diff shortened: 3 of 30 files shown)


tiagotaveiragomes:
Carrying this branch's code into a fork that also builds the `duckdb` binary turns up one link defect. Reporting it here rather than in review prose because it is measured, not predicted.

**`DEFAULT_CACHE_TTL_SECONDS` / `MAX_CACHE_TTL_SECONDS` are ODR-used and have no out-of-line definition.**

`src/storage/ducklake_catalog.cpp` on 594098c1 passes `MAX_CACHE_TTL_SECONDS` straight into the `InvalidInputException` format call:

```cpp
throw InvalidInputException("encryption_cache_ttl_seconds must be between 0 and %lld, not %lld",
                            DuckLakeEncryptionProvider::MAX_CACHE_TTL_SECONDS, ...);
```

That is a variadic. A vararg argument takes no lvalue-to-rvalue conversion, so the constant is ODR-used and the linker wants an address. DuckLake builds at `CMAKE_CXX_STANDARD 11` (`duckdb/CMakeLists.txt`), where a `static constexpr` member still needs its out-of-line definition; the C++17 implicit-inline rule that would make this a non-issue does not apply.

What makes it easy to miss is that it is asymmetric. The LOADABLE extension links fine. Everything that links `libducklake_extension.a` statically does not:

```
undefined reference to `duckdb::DuckLakeEncryptionProvider::MAX_CACHE_TTL_SECONDS'
```

reported for `duckdb`, `test/unittest` and `tools/plan_serializer`.

Fix is two lines in `src/storage/ducklake_encryption_provider.cpp`:

```cpp
constexpr int64_t DuckLakeEncryptionProvider::DEFAULT_CACHE_TTL_SECONDS;
constexpr int64_t DuckLakeEncryptionProvider::MAX_CACHE_TTL_SECONDS;
```

Whether this reds your own CI depends on whether it builds the statically linked targets on this branch. It reds ours.

tiagotaveiragomes:
Fixed in the updated branch (ba15ad03): out-of-line definitions for `DEFAULT_CACHE_TTL_SECONDS` and `MAX_CACHE_TTL_SECONDS` added in `ducklake_encryption_provider.cpp`; `nm` on the compiled object shows both symbols defined.

PR #1425, opened 2026-08-30
Stats filter left join

J-Meyers:
Splits a piece out of #1421 that stands on its own: join each column's stats CTE once instead of testing membership against it with a pair of `IN (SELECT ...)` subqueries. A file with no stats row for the column becomes `col_N_stats.data_file_id IS NULL` rather than a `NOT IN`.

Once several stats CTEs are in scope the stat names are ambiguous, so the existing virtual leaf methods take the CTE alias and qualify what they emit. No metadata backend needs changes.

Behaviour is unchanged. Measured against a copy of a 590k-file catalog (11.4M rows in `ducklake_file_column_stats`), filtering one table on N columns - one condition per column, which is the shape `main` already generates:

| filtered columns | before | after |
| --- | --- | --- |
| 1 | 17 ms | 15 ms |
| 2 | 27 ms | 26 ms |
| 4 | 51 ms | 30 ms |
| 6 | 74 ms | 30 ms |
| 8 | 102 ms | 32 ms |
| 10 | 126 ms | 32 ms |

Roughly 12 ms per filtered column before, flat after. It matters more under #1421, where one column can carry many conditions: on the same filter tree a 40 branch disjunction goes from 730 ms to 52 ms.

Every stats CTE is now referenced exactly once, so `reference_count` and the `MATERIALIZED` hint it drove are both gone - `cte_inlining` inlines at a reference count of one regardless of the hint, and `MATERIALIZED` would have blocked the inline. Also drops three things in this path that turned out to be dead: `GenerateFilterFromTableFilter`, `FilterPushdownQueryComponents::order_by_clause` and the single-argument `FilterSQLResult` constructor.

`ducklake_file_column_stats` holds at most one row per `(table_id, column_id, data_file_id)`, so the join cannot duplicate a data file row. The Top-N path already depended on this; it is now load-bearing for every filtered scan, which seems worth stating in the spec.

`topn_file_pruning.test` used `LEFT JOIN col_1_stats` to pick out the Top-N query, which no longer identifies it now that every filtered query has that join, and its PostgreSQL-only assertion matched on `col_1_stats AS MATERIALIZED`.

Done with AI help

### DIFF
--- a/src/storage/ducklake_metadata_manager.cpp
+++ b/src/storage/ducklake_metadata_manager.cpp
@@ -1,4 +1,6 @@
 #include "storage/ducklake_metadata_manager.hpp"
+
+#include <algorithm>
 #include "duckdb/common/file_system.hpp"
 #include "functions/ducklake_table_functions.hpp"
 #include "storage/ducklake_transaction.hpp"
@@ -1292,17 +1294,23 @@ static void SetSnapshotFilter(const DuckLakeSnapshot &snapshot, idx_t max_partia
 	file_entry.snapshot_filter_max = snapshot.snapshot_id;
 }
 
+//! Reference a stats column through the join that brought it in
+static string StatsColumn(const string &stats_alias, const string &stat) {
+	return stats_alias + "." + stat;
+}
+
 static bool IsSimpleFilterSubject(const Expression &expr) {
 	return expr.GetExpressionClass() == ExpressionClass::BOUND_REF ||
 	       expr.GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF;
 }
 
 string DuckLakeMetadataManager::GenerateFilterFromTableFilter(const ExpressionFilter &filter, const LogicalType &type,
-                                                              unordered_set<string> &referenced_stats) {
+                                                              unordered_set<string> &referenced_stats,
+                                                              const string &stats_alias) {
 	if (!filter.expr) {
 		return string();
 	}
-	return GenerateFilterFromExpression(*filter.expr, &type, referenced_stats);
+	return GenerateFilterFromExpression(*filter.expr, &type, referenced_stats, stats_alias);
 }
 
 bool DuckLakeMetadataManager::ValueIsFinite(const Value &val) {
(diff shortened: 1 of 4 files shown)


src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1609,24 +1617,22 @@ FilterSQLResult DuckLakeMetadataManager::ConvertFilterPushdownToSQL(const Filter
...
-			                                 "data.data_file_id IN (SELECT data_file_id FROM %s WHERE %s(%s)))",
-			                                 cte_name, cte_name, null_checks.c_str(), filter_condition.c_str());
+			conditions += StringUtil::Format("(%s.data_file_id IS NULL OR (%s(%s)))", cte_name.c_str(),
+			                                 null_checks.c_str(), filter_condition.c_str());
 		}
 
-		CTERequirement req(column_filter.column_field_index, referenced_stats);
-		req.reference_count = 2;
-		result.required_ctes.emplace(column_filter.column_field_index, std::move(req));
+		result.required_ctes.emplace(column_filter.column_field_index,
```
pdet, line 1635: Will this work for all null files?

e.g.,
```sql
ATTACH 'ducklake:/tmp/g.db' AS dl (DATA_PATH '/tmp/g_files', DATA_INLINING_ROW_LIMIT 0);
CREATE TABLE dl.t(i INTEGER);
INSERT INTO dl.t SELECT range FROM range(100);
INSERT INTO dl.t VALUES (NULL);  

SELECT count(*) FROM dl.t WHERE i = 5 OR i IS NULL; 
SELECT count(*) FROM dl.t WHERE i IS NULL; 

```

src/storage/ducklake_metadata_manager.cpp
```diff
@@ -1644,6 +1650,24 @@ string DuckLakeMetadataManager::GenerateFileColumnStatsCTEBody(const CTERequirem
...
 }
 
+string DuckLakeMetadataManager::GenerateStatsJoinList(const unordered_map<idx_t, CTERequirement> &requirements) {
+	// sorted so the generated query is stable across runs - the requirements live in a hash map
+	vector<idx_t> field_indexes;
+	field_indexes.reserve(requirements.size());
+	for (const auto &entry : requirements) {
+		field_indexes.push_back(entry.first);
+	}
+	std::sort(field_indexes.begin(), field_indexes.end());
```
pdet: Is this semantically necessary?

pdet (commented):
Thanks for the PR, a couple of comments.

src/storage/ducklake_metadata_manager.cpp
J-Meyers: No, removed 4daf7e3

src/storage/ducklake_metadata_manager.cpp
J-Meyers: This didn't introduce any issues, but didn't fix the existing issue either, cherry picked the changes from #1421 773a98a

pdet:
Thanks, LGTM

PR #1426, opened 2026-08-30
Column stats opt out

J-Meyers:
Adds a table-scoped `skip_stats_columns` option that stops `min_value` / `max_value` being recorded for the named columns, keeping `value_count`, `null_count` and `column_size_bytes`. Bounds are written for every column of every file whether or not anyone filters on it, and for a wide column - a BLOB payload, an opaque JSON blob - they can dominate the stats row while pruning nothing.

```sql
CALL ducklake.set_option('skip_stats_columns', 'payload, raw_json', table_name => 't');
CALL ducklake.set_option('skip_stats_columns', ['payload', 'raw_json'], table_name => 't');
CALL ducklake.set_option('skip_stats_columns', '', table_name => 't');  -- clears
```

A name containing a comma is rejected in the delimited form, since splitting it would resolve to different columns; the list form names it unambiguously.

The skip covers the named field and its subtree, and applies wherever stats are produced: INSERT / COPY / CTAS, UPDATE, MERGE INTO, inlined data and its flush, `merge_adjacent_files`, `rewrite_data_files`, and `ducklake_add_data_files` - both footer stats and hive-path stats, the latter being the one place where skipping a column gives up real pruning.

Storage is one `ducklake_metadata` row, `scope = 'table'`, holding comma-separated field ids - no new table, column or scope value. Ids because they survive `RENAME COLUMN`; `ducklake.options()` reports the stored value as-is, as it does for every other option and as `ducklake_partition_column` already does for partition columns. It is re-read on every write to the table - once per write, not per row - so it never throws (consistent with many other settings), ignoring entries it cannot use, and it looks in the table scope only. - Previously had a custom renderer to make clear the meaning of the ids and what was resolved vs not, reverted it since not consistent with how the partition columns are handled now, but trivial to switch back.

`set_option` also rejects unknown names, partition columns, a missing `table_name`, and GEOMETRY / VARIANT. The last because `ClearBounds()` clears min/max and nothing else, while a geometry bbox lives in `extra_stats` and variant bounds in `ducklake_file_variant_stats` - skipping either would clear almost nothing. Clearing the bbox instead is unsound: `DuckLakeColumnGeoStats::Merge` has no "unknown" state, so an emptied one merges as identity and silently shrinks the table extent.

Table bounds go unknown as soon as one bound-less file exists, in either write order, which keeps mixed populations sound; clearing the list does not bring them back, since `MergeStats` will not re-seed past `bounds_unknown`.

`add_data_files` reads the footer bounds once per row group per column, so a skipped column now never materializes them rather than dropping them afterwards.

Done with AI help


test/sql/stats/skip_stats_columns.test
```diff
@@ -0,0 +1,451 @@
+# name: test/sql/stats/skip_stats_columns.test
```
pdet, line 1: Can we add a test where we rollback a setting?
e.g.,
```
CREATE TABLE dl.rb(id INTEGER, s VARCHAR);
BEGIN;
CALL dl.set_option('skip_stats_columns', 's', table_name => 'rb');
ROLLBACK;
INSERT INTO dl.rb VALUES (1, 'aaa');
SELECT min_value, max_value FROM meta.ducklake_file_column_stats WHERE column_id = 2;

Comment (pdet) [reply test/sql/stats/skip_stats_columns.test:1]: Can we add a test where we directly change the type of a nested type and see if it fails??

```sql
CREATE TABLE dl.nested(id INTEGER, st STRUCT(x INTEGER, y VARCHAR));
CALL dl.set_option('skip_stats_columns', 'st', table_name => 'nested');
ALTER TABLE dl.nested ALTER COLUMN st TYPE STRUCT(x INTEGER, y VARCHAR, g GEOMETRY);
```

pdet: Can we add a separate test with multiple connections to ensure the option is loaded with the snapshot?

pdet: Can we also split up this file on different concerns? almost 500 lines of tests is a bit difficult to review.

src/functions/ducklake_set_option.cpp
```diff
@@ -68,6 +69,79 @@ static void ValidateNoReservedInliningColumns(ClientContext &context, Catalog &c
...
+	for (auto &field : partition_data->fields) {
+		if (field.field_id == field_id.GetFieldIndex()) {
+			throw InvalidInputException("Statistics cannot be skipped for partition column \"%s\"", field_id.Name());
+		}
+	}
+}
+
+//! Resolves a LIST of column names, or a comma-separated string, to the field ids the option
+//! stores - ids are used because they survive a rename
+static string ResolveSkippedStatsColumns(DuckLakeTableEntry &table, const Value &val, const string &raw_value) {
```
pdet: I think we should limit the multi-column options with just a list.

```
CALL dl.set_option('skip_stats_columns', 's', table_name => 't');            -- one column
CALL dl.set_option('skip_stats_columns', ['s', 'd'], table_name => 't');     -- several
CALL dl.set_option('skip_stats_columns', ['a,b'], table_name => 't');        -- odd name, no special case
CALL dl.set_option('skip_stats_columns', [], table_name => 't');             -- clear 
```

pdet (commented):
Thanks for the PR, a few comments.

src/functions/ducklake_set_option.cpp
J-Meyers: Done

test/sql/stats/skip_stats_columns.test
J-Meyers: Options, not just this one are not specific to a single transaction and aren't isolated as it is right now, broken out in a different PR, resolving it first should simplify this one: https://github.com/duckdb/ducklake/pull/1429

Nested type is now properly handled and added explicit tests

Tests split into different files

pdet:
Hi @J-Meyers thanks for the work, looks good to me, do you mind fixing the conflict?

J-Meyers:
> Hi @J-Meyers thanks for the work, looks good to me, do you mind fixing the conflict?

Done, however there seems to be a flaky test, failing on various other branches as well

Created https://github.com/duckdb/duckdb/pull/25378

src/storage/ducklake_catalog.cpp
```diff
@@ -953,51 +953,73 @@ optional_idx DuckLakeCatalog::GetCatalogVersion(ClientContext &context) {
...
+	}
+	scope[option.option.key] = option.option.value;
+	return undo;
+}
+
+void DuckLakeCatalog::UndoConfigOption(const DuckLakeConfigOptionUndo &undo) {
+	lock_guard<mutex> guard(config_lock);
+	auto &scope = GetOptionScope(options, undo.option);
+	auto entry = scope.find(undo.option.option.key);
+	if (entry == scope.end() || entry->second != undo.option.option.value) {
```
pdet, line 984: The clanker flagged an issue here where a rollback removes another transaction’s committed option.

e.g.,
```sql
-- Connection 1
LOAD ducklake;
ATTACH 'ducklake::memory:' AS lake (
    METADATA_CATALOG 'meta',
    DATA_PATH '/tmp/ducklake-repro-rollback',
    DATA_INLINING_ROW_LIMIT 0
);
CREATE TABLE lake.t(s VARCHAR);

BEGIN;
CALL lake.set_option('skip_stats_columns', 's', table_name => 't');

-- Connection 2, autocommit
CALL lake.set_option('skip_stats_columns', 's', table_name => 't');

-- Connection 1
ROLLBACK;
INSERT INTO lake.t VALUES ('aaa');

SELECT value
FROM meta.ducklake_metadata
WHERE key = 'skip_stats_columns';
-- Returns '1': the committed option still exists.

SELECT min_value, max_value
FROM meta.ducklake_file_column_stats;
-- Actual:   aaa, aaa
-- Expected: NULL, NULL
```

test/sql/stats/skip_stats_columns_transaction.test
```diff
@@ -0,0 +1,136 @@
+# name: test/sql/stats/skip_stats_columns_transaction.test
```
pdet, line 1: The clanker also flagged an issue where delete rewrite recreates skipped bounds.

e.g.,
```sql
LOAD ducklake;
ATTACH 'ducklake::memory:' AS lake (
    METADATA_CATALOG 'meta',
    DATA_PATH '/tmp/ducklake-repro-rewrite',
    DATA_INLINING_ROW_LIMIT 100
);
CREATE TABLE lake.t(s VARCHAR);
CALL lake.set_option('skip_stats_columns', 's', table_name => 't');

INSERT INTO lake.t VALUES ('old');
DELETE FROM lake.t;
CALL ducklake_flush_inlined_data('lake');

INSERT INTO lake.t VALUES ('new');

SELECT min_value, max_value
FROM meta.ducklake_table_column_stats;
-- NULL, NULL

CALL ducklake_rewrite_data_files('lake', 't', delete_threshold => 0);

SELECT min_value, max_value
FROM meta.ducklake_table_column_stats;
-- Actual:   new, new
-- Expected: NULL, NULL
```

pdet (commented):
Hi @J-Meyers thanks for the PR again, looks good to me, just raising a few issues from ze clanker.

src/storage/ducklake_catalog.cpp
J-Meyers: The undo can't fix this in place, it only compares the current value against the one it wrote, so it can't tell the difference between another transaction setting a different value and another transaction setting the same value

#1429 removes this potential source of issues entirely, because options are staged on the transaction and only published on commit there is nothing to undo. Merging that first resolves this and then I'll drop the undo. - Keeping the undo for now since removing it reintroduces other issues on this branch and the tests will fail without.

test/sql/stats/skip_stats_columns_transaction.test
J-Meyers: Fixed and added corresponding test

pdet:
Thanks!

PR #1428, opened 2026-09-01
Support file pruning with prefix range filter

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/10598

This PR introduces PRF support in DuckLake by generalizing the dynamic TopN pruning.


src/storage/ducklake_multi_file_reader.cpp
```diff
@@ -153,6 +155,46 @@ static bool CanSkipFileByTopNDynamicFilter(const DuckLakeFileListEntry &file_ent
...
+		column_stats.has_min = stats_entry->second.has_min;
+		column_stats.has_max = stats_entry->second.has_max;
+
+		Value casted_bound;
+		if (RequiresValueComparison(column_filter.column_type) &&
+		    ((column_stats.has_min &&
+		      !Value(column_stats.min).DefaultTryCastAs(column_filter.column_type, casted_bound, nullptr)) ||
+		     (column_stats.has_max &&
+		      !Value(column_stats.max).DefaultTryCastAs(column_filter.column_type, casted_bound, nullptr)))) {
+			continue;
```
pdet: Should we move the failure handilng to `CreateNumericStats`?

src/include/storage/ducklake_metadata_info.hpp
```diff
@@ -386,6 +386,13 @@ enum class DuckLakeDataType {
 	TRANSACTION_LOCAL_INLINED_DATA,
 };
 
+struct DuckLakeFileColumnStats {
+	string min;
+	string max;
+	bool has_min = false;
+	bool has_max = false;
+};
```
pdet, line 394: Maybe we can also use it in `DuckLakeColumnStatsInfo`

src/storage/ducklake_multi_file_reader.cpp
```diff
@@ -153,6 +155,46 @@ static bool CanSkipFileByTopNDynamicFilter(const DuckLakeFileListEntry &file_ent
 	return false;
 }
 
+static bool CanSkipFileByPrefixRangeFilter(const DuckLakeFileListEntry &file_entry,
```
pdet: Would it make sense to merge with `CanSkipFileByTopNDynamicFilter` into a `CanSkipFileByRuntimeFilter`?

pdet (commented):
Thanks for the PR, just a couple of comments.

Could we also have the Blob stats on a different PR?

src/include/storage/ducklake_metadata_info.hpp
d-justen: I think I would leave those separate for now. If I understand this correctly `DuckLakeColumnStatsInfo` is more of a SQL serialization structure, in which `min_val`/`max_val` may mismatch the representation here (e.g., as the literal string "NULL"). Trying to harmonize this is probably out of scope for this PR.

pdet:
There is a merge conflict, but then good to go imo

pdet:
Thanks!

Issue #1267, opened 2026-06-26
BUG: quack-based catalog doesn't support rollback


dentiny:
### What happens?

See attached repro SQL for details, should be straightforward.

### To Reproduce

```sql
# name: test/sql/quack/quack_query_by_name_rollback_repro.test
# description: quack_query_by_name should participate in the outer transaction rollback
# group: [quack]

require quack

require httpfs

set ignore_error_messages __none__

statement ok con1
CALL quack_serve('quack:localhost', token='asdf');

statement ok con2
CREATE SECRET (TYPE quack, TOKEN 'asdf');

statement ok con2
ATTACH 'quack:localhost' AS rpc;

statement ok con2
BEGIN;

statement ok con2
CALL system.main.quack_query_by_name('rpc', 'CREATE TABLE rollback_bug(i INTEGER); INSERT INTO rollback_bug VALUES (42);');

statement ok con2
ROLLBACK;

# Refresh the attached catalog snapshot so we observe the remote server state.
statement ok con2
CALL quack_clear_cache();

# Correct behavior: table creation/insert should have rolled back remotely too.
statement error con2
SELECT * FROM rpc.rollback_bug;
----
Table with name rollback_bug does not exist
```
Build command
```sh
ENABLE_QUACK=1 BUILD_EXTENSION_TEST_DEPS=full CMAKE_BUILD_PARALLEL_LEVEL=10 make reldebug
```
The error message I get is
```sh
================================================================================
Query unexpectedly succeeded! (test/sql/quack/quack_query_by_name_rollback_repro.test:34)!
================================================================================
SELECT * FROM rpc.rollback_bug;
================================================================================
i
INTEGER
[ Rows: 1]
42
```
The root cause is, when ducklake issues query through quack, it sends metadata SQL through quack_query_by_name, which sends a PrepareRequestMessage directly and does not route through QuackTransaction's remote BEGIN/ROLLBACK path.

### OS:

macos

### DuckDB Version:

v1.5.4

### DuckLake Version:

96a4574d65665d15f5ca158892d84493a19d4c62

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

[...]

PR #1429, opened 2026-09-02
Transactional config options

J-Meyers:
Motivated by https://github.com/duckdb/ducklake/pull/1426#discussion_r3914313693

`ducklake_set_option` writes its `ducklake_metadata` row at statement time and mutates the catalog's in-memory option map in the same call. It is the only change that does this - everything else is staged on the transaction and written by `FlushChanges` at commit. So a rolled back option stays in effect, because the map is never restored:

```sql
BEGIN;
CALL dl.set_option('parquet_compression', 'gzip', table_name => 't');
ROLLBACK;
INSERT INTO dl.t VALUES (1);  -- writes a GZIP file
```

And since there is one copy of the map, an uncommitted option is visible to every other connection - `con1` sets `data_inlining_row_limit` in an open transaction and `con2`'s next insert inlines against a value that may never commit.

This stages options on `DuckLakeTransaction` instead. `SetConfigOption` records the option against its scope and writes nothing; `Commit` writes the rows before the flush so they land in the same metadata transaction, and publishes to the shared map only once that succeeded. A rollback needs no cleanup. Reads consult the transaction first, and `ducklake_options()` overlays the staged values so a transaction sees what it just set.

## Relationship to #1267

The rollback half is duckdb/ducklake#1267: on quack a statement sent through `quack_query_by_name` never starts a server-side transaction, so `QuackTransaction` reads `TRANSACTION_NOT_YET_STARTED` and sends no `ROLLBACK`. duckdb/duckdb-quack#182 is the open fix, and it is not ready to carry this - patched in with DuckLake untouched it fixes `issue_852_cache_growth.test` but regresses both tests added here, which pass on quack today, because `GetConnection()` already opens an explicit transaction and `QuackMetadataManager` issues its own `ROLLBACK; BEGIN TRANSACTION;` on error.

Done with AI help

### DIFF
--- a/src/storage/ducklake_transaction.cpp
+++ b/src/storage/ducklake_transaction.cpp
@@ -723,7 +723,8 @@ DuckLakeTransaction::DuckLakeTransaction(DuckLakeCatalog &ducklake_catalog, Tran
       local_catalog_id(DuckLakeConstants::TRANSACTION_LOCAL_ID_START), catalog_version(0) {
 	metadata_manager = DuckLakeMetadataManager::Create(*this);
 	schema_pins = make_uniq<DuckLakeSchemaPinState>();
-	state = make_uniq<DuckLakeTransactionState>(db, ducklake_catalog.IsCommitInfoRequired(), new_name_maps,
+	// nothing is staged yet, so this reads the committed value
+	state = make_uniq<DuckLakeTransactionState>(db, ducklake_catalog.IsCommitInfoRequired(nullptr), new_name_maps,
 	                                            ducklake_catalog.DataPath(), ducklake_catalog.Separator());
 }
 
@@ -767,6 +768,8 @@ void DuckLakeTransaction::Start() {
 }
 
 void DuckLakeTransaction::Commit() {
+	// written before the commit below so they land in the same metadata transaction
+	WriteConfigOptions();
 	if (ChangesMade()) {
 		FlushChanges();
 	} else if (connection) {
@@ -775,6 +778,7 @@ void DuckLakeTransaction::Commit() {
 			DropEmptySupersededInlinedTablesClientSide();
 		}
 	}
+	ApplyConfigOptions();
 	FlushNameMapCacheInvalidations();
 	connection.reset();
 	state->local_changes.Clear();
@@ -1464,6 +1468,8 @@ void DuckLakeTransaction::RunCommitLoop(DuckLakeSnapshot transaction_snapshot,
 	context.prepare_retry = [&]() {
 		metadata_manager->ClearInlinedTableCaches();
 		connection->BeginTransaction();
+		// the rolled back attempt took the option rows with it
+		WriteConfigOptions();
 		snapshot.reset();
 	};
 	context.query_metadata = [&](string q) {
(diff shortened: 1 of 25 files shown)


src/storage/ducklake_transaction.cpp
```diff
@@ -769,6 +770,8 @@ void DuckLakeTransaction::Start() {
 }
 
 void DuckLakeTransaction::Commit() {
+	// written before the commit below so they land in the same metadata transaction
+	WriteConfigOptions();
```
pdet, line 774: Shouldn't this run in the commit's loop rollback protection? What happens in conflicts?

J-Meyers: Rollback protection: It does and this is tested in config_option_commit_retry.test I've made the change so that it writes with a single delete + insert rather than branching in C++ so replays are now trivially safe and reduces the number of round trips, previously two per option now one for all staged options regardless of how many

Conflicts: Conflicts on options are complex, right now writes to options are NOT checked for conflicts, two transactions writing the same key and scope would both insert because there is no uniqueness restriction on those columns, this is a preexisting issue, this only applies when either is an option only commit - otherwise they are retried due to snapshot id incrementing

### Other remaining (prexisting) issues:
- Too fresh options within an instance: A transaction that began before another set an option picks it up mid-flight
- Too stale across instances: 'options.config' is populated once at ATTACH and there is no reload path, Instance B never sees Instance A's committed option change until reattaching
- Invalidating cached option reads cheaply is hard
- Treating an option I read changed as a conflict would likely abort large numbers of unrelated transactions (depending on actual frequency of option changes)
- Options are not snapshot versioned, this would fix many issues, but has some complexities since some things like version data_path and encrypted don't make sense to associate with specific snapshots and run before any snapshot exists and version determines which migrations run and what the actual schema is which introduces a potential circular ordering

pdet:
Hi, I ran the clanker on it, and it found two issues:

1. Concurrent schema change bypasses validation
```sql
-- Connection A
ATTACH 'ducklake:/tmp/stats_race.db' AS dl (
    DATA_PATH '/tmp/stats_race_data',
    DATA_INLINING_ROW_LIMIT 0,
    METADATA_CATALOG 'meta'
);
CREATE TABLE dl.t(i INTEGER);

BEGIN;
CALL dl.set_option('skip_stats_columns', 'i', table_name => 't');

-- Connection B
ALTER TABLE dl.t SET PARTITIONED BY (i);

-- Connection A
COMMIT;

INSERT INTO dl.t VALUES (1);
SELECT column_id, min_value, max_value
FROM meta.ducklake_file_column_stats;
```

Actual: both commits succeed; column i has NULL min/max despite being a partition column.
Expected: A’s commit rejects the option because B changed the schema incompatibly.

2. Quack retains an option after a failed commit
```sql
CREATE SECRET review_quack (
    TYPE quack,
    TOKEN 'review-token'
);

ATTACH 'ducklake:quack:localhost:20057' AS dl (
    DATA_PATH '/tmp/quack_failed_commit_data',
    DATA_INLINING_ROW_LIMIT 0,
    METADATA_CATALOG 'meta'
);

CREATE TABLE dl.t(i INTEGER);
CALL dl.set_option('require_commit_message', true);

BEGIN;
CALL dl.set_option('skip_stats_columns', 'i', table_name => 't');
INSERT INTO dl.t VALUES (1);
COMMIT;
-- Fails: commit information is required.

SELECT count(*) FROM dl.t;
-- 0: the insert did not commit.

SELECT key, value
FROM meta.ducklake_metadata
WHERE key = 'skip_stats_columns';
-- Actual: skip_stats_columns | 1
-- Expected: no rows.
```

PR #1436, opened 2026-09-07
fix(views): Handle NULL ducklake_view.column_aliases when loading views

pscls:
tl;dr; This PR fixes a DuckLake spec discrepancy. We can either fix it in the extension, or the official spec should be updated to `NOT NULL`. I'm happy with both; please feel free to push back.

**Background**

The official DuckLake spec states for the [ducklake_view](https://ducklake.select/docs/stable/specification/tables/ducklake_view) table:
> `column_aliases` contains a possible rename of the view columns. **Can be NULL if no rename is set**.

DuckLake itself always writes an empty string into the `column_aliases` for non-existing aliases ([see here](https://github.com/duckdb/ducklake/blob/0f569a6bba3028637c5020ccd31af24739e3fb42/src/storage/ducklake_metadata_manager.cpp#L3133-L3150)), but other engines might write a `NULL` value if no aliases exist. This is consistent with the Ducklake specs (see above).

Trying to load such catalog data will fail with `Invalid Input Error: Failed to parse quoted value - expected a quote`. Views are loaded as part of reading the catalog; a single such row makes the whole catalog unreadable, not just the affected view

**Summary**

This PR skips the parse when the column is `NULL`, leaving `column_aliases` empty. This will resolve the read to the same result as the empty-string encoding DuckLake writes itself.

I deliberately did not change the write path in [DuckLakeMetadataManager::WriteNewViews](https://github.com/duckdb/ducklake/blob/0f569a6bba3028637c5020ccd31af24739e3fb42/src/storage/ducklake_metadata_manager.cpp#L3133-L3150) as this would be a breaking change. Older DuckLake versions would not be able to read the table anymore, as they do not support `NULL` values. The write fix would simply be:
```
string column_aliases_sql =
	view.column_aliases.empty() ? "NULL" : SQLString(DuckLakeUtil::ToQuotedList(view.column_aliases));
view_insert_sql += StringUtil::Format("(%d, '%s', {SNAPSHOT_ID}, NULL, %d, %s, %s, %s, %s)", view.id.index,
										view.uuid, schema_id, SQLString(view.name), SQLString(view.dialect),
										SQLString(view.sql), column_aliases_sql);
```

**Testing**

The new test `test/sql/view/view_null_column_aliases.test` creates 3 views with empty-string aliases (current write behavior), `NULL` aliases (written directly into the metadata catalog to simulate another spec-compliant writer), and real aliases, then re-attaches and verifies all three views bind and return the expected columns.


pdet:
Thanks!

PR #1440, opened 2026-09-09
CREATE and CTAS SORTED BY and PARTITIONED BY- #1107

pdet:
Picking up #1107


Alex-Monahan:
Thank you so much! I owe you one for your help on this for sure. It has stayed on my list just couldn't rise to the top of it with everything else piling on.

pdet:
Not a problem @Alex-Monahan, my pleasure, and thanks for putting the PR up in the first place! 

ps: I accept payment in a beer when you come to Amsterdam again 😂

PR #1443, opened 2026-09-12
Pull the MinIO images from quay.io so the MinIO CI job can start

tiagotaveiragomes:
### Symptom

The `Tests using MinIo (AWS)` job fails in `./scripts/start-docker.sh` before any test runs:

```
pull access denied for minio/mc, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
##[error]Process completed with exit code 1.
```

### Root cause

`scripts/docker-compose.yml` pulls `minio/minio` and `minio/mc` from Docker Hub. Those references are no longer anonymously pullable. I can only report the observable fact, not why they were removed:

```
$ docker pull -q minio/minio           -> FAILED
$ docker pull -q minio/mc              -> FAILED
$ docker pull -q quay.io/minio/minio   -> OK
$ docker pull -q quay.io/minio/mc      -> OK
```

`quay.io/minio/minio` is also the reference MinIO publishes in its own [`docs/docker/README.md`](https://github.com/minio/minio/blob/master/docs/docker/README.md), and `quay.io/minio/mc` is the matching client image in the same registry namespace. Tags are left floating, as before, to keep the diff to the registry change.

### Verification

Ran the same entry point CI uses, `./scripts/start-docker.sh`, on this branch:

```
+ docker compose up --detach
 Container minio Started
 Container mc Started
Waiting for MinIO to become ready...
+ curl -s -o /dev/null http://127.0.0.1:9000/minio/health/ready
MinIO is ready

$ docker compose logs mc
mc  | Added `local` successfully.
mc  | Bucket created successfully `local/mybucket`.

$ docker compose images
minio   quay.io/minio/minio   latest
mc      quay.io/minio/mc      latest
```

and confirmed the bucket is really there from a separate client:

```
$ docker run --rm --network scripts_s3_net --entrypoint sh quay.io/minio/mc \
    -c "mc alias set l http://minio:9000 admin password >/dev/null && mc ls l"
[2026-09-12 22:46:10 UTC]     0B mybucket/
```

Then `docker compose down`.

### Unrelated pre-existing issue noticed while verifying

The `mc` service never prints its final `MinIO bucket initialized.` line and exits 2:

```
mc  | /bin/sh: -c: line 7: syntax error near unexpected token `fi'
```

The `entrypoint: >` folded block joins `# Ensure bucket exists` with the `if ...; then` line that follows it, commenting out the `if` and orphaning the `fi`. The bucket still gets created because the `mc mb` line is indented further and so survives as its own line. This is independent of the registry, so it is not touched here.


pdet:
Thanks!

PR #1444, opened 2026-09-12
Fix mc entrypoint being mangled by YAML folding

tiagotaveiragomes:
### Symptom

The `mc` init container in `scripts/docker-compose.yml` dies with a shell syntax error instead of staying up:

```
mc  | Added `local` successfully.
mc  | Bucket created successfully `local/mybucket`.
mc  | /bin/sh: -c: line 7: syntax error near unexpected token `fi'
```

Because the shell aborts there, neither `echo 'MinIO bucket initialized.'` nor `tail -f /dev/null` ever runs, and the container exits. The bucket is still created — `mc mb` is indented more deeply and survives as its own line — so the service looks like it works while both its existence guard and its keep-alive are dead.

### Root cause

`entrypoint: >` is a YAML **folded** block scalar, which joins consecutive non-blank lines with a space. That merges the `# Ensure bucket exists` comment with the `if` line right after it, so the shell actually receives:

```
# Ensure bucket exists if ! /usr/bin/mc ls local/mybucket > /dev/null 2>&1; then
  /usr/bin/mc mb local/mybucket;
fi
```

The `if` is commented out, leaving `fi` orphaned.

### Fix

Use a **literal** block scalar (`entrypoint: |`), which preserves the newlines so the comment terminates at its own line. One character; no other change to the script.

### Verification

Parsed with `yaml.safe_load` plus `shlex.split` (how Compose word-splits a string entrypoint) — argv is exactly `['/bin/sh', '-c', <script>]`, with the comment on its own line, and `sh -n` accepts the script. The same check on the pre-fix file reproduces the failure, confirming the check actually fires:

```
BAD folded, comment line: ['# Ensure bucket exists if ! /usr/bin/mc ls local/mybucket > /dev/null 2>&1; then']
BAD rc 2 err /bin/sh: -c: line 6: syntax error near unexpected token `fi'
```

Then ran the repo's own entry point, `scripts/start-docker.sh`:

```
$ docker compose logs mc
mc  | Added `local` successfully.
mc  | Bucket created successfully `local/mybucket`.
mc  | MinIO bucket initialized.

$ docker compose ps
NAME      IMAGE                 COMMAND                   SERVICE   CREATED              STATUS
mc        quay.io/minio/mc      "/bin/sh -c '\nuntil …"   mc        About a minute ago   Up 40 seconds
minio     quay.io/minio/minio   "/usr/bin/docker-ent…"    minio     About a minute ago   Up About a minute

$ docker compose exec -T mc /usr/bin/mc ls local/
[2026-09-12 22:51:53 UTC]     0B mybucket/
```

No syntax error, `MinIO bucket initialized.` is present, `mc` stays `Up` because `tail -f /dev/null` now executes, and the bucket exists.

Note on the images in that output: `minio/minio` and `minio/mc` on Docker Hub are no longer anonymously pullable, so to run the stack locally I temporarily pointed the two `image:` lines at `quay.io/...`. That was an uncommitted local change, reverted before committing — this PR touches only the `entrypoint` line, and the fix itself is image-independent.

### Relationship to #1443

Independent of #1443 and touches no overlapping line. #1443 is what makes the images pullable again; this PR fixes the entrypoint. They can be reviewed and merged in either order.


pdet:
Thanks!

PR #1445, opened 2026-09-13
Access gstate.readers in DuckLakeGetMetrics under lock

Flogex:
The new _scan_profiling_empty_build_side.test_ test failed previously with this stack trace:

```
./build/debug/test/unittest "test/sql/general/scan_profiling_empty_build_side.test"
1. test/sql/general/scan_profiling_empty_build_side.test:52
Query unexpectedly failed (test/sql/general/scan_profiling_empty_build_side.test:52)
 (test/sql/general/scan_profiling_empty_build_side.test:52)!
SELECT count(*) FROM ducklake.t WHERE i IN (SELECT i FROM empty_ids);
Actual result:
INTERNAL Error: Attempted to dereference unique_ptr that is NULL!

Stack Trace:

2  duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 220
3  duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 84
4  duckdb::ThrowNullUniquePtrDereference() + 420
5  duckdb::unique_ptr<duckdb::MultiFileReaderData, std::__1::default_delete<duckdb::MultiFileReaderData>, true>::AssertNotNull(bool) + 72
6  duckdb::unique_ptr<duckdb::MultiFileReaderData, std::__1::default_delete<duckdb::MultiFileReaderData>, true>::operator->() const + 128
7  duckdb::DuckLakeGetMetrics(duckdb::TableFunctionGetMetricsInput&) + 1684
8  duckdb::PhysicalTableScan::GetMetrics(duckdb::ClientContext&, duckdb::GlobalSourceState&, duckdb::LocalSourceState&, duckdb::OperatorMetrics&) const + 1480
9  duckdb::OperatorProfiler::FinishSource(duckdb::PhysicalOperator const&, duckdb::GlobalSourceState&, duckdb::LocalSourceState&) + 780
10 duckdb::PipelineExecutor::PushFinalize() + 2960
11 duckdb::PipelineExecutor::FlushAndFinalize(duckdb::ExecutionBudget&) + 2388
12 duckdb::PipelineExecutor::Execute(unsigned long long) + 3540
13 duckdb::PipelineExecutor::Execute() + 84
14 duckdb::PipelineTask::ExecuteTask(duckdb::TaskExecutionMode) + 1752
...
```

This is triggered when running a `DUCKLAKE_SCAN` on the probe side of a hash join with empty build side while having the profiler enabled.

### Root cause

- Query involves a hash join with an empty build side. Returns `SinkFinalizeType::NO_OUTPUT_POSSIBLE`. The probe pipeline (`DUCKLAKE_SCAN`) gets scheduled anyway.
- `PipelineExecutor` calls `MultiFileInitLocal`. The probe pipeline terminates after its first chunk because the hash table is empty, so `FinishProcessing` is called.
- `MultiFileInitLocal` takes `gstate.lock`, and when no known file is left, appends a new `MultiFileReaderData` to the shared vector.
  ```cpp
  global_state.readers.push_back(make_uniq<MultiFileReaderData>(scanned_file));
  ```
- Concurrently, `PipelineExecutor::PushFinalize()` runs and calls `function.dynamic_to_string`.
- `DuckLakeDynamicToString` accesses `gstate.readers[i]->file_state` without acquiring `gstate.lock` beforehand, but the `unique_ptr` was moved during reallocation, resulting in an `InternalException("Attempted to dereference unique_ptr that is NULL!")` being thrown.

### Fix

`DuckLakeDynamicToString` should accesses `gstate.readers` under `gstate.lock`.
We cannot copy a `unique_ptr`, and even if we could, the copying would have to happen under the lock as well.


pdet:
Thanks!

PR #1446, opened 2026-09-13
Access gstate.readers in DuckLakeGetMetrics under lock (v1.5 backport)

Flogex:
Backport #1445. I have removed the call to `SET read_ahead_depth` from the test since this doesn't exist for v1.5 Ducklake.


pdet:
Thanks!

Issue #1175, opened 2026-05-20
Data inlining of rows with JSON data type broken on postgres catalog


bezineb5:
### What happens?

Since migrating from DuckDB 1.4.4 (DuckLake 0.3) to DuckDB 1.5.2 (DuckLake 1.0), inlining fails on postgreSQL for tables with JSON columns:

```
[...]

PR #1447, opened 2026-09-14
fix postgres inline struct

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9734 

Fix data inlining for `STRUCT` values whose types are not natively supported.

The `STRUCT` handling now mirrors the existing `ARRAY` and `MAP` handling: when the type is not natively supported, the value is stored as `VARCHAR` using `ToString()`, which produces a parseable representation.

## Tests

Added regression coverage for the three reported issues:

* `STRUCT` containing `TIMESTAMP` [#9734](https://github.com/duckdblabs/duckdb-internal/issues/9734)
* `STRUCT` containing `TIMESTAMPTZ` #1155 
* `JSON` values during data inlining #1175


test/sql/data_inlining/postgres_inlining.test
```diff
@@ -0,0 +1,84 @@
+# name: test/sql/data_inlining/postgres_inlining.test
+# description: Test inlined data on postgres
+# group: [data_inlining]
+
+require ducklake
+
+require parquet
+
+require postgres_scanner
```
pdet, line 9: Do we need this? I'm wondering if the test is sufficiently generic to the point that it could just run in all our configs, hence testing on other catalogs, like sqlite.

pdet:
Thanks!

PR #1451, opened 2026-09-16
Honor ducklake_default_data_inlining_row_limit at CREATE TABLE

### LINKED ISSUE #1184: CREATE TABLE with >1600 columns fails on Postgres catalog even when data inlining is disabled
### What happens?

**What happens**

When using DuckLake with a Postgres metadata catalog, `CREATE TABLE` for a table with more than 1600 columns fails with PG’s `tables can have at most 1600 columns error`, even after `SET ducklake_default_data_inlining_row_limit = 0`.

It seems the root cause is that `CREATE TABLE` in DuckLake unconditionally emits a `CREATE TABLE ducklake_inlined_data_<table_id>_0 (<every user col> BYTEA, ...)` statement against the PG catalog in the same transaction as the catalog metadata writes. That backing-table CREATE hits PG’s per-table 1600-column hard limit.

`SET ducklake_default_data_inlining_row_limit = 0` suppresses row inlining but the inlined-data backing-table schema is still created with all user columns (it's just left empty).

We are trying to work with a healthcare dataset containing some 1700 fields, and running into this error. So far mitigation attempts have failed for the same issue:
- parquet-out + ducklake_add_data_files — still requires CREATE TABLE on first write
- batched `ALTER TABLE ADD COLUMN` to seed under 1600 then wide; each ALTER produces a brand-new `ducklake_inlined_data_<tid>_<v>`; the ALTER that hits the 1600 limit fails.
- Setting `data_inlining_row_limit = 0` at `ATTACH` time — same surface as the session SET, no effect on backing creation

A related symptom we hit in development from the same backing-table-on-every-schema-bump behavior: our Ducklake catalog accumulated 2,288 `ducklake_inlined_data_*` PG tables with only 1 actually in use; this `pg_class x pg_attribute x pg_namespace` join run by the DuckDB Postgres extension runs at every `ATTACH 'ducklake:'` and then scales linearly with this count, so under high concurrency our `ATTACH` cost went from ~1s to ~30s. We worked around it locally by scoping a post-write `DROP TABLE` of empty backings, but the underlying behavior is the same: the inlined backing-table is created unconditionally regardless of inlined-disabled setting.

**Environment**
- DuckDB 1.5.2
- Ducklake extension: 415a9ebd
- Postgres: Azure Flexible Server 17.x
- Client OS : Linux (AKS pod), reproduced on macOS local
- Workload: Dagster code-location pod with WIF auth; identical behavior on laptop with `az login` credentials

**Related**
- #781 - Inlining feature noted as “unsupported” on Postgres catalogs despite ducklake_set_option silently accepting it (open)
- #921 - data_inlining_row_limit not honored on MERGE UPDATE (closed; row-side variant of the same theme)
- #1088 - Empty inlined backing tables not cleaned up (closed; we work around this in user space)
- #1018 - Unbounded memory growth on wide sparse tables (closed; we hit this when probing ALTER-based workarounds — 200 sequential ALTER ADD COLUMNs OOM-killed a 2 GiB pod)

generated with help from Claude Code - behavior verified by me and report drafted by me

### To Reproduce

```
# Local Postgres for repro
docker run --rm -d --name pg-ducklake-repro \
  -e POSTGRES_PASSWORD=test -e POSTGRES_DB=ducklake_catalog \
  -p 5432:5432 postgres:17

# Wait for it
sleep 5

python3 - <<'PY'
import duckdb, traceback

con = duckdb.connect()
con.execute("INSTALL postgres; LOAD postgres")
con.execute("""
    CREATE SECRET pg_secret (
        TYPE postgres,
        HOST 'localhost', PORT 5432,
        DATABASE 'ducklake_catalog',
        USER 'postgres', PASSWORD 'test'
    )
""")
con.execute(
    "ATTACH 'ducklake:postgres:dbname=ducklake_catalog "
    "host=localhost user=postgres password=test' AS lake "
    "(DATA_PATH '/tmp/ducklake_data/')"
)

# This is the documented disable
con.execute("SET ducklake_default_data_inlining_row_limit = 0")

# 1601-col CREATE — one past PG's 1600 limit
cols_def = ", ".join(f'c{i} INTEGER' for i in range(1601))
try:
    con.execute(f"CREATE TABLE lake.main.wide_test ({cols_def})")
    print("UNEXPECTED: CREATE succeeded")
except Exception as e:
    print("FAILED as documented in this issue:")
    print(f"  {type(e).__name__}: {e}")
PY
```

Expected output:
```
FAILED as documented in this issue:
  TransactionException: TransactionContext Error: Failed to commit DuckLake transaction.
  ...
  Error: HTTP 500 from Postgres ("tables can have at most 1600 columns")
  Statement: CREATE TABLE "public".ducklake_inlined_data_<tid>_0 (
    row_id BIGINT, begin_snapshot BIGINT, end_snapshot BIGINT,
    "c0" BYTEA, "c1" BYTEA, ... , "c1600" BYTEA
  );
```

generated with help from Claude Code - behavior verified by me and report drafted by me

### OS:

Linux (AKS pod)

### DuckDB Version:

1.5.2

### DuckLake Version:

415a9ebd

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Seth Glickman

### Affiliation:

Biome

No - I cannot share the data sets because they are confidential

- [x] Yes, I have

- [x] Yes, I have

utay:
Fix #1184

Setting `ducklake_default_data_inlining_row_limit = 0` disables row inlining, but `WriteNewInlinedTables` still created empty `ducklake_inlined_data_*` tables because it used the context-free limit, which defaults to 10.

This PR uses the context-aware limit, preventing both the metadata entry and physical backing table from being created when inlining is disabled globally.


pdet:
Thanks for the PR, can we retarget to `main`?

utay:
> Thanks for the PR, can we retarget to `main`?

Done!

pdet:
Seems that CI got borked, can you maybe push an empty commit to kickstart it?

utay:
> Seems that CI got borked, can you maybe push an empty commit to kickstart it?

Arg, missed it. All good now

PR #1453, opened 2026-09-16
Check compaction conflicts at file granularity

### LINKED ISSUE #1452: Compaction conflict detection is table-level, so concurrent compaction of one table always conflicts even on disjoint file sets
### What happens

Two concurrent `ducklake_merge_adjacent_files` transactions on the **same table** always conflict at commit, even when they provably touch **disjoint file sets**:

```
Transaction conflict - attempting to compact table with index "128"
  - but another transaction has compacted it
```

This makes **any** form of parallel compaction on a single table impossible — whether workers are split by file size, by partition, or by shard.

### Why — table-granularity check with no file-level fallback

`src/include/storage/ducklake_transaction_changes.hpp:44`

```cpp
set<TableIndex> tables_merge_adjacent;      // records table_id only — never which files
```

`src/storage/ducklake_transaction_state.cpp:271-276`

```cpp
for (auto &table_id : changes.tables_merge_adjacent) {
    ConflictCheck(table_id, other_changes.dropped_tables,        "compact table", "dropped it");
    ConflictCheck(table_id, other_changes.tables_deleted_from,   "compact table", "deleted from it");
    ConflictCheck(table_id, other_changes.tables_merge_adjacent, "compact table", "compacted it");  // <-- throws
    ConflictCheck(table_id, other_changes.tables_rewrite_delete, "compact table", "compacted it");
}
```

Because compaction only ever records `table_id`, the check has no file information to fall back on and throws immediately.

### The delete path already solves this — compaction just doesn't use it

`src/storage/ducklake_transaction_state.cpp:233-256` implements a **two-phase** check: a cheap table-level pre-filter, escalating to file granularity only when it hits.

```cpp
if (!changes.tables_deleted_from.empty()) {
    bool check_for_matches = false;
    for (auto &table_id : changes.tables_deleted_from) {
        if (other_changes.tables_deleted_from.find(table_id) != other_changes.tables_deleted_from.end()) {
            check_for_matches = true; break;
        }
    }
    if (check_for_matches) {
        const auto deleted_files = GetFilesDeletedOrDroppedAfterSnapshot(executor);
        ...
        ConflictCheck(file.data_file_id, deleted_files.deleted_from_files,
                      "delete from file", "deleted from it");
    }
}
```

So two large deletes on different files of one table commit fine, while two compactions on different files of one table do not. That asymmetry is the bug. It is the same root cause as #1287 (inlined deletes over-conflicting at table granularity), on a different operation.

### Proposed fix

Give `merge_adjacent` the same two-phase treatment:

1. Keep `tables_merge_adjacent` as the cheap pre-filter — no serialisation change, no `ducklake_snapshot_changes` growth.
2. On a table-level hit, escalate to a file-level check rather than throwing. Compaction retires files by setting `ducklake_data_file.end_snapshot`, so the catalog already holds the answer: the transaction need only verify that every file it intends to retire still has `end_snapshot IS NULL` as of its snapshot. `GetFilesDeletedOrDroppedAfterSnapshot` and `set<DataFileIndex> deleted_from_files` (`ducklake_metadata_info.hpp:308`) look directly reusable.

The same escalation would apply to `tables_rewrite_delete`.

**Alternative considered and rejected:** keying by `(table_id, partition_id)`. Cheaper, and sufficient for partition-sharded workers — but it does **not** help tiered compaction, where two workers sweep all partitions selecting different file-size bands and so collide on the same partition key. File granularity covers both.

### Why this matters

Compaction is currently single-threaded per table, so throughput is bounded by one worker. On a high-ingest lake that is the limiting factor: in our benchmark ingest produced ~100 small files/min, and a single compactor could not stay ahead without continuous full-table passes.

Two designs this blocks today, both correct by construction and rejected only by the coarse check:

* **Tiered** — one worker on small files (`max_file_size := 12MB`), another on larger (`min_file_size := 12MB, max_file_size := 32MB`). Disjoint by size.
* **Sharded** — workers partitioned by hash of partition values. Disjoint by partition.

### Reproduction

Two connections against one DuckLake catalog, disjoint size bands:

```sql
-- connection A
CALL ducklake_merge_adjacent_files('lake', max_file_size := 12000000);
-- connection B, concurrently
CALL ducklake_merge_adjacent_files('lake', min_file_size := 12000000, max_file_size := 32000000);
```

B fails with the conflict above. Observed with two separate compactor processes against a Postgres-backed catalog under continuous ingest; conflicts appeared within minutes and recurred on roughly every overlapping cycle.

### Environment

DuckDB v1.5.2, `ducklake` extension from the repository, PostgreSQL 18 catalog, S3 data path. The check is unchanged on `main`, so this also applies to the v2.0 line.

gtaneja-sentrinox:
Fixes #1452.

### Problem

Two transactions compacting the same table always conflict, even when they retire completely disjoint sets of files:

```
Transaction conflict - attempting to compact table with index "128" - but another transaction has compacted it
```

`TransactionChangeInformation` records compaction as `set<TableIndex> tables_merge_adjacent` / `tables_rewrite_delete` — table id only, no file identity. So at commit time `CheckForConflicts` cannot distinguish two compactions that fight over the same files from two that never touch the same file.

That makes compaction effectively single-writer per table, and blocks splitting the work across concurrent writers — by partition, or by file-size band (a small-file tier and a large-file tier running side by side, so small merges are not starved behind large ones).

### Fix

Escalate to file granularity when two compactions touch the same table, mirroring the two-phase check the delete path a few lines above already performs. The table-level check still short-circuits: the extra queries only run when another transaction compacted a table we are also compacting.

Two sources identify the files the other transaction retired:

- `GetFilesDeletedOrDroppedAfterSnapshot` — files that gained a delete file or an `end_snapshot` after our snapshot. This covers `ducklake_rewrite_data_files`, which sets `end_snapshot` on its source files.
- `GetMissingDataFiles` (new, second commit) — which of our own source file ids no longer exist in `ducklake_data_file`. This is required because `ducklake_merge_adjacent_files` does **not** set `end_snapshot`: `WriteMergeAdjacent` deletes the source rows outright and only records them in `ducklake_files_scheduled_for_deletion`, which carries no snapshot id.

A conflict is raised only if the other transaction retired one of the files we are retiring. `compact table` conflicts against **drop** and **delete** are unchanged — only compaction-vs-compaction is narrowed.

The first revision of this PR relied on `GetFilesDeletedOrDroppedAfterSnapshot` alone. Running the new test on a real build showed that this let two *overlapping* merges both commit and duplicated the merged rows, which is what the second commit fixes.

### Tests

- `test/sql/concurrent/concurrent_compaction_disjoint_files.test` writes four parquet files directly and registers them with `ducklake_add_data_files`, so the sizes are deterministic and the size bands cannot straddle a file — two files of ~361 B and two of ~81 KB, split at 10,000 bytes.
  - `con1` compacts the small band, `con2` the large band, concurrently — both must commit, 4 files become 2, and all 40010 rows survive.
  - Regression guard: `con3` and `con4` both compacting the *same* band must still produce `Transaction conflict`, and afterwards exactly 2 files are active with no duplicated rows.
- `test/sql/rewrite_data_files/test_rewrite_concurrency.test`: the expected conflict message now names the file (`attempting to compact file with index "0" - but another transaction has compacted or deleted it`) instead of the table.

### What I verified

Built and tested against DuckDB v1.5.5 on the `v1.5-variegata` branch (the code this PR touches is identical there and on `main`) in the extension-ci-tools manylinux images, for both `linux_amd64` and `linux_arm64`:

- the new test and the full `test/sql/concurrent` and `test/sql/compaction` suites pass on both architectures;
- the full non-slow `test/sql/*` suite (475 tests) passes on `linux_arm64`.

The branch has not been compiled against `main`'s DuckDB locally — CI will be the first run there.


src/storage/ducklake_transaction_state.cpp
```diff
@@ -269,17 +268,45 @@ void DuckLakeTransactionState::CheckForConflicts(const TransactionChangeInformat
...
+	}
+	if (compaction_overlap) {
+		// Another transaction compacted a table that we are also compacting. That only conflicts if it
+		// retired one of the files we are retiring - two compactions over disjoint file sets (e.g. split by
+		// file size or by partition) are independent. Escalate to file granularity, mirroring the check
+		// already performed for deletes above.
+		//
+		// A rewrite of deletes sets end_snapshot on its source files, which GetFilesDeletedOrDroppedAfterSnapshot
+		// reports. Merging adjacent files instead removes the source rows from ducklake_data_file entirely
+		// (see WriteMergeAdjacent), so we additionally verify that every file we are retiring still exists.
```
pdet, line 294: nit: lets avoid multiline comments

test/sql/concurrent/concurrent_compaction_disjoint_files.test
```diff
@@ -0,0 +1,130 @@
+# name: test/sql/concurrent/concurrent_compaction_disjoint_files.test
```
pdet, line 1: Can we also add a similar concurrent test for `ducklake_rewrite_data_files`?

pdet (commented):
Thanks, just had two comments

gtaneja-sentrinox:
Thanks. I will get the fix by weekend

PR #1458, opened 2026-09-19
Validate SET NOT NULL with unknown null statistics

jdctinuiti:
## Summary

When a column-stat row has unknown NULL statistics, `SET NOT NULL` currently skips validation and can accept a column containing NULLs. Scan the live rows when the NULL count is unknown, using the existing validation path. Known-zero statistics retain the fast path; missing statistics still fail closed.

The regression sets unknown metadata explicitly, reattaches to avoid cached statistics, and checks rejection for a column containing NULLs and success for a NULL-free column.

## Validation

- Existing `test/sql/constraints/not_null.test`: 81 assertions passed, including the new regression.
- `make format-fix` and `git diff --check` passed.


pdet:
Thanks!

PR #1466, opened 2026-09-22
Apply patches from duckdb/duckdb

maiadegraaf:
This PR applies patches from duckdb/duckdb:
- `0012-single-query-result-type.patch`
- `0013-skip-slow-musl-tests.patch`
- `0014-skip-schema-cache-tests-on-musl.patch`

It also bumps the duckdb submodules and `.github/workflows/MainDistributionPipeline.yml` to v2.0.0.


pdet:
Seems one of the workflow has a problem, can you have a look @maiadegraaf ?

maiadegraaf:
@pdet seems to be a test thats failing now. Could you have a look?

pdet:
> @pdet seems to be a test thats failing now. Could you have a look?

that's typically new tests from the duckdb base that are failing, you probably just need to skip them in the appropriate test config.

maiadegraaf:
Added the tests to be skipped except `nan_min_statistics.test` since it looks like a real bug?

pdet:
> Added the tests to be skipped except `nan_min_statistics.test` since it looks like a real bug?

In this case you can also skip it with a "FIXME"

